PR #4, opened 2019-10-16
[DB-28313] Enable passing of certificates directly to the engine

vegichan:
Allow the passing of certificates directly to the engine (docker) without using the CSR signing capabilities of the admin.

This allows the use of CA=false certificates created by things like Vault.

To start an engine with the same certificate as the admin provide the following option:
database.te.otherOptions.keyStore = <KEYSTORE_PATH>

Next steps:
- expand on our documentation and explain what the different models are
- verify the use of certificate vs CSR in the admin (requires change to how logging is done)
- re-test with CA=false certificates

### DIFF
--- a/stable/database/values.yaml
+++ b/stable/database/values.yaml
@@ -54,6 +54,10 @@ admin:
   # tlsCACert:
   #   secret: nuodb-ca-cert
   #   key: ca.cert
+  # tlsKeyStore:
+  #   secret: nuodb-server-keystore
+  #   key: server.p12
+  #   password: changeIt
   # tlsClientPEM:
   #   secret: nuodb-client-pem
   #   key: client.pem
@@ -152,7 +156,12 @@ database:
     # labels
     # Additional Labels given to the SMs started
     labels: {}
+
     engineOptions: {}
+
+    # named key/value pairs that need to be passed to the image, such as
+    # keystore: "/etc/nuodb/keys/nuoadmin.p12"
+    otherOptions: {}
 
     # Additional Options for DaemonSets
     # affinityNoHotCopyDS: {}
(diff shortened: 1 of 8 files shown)


stable/database/values.yaml
```diff
@@ -54,6 +54,10 @@ admin:
   # tlsCACert:
   #   secret: nuodb-ca-cert
   #   key: ca.cert
+  # tlsKeyStore:
+  #   secret: nuodb-server-keystore
+  #   key: server.p12
```
sivanov-nuodb: Is `server.p12` going to work? We always mount the secret in nuoadmin.p12.

stable/database/values.yaml
```diff
@@ -152,8 +156,13 @@ database:
     # labels
     # Additional Labels given to the SMs started
     labels: {}
+
     engineOptions: {}
 
+    # named key/value pairs that need to be passed to the image, such as
+    # keystore: "/etc/nuodb/keys/nuoadmin.p12"
```
sivanov-nuodb: Isn't this hardcoded in the secret mounting? What happens if the user configures something like this:
```
database.te.otherOptions.keystore: "/etc/nuodb/keys/stanimir.p12"
database.sm.otherOptions.keystore: "/etc/nuodb/keys/stanimir.p12"
admin.tlsKeyStore.key:  "stanimir.p12",
```
Is it going to work?

stable/database/values.yaml
vegichan: its the same text as in the admin charts

stable/database/values.yaml
vegichan: yeah, that should work too. Keep in mind that all these commented out sections only exist as documentation/examples.

PR #7, opened 2019-10-21
[DB-28313] add missing test logic

vegichan:
Verify that the TLS chain is as expected in the admin

### DIFF
--- a/test/minikube/minikube_base_database_test.go
+++ b/test/minikube/minikube_base_database_test.go
@@ -113,10 +113,10 @@ func verifyPacketFetch(t *testing.T, namespaceName string, admin0 string) {
 	t.Log("tar contents: ", output)
 }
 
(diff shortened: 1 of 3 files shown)


test/minikube/minikube_base_database_test.go
```diff
@@ -113,10 +113,10 @@ func verifyPacketFetch(t *testing.T, namespaceName string, admin0 string) {
 	t.Log("tar contents: ", output)
 }
 
-func startDatabase(t *testing.T, namespaceName string, adminPod string, options *helm.Options) {
+func startDatabase(t *testing.T, namespaceName string, adminPod string, options *helm.Options) (helmChartReleaseName string) {
```
NikTJ777: Ah!
I had considered doing this when I refactored the tests for Backup - great to see it being done!

PR #13, opened 2019-10-30
[DB-28964] Externalize config files using generalized idiomatic approaches.

rbuck:
**Changes**

- Add idiomatic approach for handling configuration files to the database chart.
- Tweak the admin chart `admin.configFiles` variable to follow the same idiom.
- Add doc.
- Fixed the unit test; ONLY expose a ConfigMap IFF the `admin.configFiles` value is set.

**Test Changes Explained**

- verifyCustomFileDoesNotGetMounted in TestKubernetesInvalidLicense is no longer valid as the intent of this change is to permit ANY file in `/usr/etc/` to be overridden.
- an empty ({}) config map is pointless, so we idiomatically don't emit it (just like in MySQL, etc...)

**Testing Performed**

- Full end-to-end on GKE.
- Lint tests.
- Admin unit tests to verify common code approach.

**See Also**

The approach is very common, no need to reinvent how to do this. Lets just rip off the upstream project with known, trusted, working code...

- https://github.com/helm/charts/blob/master/stable/mysql/values.yaml
- https://github.com/helm/charts/blob/master/stable/mysql/templates/_helpers.tpl
- https://github.com/helm/charts/blob/master/stable/mysql/templates/configurationFiles-configmap.yaml
- https://github.com/helm/charts/blob/master/stable/mysql/templates/deployment.yaml

### DIFF
--- a/stable/admin/templates/statefulset.yaml
+++ b/stable/admin/templates/statefulset.yaml
@@ -111,8 +111,8 @@ spec:
         volumeMounts:
         - name: log-volume
           mountPath: /var/log/nuodb
-        {{- if .Values.admin.configFiles }}
-        {{- range $key, $val := .Values.admin.configFiles }}
+        {{- with .Values.admin.configFiles }}
+        {{- range $key, $val := . }}
         - name: configurations
           mountPath: {{ $.Values.admin.configFilesPath }}{{ $key }}
           subPath: {{ $key }}
--- a/stable/admin/templates/statefulset.yaml
+++ b/stable/admin/templates/statefulset.yaml
@@ -111,11 +111,13 @@ spec:
         volumeMounts:
         - name: log-volume
           mountPath: /var/log/nuodb
-{{- if hasKey .Values.admin.configFiles "nuodb.lic" }}
-        - name: nuodb-license-volume
-          mountPath: /etc/nuodb/nuodb.lic
-          subPath: nuodb.lic
-{{ end }}
+        {{- if .Values.admin.configFiles }}
+        {{- range $key, $val := .Values.admin.configFiles }}
+        - name: configurations
+          mountPath: {{ $.Values.admin.configFilesPath }}{{ $key }}
+          subPath: {{ $key }}
+        {{- end -}}
+        {{- end }}
         - name: raftlog
           mountPath: /var/opt/nuodb
         {{- if .Values.admin.tlsCACert }}
@@ -166,14 +168,11 @@ spec:
       {{- end }}
       - name: log-volume
         emptyDir: {}
-{{- if hasKey .Values.admin.configFiles "nuodb.lic" }}
-      - name: nuodb-license-volume
+      {{- if .Values.admin.configFiles }}
+      - name: configurations
         configMap:
-          name: nuodb-license-file
-          items:
-            - key: nuodb.lic
-              path: nuodb.lic
-{{ end }}
+          name: {{ template "admin.fullname" . }}-configuration
+      {{- end }}
 {{- if not .Values.admin.persistence.enabled }}
       - name: raftlog
         emptyDir: {}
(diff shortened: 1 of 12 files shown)


stable/admin/templates/statefulset.yaml
```diff
@@ -111,11 +111,13 @@ spec:
         volumeMounts:
         - name: log-volume
           mountPath: /var/log/nuodb
-{{- if hasKey .Values.admin.configFiles "nuodb.lic" }}
-        - name: nuodb-license-volume
-          mountPath: /etc/nuodb/nuodb.lic
-          subPath: nuodb.lic
-{{ end }}
+        {{- if .Values.admin.configFiles }}
```
NikTJ777: Minor point:
For this kind of idiom, I prefer:
```
{{- with .Values.admin.configFiles }}
{{ - range $key, $val := . }}
```
It avoids repeating the .Values reference, and avoids bugs - either now or later - where the 2 references diverge.

stable/admin/templates/statefulset.yaml
```diff
@@ -111,11 +111,13 @@ spec:
...
+        {{- range $key, $val := .Values.admin.configFiles }}
```
NikTJ777: Minor point:
For this kind of idiom, I prefer:

{{- with .Values.admin.configFiles }}
{{ - range $key, $val := . }}
It avoids repeating the .Values reference, and avoids bugs - either now or later - where the 2 references diverge.

stable/admin/values.yaml
```diff
@@ -114,7 +117,7 @@ admin:
   # --set admin.configFiles.nuodb\\.lic=<BASE64-TEXT-HERE>
```
NikTJ777: Is it worth handling the `dot` in the pathname specially?
Perhaps converting any `underscore` char in the pathname to a `dot` - or similar?

stable/database/templates/statefulset.yaml
```diff
@@ -133,6 +133,13 @@ spec:
         resources:
 {{ toYaml .Values.database.sm.resources | trim | indent 10 }}
         volumeMounts:
+        {{- if .Values.database.configFiles }}
+        {{- range $key, $val := .Values.database.configFiles }}
```
NikTJ777: There are a few of then, aren't there?

stable/database/templates/statefulset.yaml
```diff
@@ -352,6 +364,13 @@ spec:
...
```
NikTJ777: And one last one - just it doesn't get overlooked...

stable/database/values.yaml
```diff
@@ -122,6 +122,14 @@ database:
     ping-timeout: 60
     max-lost-archives: 0
 
+  # Custom NuoDB configuration files path
+  configFilesPath: /etc/nuodb/
+
+  # Custom NuoDB configuration files used to override default NuoDB settings
+  configFiles: {}
+    # nuodb.config: |-
```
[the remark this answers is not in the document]
rbuck: No we don't. But we're no different than any other database in this respect.

stable/admin/values.yaml
rbuck: Not really.

stable/database/values.yaml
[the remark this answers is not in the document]
rbuck: This is really a design flaw of Helm and similar systems that conflate the meaning of dot.

stable/database/templates/statefulset.yaml
rbuck: Fixed

rbuck:
> * Does doc need updating to mention all possible configs that can be supplied at the moment - right now it only mentions the license file in admin for example.  Could also be nuoadmin.conf?
> * Do we need to worry about overlapping options that can be passed on cmd line via values admin.options for example, and also supplied in a config file?  I guess nuo will take the command line args as overriding?

I would hope that Nuo handles CLI overriding. If it didn't, oh boy!

PR #18, opened 2019-11-04
Add engine readiness probe

vegichan:
remove `admin.persistence.enabled` and make the persistent mode the default. This matches the settings in the database layer (no database.sm.persistence.enabled) exists and it the safe/reasonable default

Add a readiness probe to the engine layer. It makes sure that the process is fully up (seen from the admin layer) before it is marked as READY.

Add small integration tests for the probe.

Add upgrade tests and rolling upgrade tests. These show the readiness probe the best. A deployment/statefulset only restarts one engine before it proceeds with the next.

Refactor a few tests. Move some of the smaller tests back into `short` to rebalance the short vs long runs. Both are now around 30 min.

### DIFF
--- a/stable/database/files/readinessprobe
+++ b/stable/database/files/readinessprobe
@@ -1,3 +1,21 @@
 #!/bin/sh
 
-nuocmd show domain | grep `hostname` | sed 's/.*start_id = //;s/].*//' | awk '{print "check process --check-running --start-id " $1'} | xargs nuocmd
+function die() {
+  retval=$1
+  shift
+  echo "$@"
+  exit $retval
+}
+
+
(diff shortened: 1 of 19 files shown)


stable/database/files/readinessprobe
```diff
@@ -0,0 +1,21 @@
...
+
+function die() {
+  retval=$1
+  shift
+  echo "$@"
+  exit $retval
+}
+
+
+processes=`nuocmd show domain --process-format "{address} {start_id}" | grep \`hostname\` | awk '{print $2}'`
```
sivanov-nuodb: Is this going to work if the user specifies `altAddr` for and engine? Most probably `grep` won't match anything and pod will never become ready.

PR #19, opened 2019-11-05
Ac/make admin optional or internal

acabrele:
- Made admin L4 load balancer optional and with internal / external IP option
- Added TE L4 load balancer option with internal / external IP options (e.g for DB Visualiser)
- Test updates
- Doc updates

### DIFF
--- a/stable/admin/README.md
+++ b/stable/admin/README.md
@@ -171,6 +171,8 @@ The following tables list the configurable parameters for the `admin` option of 
 | `replicas` | Number of NuoDB Admin replicas | `1` |
 | `lbPolicy` | Load balancer policy name | `nil` |
 | `lbQuery` | Load balancer query | `nil` |
+| `loadBalancerService.enabled` | Whether to deploy a Layer 4 cloud load balancer service for the admin layer | `false` |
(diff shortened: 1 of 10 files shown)


stable/admin/README.md
```diff
@@ -171,6 +171,8 @@ The following tables list the configurable parameters for the `admin` option of
 | `replicas` | Number of NuoDB Admin replicas | `1` |
 | `lbPolicy` | Load balancer policy name | `nil` |
 | `lbQuery` | Load balancer query | `nil` |
+| `loadBalancerService.enabled` | Whether to deploy a Layer 4 cloud load balancer service for the admin layer | `false` |
```
adriansuarez: # The name `loadBalancer` is very overloaded. Can we call this `adminK8sService` or something that makes it clear that this has nothing to do with load-balancing of TEs?
# We currently rely on a round-robin among admins to distribute database processes among admin processes (each database process is connected to exactly one admin process). If we do not use K8s round-robin load-balancing, will all database processes be bound to a particular admin process?

acabrele: Sure I'll do that in the new PR

acabrele: On 2nd point, this is specifically for disabling/enabling the external L4 service, the internal headless service is still in place, so admin connection using that would still be round robin.

PR #23, opened 2019-11-12
Ac/new make admin optional or internal

acabrele:
- Made admin L4 load balancer optional and with internal / external IP option

- Added TE L4 load balancer option with internal / external IP options (e.g for DB Visualiser)

- Test updates

- Doc updates

### DIFF
--- a/stable/admin/README.md
+++ b/stable/admin/README.md
@@ -171,6 +171,8 @@ The following tables list the configurable parameters for the `admin` option of 
 | `replicas` | Number of NuoDB Admin replicas | `1` |
 | `lbPolicy` | Load balancer policy name | `nil` |
 | `lbQuery` | Load balancer query | `nil` |
+| `loadBalancerService.enabled` | Whether to deploy a Layer 4 cloud load balancer service for the admin layer | `false` |
+| `loadBalancerService.internalIP` | Whether to use an internal or external IP address for the load balancer | `nil` |
 | `resources` | Labels to apply to all resources | `{}` |
 | `affinity` | Affinity rules for NuoDB Admin | `{}` |
 | `nodeSelector` | Node selector rules for NuoDB Admin | `{}` |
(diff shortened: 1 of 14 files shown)


test/minikube/minikube_long_admin_test.go
```diff
@@ -14,7 +14,11 @@ func TestKubernetesBasicAdminThreeReplicas(t *testing.T) {
 	testlib.AwaitTillerUp(t)
 
 	options := helm.Options{
-		SetValues: map[string]string{"admin.replicas": "3"},
+		SetValues: map[string]string{
+			"admin.replicas": "3",
+			"admin.k8sL4Service.enabled": "true",
+			"admin.k8sL4Service.internalIP": "false",
```
vegichan: is this something leftover from testing? Why is it required? All these tests are being run in minikube, not in AWS/Google/Azure. There is no L4Service

acabrele: Without it the existing test fails - minikube creates the service but it never gets it's external IP address (since there is no cloud).  So it's either remove the test or add the new values so it still creates in the same way, and the test passes...

vegichan: Got it. I want to make sure that the default properties work and give you a functional environment.

acabrele: As discussed, removed all minikube L4 tests

stable/admin/README.md
```diff
@@ -171,6 +171,8 @@ The following tables list the configurable parameters for the `admin` option of
 | `replicas` | Number of NuoDB Admin replicas | `1` |
 | `lbPolicy` | Load balancer policy name | `nil` |
 | `lbQuery` | Load balancer query | `nil` |
+| `k8sL4Service.enabled` | Whether to deploy a Layer 4 cloud load balancer service for the admin layer | `false` |
```
NikTJ777: I'm not a huge fan of the name including the implementation details.
Sort of like the old "hungarian" notation in Visual C++.

If the `k8sL4` qualification needed to distinguish this option from other `services` for the admin layer?

NikTJ777: Could we name it "service" or "lbService" or similar?

acabrele: Aarrgh... naming... I had load balancer in there before and Adrian said that's too overloaded and confuses with lb policy / nuodb lb's...

k8sService?
service?
externalService?
... names in a hat :-)

NikTJ777: Sorry to put you through the "naming from hell" pain...

I would be happy with "externalService" or "externalAccess".
People like BobB prefer names that are consistent with the underlying feature, so I assume some would prefer "externalService" for that reason - but in truth the term "service" is badly overloaded already - so a name relying on it is less helpful.

acabrele: I'm going for externalAccess... :-)

PR #24, opened 2019-11-12
[DB-29147] Fix admin chart lb-policy job to avoid creating dead jobs at startup.

rbuck:
**Changes**

Change the admin lb-policy job to `--wait-forever` instead of polling loop to avoid a hailstorm of dead job objects when Admin is coming up.

Thanks, Dirk Butson, for the great feedback and patch!!!

**Testing Performed**

Test on GKE.


vegichan:
@butson the issue here is that if the admin does not come up within 5 min, the job will fail?

PR #30, opened 2019-12-05
add minimum supported NuoDB version to README

vegichan:


vegichan:
@sivanov-nuodb NuoDB New Admin was not GA in 3.4.

PR #33, opened 2019-12-11
Make DB Services for direct TE connections turn-offable.

acabrele:
-Add example values entry for disabling
-Add logic to service templates


stable/database/values.yaml
```diff
@@ -190,6 +190,11 @@ database:
     externalAccess: {}
       # enabled: false
       # internalIP: true
+    
+    ## By default, database clusterip and headless services for direct TE connections are enabled,
+    ## but can be optionally disbaled here
```
vegichan: disabled

PR #37, opened 2019-12-20
fix 4.0.3 backup / restore incompatibility

acabrele:
-Use new kv format in nuosm script
-Use nuodocker in backup test
-Uuse nuodb 4.0.3


acabrele:
No longer needed as Nik has delivered PR38

PR #38, opened 2019-12-24
Enhanced backup and restore including PiT backup

NikTJ777:
* Add support for journal hotcopy backup;
  - See new sub-sections under database.hotcopy section;
  - Add cronjobs to database chart for full-hotcopy, incremental-hotcopy, and journal-hotcopy;
    - ensure that full-hotcopy and journal-hotcopy are started if hotcopy.journalBackup is enabled;

* Add a cronjob to the database chart that can be triggered by a restore, to force a new full-backup immediately after a restore operation;
  - see post-restore;

* Add autoImport and autoRestore values to explicitly control automated use of Import and Restore backups when initialising or reinitialising an SM;

* Add support for separate backup "groups" within a database - eg 2 separate datacentres;

* major rewrite of restore logic in the nuosm script;
* Encapsulate backup logic into new script 'nuobackup' which is mounted from a configMap;
* Encapsulate restore logic into a new script 'nuorestore' which is mounted fro a configMap;

* Add cloud.clusterName to database values, and include clusterName into Admin, TE and SM pod names;
   - support separate backup schedules for each backup group;

* Change structure of database.envFrom and admin.envFrom sections;

* add helper template for iterating a set of key-value pairs

## Updated Tests
* Added clusterName (default = 'cluster0') to all tests that verify the Admin, TE, or SM pod names;
* Adjusted tests that verify the number of secrets;
* Adjust tests of envFrom structure;
* rewrote DatabaseRestore Test to more accurately reflect a restore workflow;
  - avoid a number of Await issues;
  - original test code commented out - will be removed during the PR;
* Add diganostics when fidnPod() fails;

* Added tests for Admin KV get and set;


test/integration/template_resources_test.go
```diff
@@ -240,11 +240,13 @@ func TestPullSecretsRenderAllNuoDB(t *testing.T) {
...
 	helm.RenderTemplate(t, options, "../../stable/database", []string{"templates/statefulset.yaml"})
 	helm.RenderTemplate(t, options, "../../stable/database", []string{"templates/deployment.yaml"})
+	helm.RenderTemplate(t, options, "../../stable/database", []string{"templates/cronjob.yaml"})
+	helm.RenderTemplate(t, options, "../../stable/database", []string{"templates/job.yaml"})
 
 	helm.RenderTemplate(t, options, "../../stable/transparent-hugepage", []string{"templates/daemonset.yaml"})
 
 	helm.RenderTemplate(t, options, "../../stable/backup", []string{"templates/cronjob.yaml"})
-	helm.RenderTemplate(t, options, "../../stable/backup", []string{"templates/job.yaml"})
+	//helm.RenderTemplate(t, options, "../../stable/backup", []string{"templates/job.yaml"})
```
vegichan: delete the commented-out lines please. This template no longer exists, correct?

test/integration/template_resources_test.go
```diff
@@ -259,11 +261,13 @@ func TestPullSecretsRenderAllGlobal(t *testing.T) {
...
```
vegichan: delete the commented-out lines please

test/integration/template_secrets_test.go
```diff
@@ -23,7 +23,8 @@ func TestSecretsDatabaseDefault(t *testing.T) {
 	var object v1.Secret
 	helm.UnmarshalK8SYaml(t, output, &object)
 
-	assert.Check(t, len(object.StringData) == 5)
+	// make sure we got at least the 3 minumum secrets: database-name, database-username, and database-password
+	assert.Check(t, len(object.StringData) >= 3)
```
vegichan: same comment as above. Why more than 3?

test/testlib/minikube_utilities.go
```diff
@@ -547,3 +560,30 @@ func UnmarshalJSONArray(t *testing.T, stringJSON string) []interface{} {
...
+	start = time.Now()
+	output, err = k8s.RunKubectlAndGetOutputE(t, options,
+		"exec", podName, "--", "nuocmd", "get", "value", "--key", "test/minikube",
+	)
+	assert.NilError(t, err, "Could not get KV value")
+	elapsed = time.Since(start)
+
+	assert.Check(t, elapsed.Seconds() < 2.0)
+
+	assert.Check(t, output == "testVal", fmt.Sprintf("KV get took longer than 2s: %s", elapsed))
```
vegichan: the error message is wrong

test/testlib/minikube_utilities.go
```diff
@@ -547,3 +560,30 @@ func UnmarshalJSONArray(t *testing.T, stringJSON string) []interface{} {
...
+
+	// verify the KV store can write and read in a reasonable time
+	start := time.Now()
+	output, err := k8s.RunKubectlAndGetOutputE(t, options,
+		"exec", podName, "--", "nuocmd", "set", "value", "--key", "test/minikube", "--value", "testVal", "--unconditional",
+	)
+	assert.NilError(t, err, "Could not set KV value")
+	elapsed := time.Since(start)
+
+	assert.Check(t, elapsed.Seconds() < 2.0, fmt.Sprintf("KV set took longer than 2s: %s", elapsed))
```
vegichan: why do you wanna check that it takes less than 2 sec? Is there some bug that you are trying to find? What does the failure of the test tell us?

I think that it might just tell us that the test VM was slow.

test/minikube/minikube_base_database_test.go
```diff
@@ -142,41 +142,44 @@ func verifyEngineAltAddress(t *testing.T, namespaceName string, admin0 string, e
 }
 
 func backupDatabase(t *testing.T, namespaceName string, podName string, databaseName string, options *helm.Options) {
-	randomSuffix := strings.ToLower(random.UniqueId())
+	// NTJ- currently database chart automatically starts the hotcopy Job
```
vegichan: if the logic has changed, please fix the test to represent the new logic. Do not leave commented-out code around.

test/minikube/minikube_base_database_test.go
```diff
@@ -186,7 +189,7 @@ func restoreDatabase(t *testing.T, namespaceName string) {
 	helm.Install(t, options, testlib.RESTORE_HELM_CHART_PATH, restName)
 	testlib.AddTeardown(testlib.TEARDOWN_RESTORE, func() { helm.Delete(t, options, restName, true) })
 
-	testlib.AwaitPodPhase(t, namespaceName, restName, corev1.PodSucceeded, 120*time.Second)
+	testlib.AwaitPodPhase(t, namespaceName, "restore-demo-", corev1.PodSucceeded, 120*time.Second)
```
vegichan: it took me a while to understand what this name is:
It is `JOBName`-`randomSHA` where `JOBName` is `name: restore-{{ include "restore.target" . }}`.

The SHA is not the same SHA that the Helm chart gets, it is autogenerated by K8s for the particular pod belonging to the JOB.

test/minikube/minikube_base_database_test.go
```diff
@@ -419,56 +422,91 @@ func TestKubernetesRestoreDatabase(t *testing.T) {
...
 		// run a manual backup
-		k8s.RunKubectl(t, opts,
-			"exec", admin0, "--",
-			"nuodocker", "backup", "database",
-			"--db-name", "demo",
-			"--type", "full",
-			"--timeout", "120",
-			"--backup-root", "/var/opt/nuodb/backup",
-		)
+		// k8s.RunKubectl(t, opts,
```
vegichan: Do not leave commented-out code behind

test/minikube/minikube_base_database_test.go
```diff
@@ -419,56 +422,91 @@ func TestKubernetesRestoreDatabase(t *testing.T) {
...
+		// 	"--user", "dba",
+		// 	"--password", "secret",
+		// 	"demo",
+		// 	"--file", "/opt/nuodb/samples/quickstart/sql/Teams.sql",
+		// )
 
 		defer testlib.Teardown(testlib.TEARDOWN_RESTORE)
 		restoreDatabase(t, namespaceName)
 
+		// NTJ - Now automatically triggered by the restore chart (by default - can be overridden)
```
vegichan: Do not leave commented-out code behind

test/minikube/minikube_base_database_test.go
```diff
@@ -419,56 +422,91 @@ func TestKubernetesRestoreDatabase(t *testing.T) {
...
-			"--db-name", "demo",
-		)
-
+		// k8s.RunKubectl(t, opts,
+		// 	"exec", admin0, "--",
+		// 	"nuocmd", "shutdown", "database",
+		// 	"--db-name", "demo",
+		// )
+
+		// wait for the datbase to be restarted (shutdown followed by restart)
```
vegichan: typo in database

test/minikube/minikube_base_database_test.go
```diff
@@ -419,56 +422,91 @@ func TestKubernetesRestoreDatabase(t *testing.T) {
...
+		testlib.AwaitDatabaseDown(t, namespaceName, admin0, "demo")
 		testlib.AwaitDatabaseUp(t, namespaceName, admin0, "demo", 2)
 
-		// verify that the database contains the restored data
-		tables, err := testlib.RunSQL(t, namespaceName, admin0, "demo", "show schema User")
-		assert.NilError(t, err, "error running SQL: show schema User")
-		assert.Check(t, strings.Contains(tables, "HOCKEY"), "tables returned: ", tables)
+		// // verify that the database contains the restored data
```
vegichan: Do not leave commented-out code behind

test/testlib/constants.go
```diff
@@ -40,4 +40,6 @@ const IMPORT_ARCHIVE_URL = "http://download.nuohub.org/ce_releases/restore.bak.t
 const MINIMAL_VIABLE_ENGINE_CPU = "500m"
 const MINIMAL_VIABLE_ENGINE_MEMORY = "500Mi"
 
-const BACKUP_NUODB_VERSION = "4.0.2"
\ No newline at end of file
+const BACKUP_NUODB_VERSION = "4.0.2"
+
+const BACKUP_SEMAPHORE = "/nuobackup/semaphore"
```
vegichan: this is unused

test/testlib/minikube_utilities.go
```diff
@@ -187,6 +187,10 @@ func findPod(t *testing.T, namespace string, expectedName string) (*corev1.Pod,
 		}
 	}
 
+	for _, pod := range findAllPodsInSchema(t, namespace) {
```
vegichan: I like this!

test/testlib/nuodb_database_utilities.go
```diff
@@ -19,6 +19,7 @@ type ExtractedOptions struct {
 	NrSmNoHotCopyPods int
 	NrSmPods          int
 	DbName            string
+	ClusterName		  string
```
vegichan: The new line contains a single TAB instead of SPACES. It looks OK in GoLand (on my machine), but it is screwed up in the review. Can you fix it to be the same as all the lines above?

test/testlib/nuodb_database_utilities.go
```diff
@@ -46,6 +47,11 @@ func GetExtractedOptions(options *helm.Options) (opt ExtractedOptions) {
 		opt.DbName = "demo"
 	}
 
+	opt.ClusterName = options.SetValues["cloud.clusterName"]
```
vegichan: can you add this logic to `minikube_test_test.go`?

test/testlib/nuodb_database_utilities.go
```diff
@@ -75,12 +81,12 @@ func StartDatabase(t *testing.T, namespaceName string, adminPod string, options
 	AwaitNrReplicasScheduled(t, namespaceName, smPodName, opt.NrSmPods)
 
 	tePodName := GetPodName(t, namespaceName, tePodNameTemplate)
-	AwaitPodStatus(t, namespaceName, tePodName, corev1.PodReady, corev1.ConditionTrue, 120*time.Second)
+	AwaitPodStatus(t, namespaceName, tePodName, corev1.PodReady, corev1.ConditionTrue, 240*time.Second)
```
vegichan: 4 minutes for a TE to start seems excessive. Have you encountered such long times? I have only seen it with admins, when the minikube instance needs to pull a new image.

The SM time below is cumulative. Can it really take 5min + whatever time the TE took? (lets call it 6-10 min)

stable/backup/templates/cronjob.yaml
```diff
@@ -148,6 +136,9 @@ spec:
               subPath: {{ .Values.admin.tlsClientPEM.key }}
             {{- end }}
           volumes:
+        - name: nuosm
+          mountPath: /usr/local/bin/nuosm
+          subPath: nuosm
```
vegichan: the indent here does not match. Since it doesn't work, is it required at all?

vegichan:
the tests no longer test the `backup` template. What does it do now? Should it be deleted?

vegichan:
I see no changes to the documentation

stable/restore/files/nuorestore
```diff
@@ -0,0 +1,82 @@
...
+
+: ${NUODB_BACKUP_KEY:=/nuodb/nuobackup}
+: ${NUODB_RESTORE_REQUEST_PREFIX:=/nuodb/nuosm}
+
+restore_type="database"
+db_name=$DB_NAME
+semaphore="/nuobackup/semaphore"
+restore_url=$NUODB_RESTORE_URL
+backup_group="all"
+labels=""
```
vegichan: these appear to be unused.

stable/restore/files/nuorestore
```diff
@@ -0,0 +1,82 @@
...
+        "--type="* )
+           restore_type="${opt#*=}";;
+        "--db-name" )
+           db_name="$1"; shift;;
+        "--db-name="* )
+           db_name="${opt#*=}";;
+        "--semaphore" )
+           semaphore="$1"; shift;;
+        "--semaphore="* )
+           semaphore="${opt#*=}";;
```
vegichan: not in the comment on the top

stable/restore/files/nuorestore
```diff
@@ -0,0 +1,82 @@
+#!/bin/sh
+#
+# simple script to perform a NuoDB restore.
+# Sets a flag to tell newly-starting SM(s) to perform the restore; then optionally causes the SM(s) to restart
+
+# args are:
+#   --type      - retore type: ( database | archive )
+#   --db-name   - database name to perform the restore operation on
+#   --source    - source to restore from - can be a backupset or a URL
+#   --group     - arbitrary group to select the SMs to backup. Often the datacentre or cloud name
```
vegichan: not part of the args while loop

stable/restore/files/nuorestore
```diff
@@ -0,0 +1,82 @@
...
+#   --labels    - additional arbitrary labels to select the SMs to backup
```
vegichan: not part of the args while loop

stable/admin/templates/statefulset.yaml
```diff
@@ -50,6 +50,15 @@ spec:
           mountPath: /var/opt/nuodb
         - name: log-volume
           mountPath: /var/log/nuodb
+      - name: init-raft
+        image: {{ template "init.image" . }}
+        imagePullPolicy: {{ default "" .Values.busybox.image.pullPolicy }}
+        command: ['rm' , '-rf', '/var/opt/nuodb/raftlog' ]
```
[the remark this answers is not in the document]
NikTJ777: Good find.
This is meant to be:
a) conditional on the admin.persistence.initialize flag;
b) a one-off event when restarting the admin layer - and then disabled again;

There is no easy way to delete the raftlog, and there are multiple recovery scenarios where that is needed.

However, I'll remove this from this PR, and handle it as a separate branch and PR, since criteria b) above requires a better solution.

stable/admin/values.yaml
```diff
@@ -70,6 +70,7 @@ admin:
   #   internalIP: true
 
   persistence:
+    initialize: false
```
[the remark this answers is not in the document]
NikTJ777: It is meant to allow for a "restart admin --reset-broker-state" as was supported in Old Admin, but is not supported in New Admin.

stable/backup/templates/cronjob.yaml
```diff
@@ -17,21 +17,15 @@ spec:
       template:
         spec:
           containers:
-          - name: nuodb-ce
+          - name: nuodb
```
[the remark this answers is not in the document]
NikTJ777: Someone previous had confused the pod name with the image name...

stable/database/templates/deployment.yaml
```diff
@@ -54,8 +54,10 @@ spec:
           - "300"
           - "--options"
           - "mem {{ .Values.database.te.memoryOption}} {{- range $opt, $val := .Values.database.te.engineOptions }} {{$opt}} {{$val}} {{- end}}"
+        {{- with .Values.database.te.labels }}
           - "--labels"
-          - "{{- range $opt, $val := .Values.database.te.labels }} {{$opt}} {{$val}} {{- end}}"
+          - "{{- range $opt, $val := . }} {{$opt}} {{$val}} {{- end}}"
+        {{- end }}
```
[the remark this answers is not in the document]
NikTJ777: The {{with}} element will output nothing if the tested value is null or empty.

So this new syntax avoids having "--labels" followed by an empty string if database.te.labels is empty.

test/testlib/nuodb_database_utilities.go
NikTJ777: Sure - I actually hate the number of whitespace-only changes these different IDEs are causing.

test/testlib/minikube_utilities.go
NikTJ777: :)

test/testlib/nuodb_database_utilities.go
NikTJ777: Um - difficult to say with precision.
I was having _lots_ of timeout issues on the minikube environment on my laptop.

I can try shortening the timeouts again.
At least part of the problem was an accumulation of incomplete tests in my minikube environment.

stable/backup/templates/cronjob.yaml
NikTJ777: At this stage, I can think of no purpose for the backup chart.

I can move it to the incubator tree; or just delete it.

stable/restore/files/nuorestore
NikTJ777: Good catch - thanks!

stable/restore/files/nuorestore
NikTJ777: Thanks

stable/restore/files/nuorestore
NikTJ777: Not sure what you mean.
type, db-name, and source are; group is not.

stable/backup/templates/cronjob.yaml
NikTJ777: In actual fact, the template unit tests in integration/template_resources_test.go do actually verify that the stable/backup/template/cronjob.yaml renders correctly.

But yes, with the rewrite of the integration tests in minikube/minikube_base_database_test.go, the stable/backup chart is no longer tested.

test/testlib/nuodb_database_utilities.go
NikTJ777: Sure! :)

test/testlib/constants.go
NikTJ777: BACKUP_SEMAPHORE removed

stable/database/files/nuobackup
```diff
@@ -0,0 +1,161 @@
...
+   next=$(( (current + 1) % $NUODB_MAX_BACKUP_HISTORY ))
+else
+   if [ -z "$latest" ]; then
+      echo >&2 "No full backup has completed yet - try again later"
+      exit -1
+   fi
+
+   # retrieve the latest backupset name
+   backupset=$(nuocmd --api-server $NUOCMD_API_SERVER get value --key $NUODB_BACKUP_KEY/$db_name/${backup_group}/${latest} )
+   echo >&2 "Looking up $NUODB_BACKUP_KEY/$db_name/${backup_group}/${latest} returned: $backupset"
```
sivanov-nuodb: This looks like INFO message. Why it is redirected to stderr?

stable/database/files/nuobackup
```diff
@@ -0,0 +1,161 @@
+#!/bin/sh
+#
+# simple script to automate backup
```
sivanov-nuodb: It's not very simple, isn't it?
Can we have longer description?

stable/database/templates/cronjob.yaml
```diff
@@ -0,0 +1,347 @@
...
+            {{- if .Values.admin.tlsCACert }}
+            - name: tls-ca-cert
+              mountPath: /etc/nuodb/keys/ca.cert
+              subPath: {{ .Values.admin.tlsCACert.key }}
+            {{- end }}
+            {{- if .Values.admin.tlsClientPEM }}
+            - name: tls-client-pem
+              mountPath: /etc/nuodb/keys/nuocmd.pem
+              subPath: {{ .Values.admin.tlsClientPEM.key }}
+            {{- end }}
```
sivanov-nuodb: Don't we have a helper for this? It looks like repetitive code.

stable/database/templates/cronjob.yaml
```diff
@@ -0,0 +1,347 @@
...
+            secret:
+              secretName: {{ .Values.admin.tlsCACert.secret }}
+              defaultMode: 0440
+          {{- end }}
+          {{- if .Values.admin.tlsClientPEM }}
+          - name: tls-client-pem
+            secret:
+              secretName: {{ .Values.admin.tlsClientPEM.secret }}
+              defaultMode: 0440
+          {{- end }}
```
sivanov-nuodb: Don't we have a helper for this? It looks like repetitive code.

stable/restore/values.yaml
```diff
@@ -53,21 +59,31 @@ admin:
...
   # * a URL from which to download the backup (typically a TAR.GZ file)
-  backupSet: ""
+  source: ""
+
+  # credentials for the specified target - if different to the database.autoRestore.credentials
+  credentials: ""
 
+  # autoRestart
+  # if set to "true", running the restore chart causes the database to be restarted automatically
+  autoRestart: "true"
```
sivanov-nuodb: Isn't it better the default to be `false`?

stable/database/files/nuobackup
```diff
@@ -0,0 +1,161 @@
...
+
+backup_type="full"
+db_name=$DB_NAME
+backup_group=""
+labels=""
+semaphore=""
+timeout=1800
+backup_root=$BACKUP_DIR
+
+while [[ $# -gt 0 ]] && [[ "$1" == "-"* ]] ;
```
madscientist: `[[` and `]]` are not POSIX-standard facilities.  Since you say `#!/bin/sh` at the top please use POSIX test operations: this can be written `while [ $# -gt 0 ];` then in the `case` below you can check for options not starting with `-` and `break` out of the while loop

madscientist: FYI, the best way to test a script for portability is to run it with `dash`.  On a Debian/Ubuntu based system, `/bin/sh` is already `dash` and non-POSIX scripts will simply fail to run at all.  On a RedHat/CentOS based system, you'll have to install `dash` separately and invoke it directly to test.

test/minikube/minikube_base_database_test.go
```diff
@@ -408,56 +411,91 @@ func TestKubernetesRestoreDatabase(t *testing.T) {
...
-		)
-
+		// k8s.RunKubectl(t, opts,
+		// 	"exec", admin0, "--",
+		// 	"nuocmd", "shutdown", "database",
+		// 	"--db-name", "demo",
+		// )
+
+		// wait for the datbase to be restarted (shutdown followed by restart)
+		testlib.AwaitDatabaseDown(t, namespaceName, admin0, "demo")
```
vegichan: you can not write the test like this. It will fail randomly. By the time control reaches this point, the process might be already up and this will timeout.

stable/backup/templates/cronjob.yaml
acabrele: I think move the original version referencing nuodocker to incubator...

test/minikube/minikube_base_database_test.go
NikTJ777: I take your point - although:
* database shutdown is not quick - it takes time for the processes to shut down, and report that fact to admin.

* the database restart is triggered by `k8s` - which involves guaranteed delays (timeouts).
In addition, the SM startup performs multiple `nuocmd get value` operations, each of which are slow - a second or more elapsed to complete.
So the window for any race condition relies on the database shutting down unusually quickly, and `k8s` detecting that and restarting the TEs quickly enough that all that has occurred _before_ the `check database` command has commenced.

>  It will fail randomly. 
It *may* fail randomly...

> You have to either shut the DB down manually 
That is what is happening - norestore calls `nuocmd shutdown database`;
After which, `k8s` restarts the database pods 

> or await the incarnation in the admin
Yes - I believe this is the only guaranteed deterministic solution
I will change the test to retrieve the incarnation number, and then Await for that to increase.

stable/restore/values.yaml
NikTJ777: @sivanov-nuodb 
Most users would expect that deploying the `restore` chart will *_cause_* a restore.
The `autoRestart = true` ensures that.

If the default were "false", then the default behaviour would be to set a flag that the next time the database is restarted, it should restore from backup - but it would not trigger that restart as part of deploying the `restore` chart.

stable/database/templates/cronjob.yaml
NikTJ777: Actually, I agree.
I copy-n-pasted from existing cronjobs, but there should be a helper that packages this logic up.

stable/database/files/nuobackup
NikTJ777: I will augment the comments.

stable/database/files/nuobackup
NikTJ777: The caller reads stdout to get the result of a `get-latest` operation, so we should avoid sending diagnostic information to stdout.

stable/database/files/nuobackup
NikTJ777: As always - your feedback on shell/bash scripting is much appreciated - thank you!

I have been using `set -n` as a poor-man's lint for shell scripts, so I'll try to improve that by using `dash` for that.

- Or, alternatively, I'll check whether I should just change the shebang to `/bin/bash`.

test/minikube/minikube_base_database_test.go
vegichan: let me rephrase. This has failed and this is failing. So this needs to be fixed before it can go in.

If you wanna talk about how to fix it, we can either do it here or in SLACK.

NikTJ777: My apologies for being unclear.
When I wrote:
> I will change the test to retrieve the incarnation number, and then Await for that to increase.
I meant that I am making that change now.

stable/admin/templates/_helpers.tpl
```diff
@@ -12,14 +12,15 @@ If release name contains chart name it will be used as a full name.
 */}}
 {{- define "admin.fullname" -}}
 {{- $domain := default "domain" .Values.admin.domain -}}
+{{- $cluster := default "cluster0" .Values.cloud.clusterName -}}
 {{- if .Values.admin.fullnameOverride -}}
-{{- .Values.admin.fullnameOverride | trunc 63 | trimSuffix "-" -}}
+{{- .Values.admin.fullnameOverride | trunc 53 | trimSuffix "-" -}}
```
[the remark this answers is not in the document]
NikTJ777: I've changed the trunc value to "50", and I've updated the comments to say why.

stable/database/templates/cronjob.yaml
NikTJ777: The repetitive code was already in place - I just created an additional YAML file with this duplication - in other words I duplicated the duplication).

We should create a JIRA for abstracting this YAML.

stable/admin/values.yaml
```diff
@@ -83,9 +83,11 @@ admin:
     capabilities: []
 
   ## Specify one or more configMaps to import Environment Variables from
-  # Ex:  envFrom: [ configMapRef: { name: myConfigMap } ]
-  envFrom: []
-  
+  # Ex:  configMapRef: [ myConfigMap, myOtherConfigMap ]
+  envFrom:
+    configMapRef: []
+    secretRef: []
```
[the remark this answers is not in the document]
NikTJ777: They aren't currently rendered.

I have removed the element from the values file for both admin and database, and created a JIRA - DB-29732 to implement "envFrom.secretRef".

test/testlib/nuodb_database_utilities.go
NikTJ777: I have set the timeouts to 120 secs and 200 secs respectively.
I will see how they perform in Travis.

stable/restore/README.md
```diff
@@ -162,45 +162,18 @@ The following tables list the configurable parameters of the backup chart and th
 
 | Parameter | Description | Default |
 | ----- | ----------- | ------ |
-| `timeout` | Job deadline (timeout) | 1800 |
-| `backupPvc` | PVC name for the backup volume | `nil` |
-| `archivePvc` | PVC name for the archive volume | `nil` |
-| `backupSet` | The backup set name to restore from | `nil` |
+| `restore.target` | Where to restoe `TO` | `{{ .Values.database.name }}` |
+| `restore.source` | Where to retore `FROM` [ backupset | url | `:latest` ] | `:latest` |
```
vegichan: typo in `restore`

NikTJ777: *ugh*

I've fixed, and it will be in the next push.

stable/database/values.yaml
```diff
@@ -137,40 +143,97 @@ database:
...
+      # Deadline for starting a job - if a job hasn't started in this time - give up
+      deadine: 1800
+
+      # timeout for completing a job - if a job hasn't completed in this time - stop waiting for it
+      timeout: 1800
+
+      successHistory: 5
+      failureHistory: 5
+
+      backupDir: /var/opt/nuodb/backup
```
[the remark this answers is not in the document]
NikTJ777: Yup - when an SM runs out of disk, it performs an auto-shutdown - fondly referred to by engineering as a *crash*.

stable/database/values.yaml
```diff
@@ -137,40 +143,97 @@ database:
...
         memory: 8Gi
+
+    ## Affinity, selector, and tolerations
+    # There are expanded as YAML, and can include variable and template references
     affinity: {}
     # nodeSelector: {}
     # tolerations: []
     
-    # labels
+    ## labels
```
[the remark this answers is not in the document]
NikTJ777: YAML does nothing special with ## - we just use it as a heading comment.

stable/database/values.yaml
```diff
@@ -137,40 +143,97 @@ database:
...
     # Additional Labels given to the SMs started
     labels: {}
 
+    # additional options to pass to the starting NuoDB process.
```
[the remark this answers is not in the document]
NikTJ777: Haven't you worked it out yet that TEs are second-class citizens?? :)

test/integration/template_database_test.go
```diff
@@ -54,7 +54,8 @@ func TestDatabaseSecretsDefault(t *testing.T) {
 	var object v1.Secret
 	helm.UnmarshalK8SYaml(t, output, &object)
 
-	assert.Check(t, len(object.StringData) == 5)
+	// check for the minimum 3 secret values: database-name, database-password, database-username
+	assert.Check(t, len(object.StringData) >= 3)
```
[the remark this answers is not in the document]
NikTJ777: The original test was verifying that the 3 primary secrets were there - by comparing with `==`.

I subsequently added more secrets - so I changed the test to ensure there were _at least_ those 3 secrets.

This means we don't have to modify the test every time a new secret is added.

So what should be the purpose of test?
* to ensure that the 3 minimal secrets are rendered: db-name, db-user, db-password
* to ensure that all known secrets that are/could be in the secret are validated?

test/integration/template_database_test.go
```diff
@@ -729,9 +730,9 @@ func TestDatabaseLabeling(t *testing.T) {
 				basicChecks(obj.Spec.Template.Spec.Containers[0].Args)
 
 				if testlib.IsStatefulSetHotCopyEnabled(&obj) {
-					assert.Check(t, ArgContains(obj.Spec.Template.Spec.Containers[0].Args, "backup enabled"))
-				} else {
-					assert.Check(t, ArgContains(obj.Spec.Template.Spec.Containers[0].Args, "backup disabled"))
+					assert.Check(t, ArgContains(obj.Spec.Template.Spec.Containers[0].Args, "backup cluster0"))
+					// } else {
+					// 	assert.Check(t, ArgContains(obj.Spec.Template.Spec.Containers[0].Args, "backup disabled"))
```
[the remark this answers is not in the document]
NikTJ777: Dammit - I thought I had.

Good catch - thanks!

test/testlib/minikube_utilities.go
NikTJ777: Or possible, the error message is correct, but in the wrong set of parens... :)

test/integration/template_database_test.go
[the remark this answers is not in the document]
vegichan: My original plan for the test was to verify that all expected things get rendered. I would understand if you wanted to check that there are more than 3 if the other 3 get only sometimes rendered and you do not want to deal with the if-clauses. But all of them get rendered all the time, right?

You could, of course, break it down and verify them all manually, but that is annoying.

test/testlib/minikube_utilities.go
NikTJ777: Two things I have observed:
1. KV set/get is slooooow.
Even on our Rancher rig, I was seeing ~1s elapsed per operation.

2. Under some - as yet undetermined - conditions, the New Admin Raft gets brain-staggeringly slow:
~10-20secs to return from an operation - including KV set/get.

The new `nuosm` performs at least 10 KV get/set operations.
If they are regularly going to take >2 sec each, then we will hit timeouts with SMs starting - simply due to the time taken by KV get/set operations.

So this test is supposed to flag the problem with slooooow KV, so we don't waste time hunting down SM/TE timeouts that are actually being caused by Raft running through treacle.

test/minikube/minikube_base_database_test.go
NikTJ777: Done in a previous push :)

test/minikube/minikube_base_database_test.go
```diff
@@ -142,51 +142,51 @@ func verifyEngineAltAddress(t *testing.T, namespaceName string, admin0 string, e
...
-	helm.Install(t, options, testlib.BACKUP_HELM_CHART_PATH, bakName)
 
 	// wait for the backup to both start _and_ complete successfully
-	backupJob := fmt.Sprintf("backup-%s-job-full", databaseName)
+	backupJob := fmt.Sprintf("hotcopy-%s-job-initial", databaseName)
 	testlib.AwaitPodPhase(t, namespaceName, backupJob, corev1.PodSucceeded, 120*time.Second)
 
 	// verify that the backup has been documented by the Admin layer
-	backupName := fmt.Sprintf("%s/%s", testlib.LAST_BACKUP_PREFIX, databaseName)
+	// backupName := fmt.Sprintf("%s/%s", testlib.LAST_BACKUP_PREFIX, databaseName)
```
vegichan: this commented out code should go

test/minikube/minikube_base_database_test.go
NikTJ777: No I am referring to datbase - as distinct to disbase... :)

test/minikube/minikube_base_database_test.go
```diff
@@ -142,51 +142,51 @@ func verifyEngineAltAddress(t *testing.T, namespaceName string, admin0 string, e
...
+		"nuodocker", "get", "current-backup", "--db-name", databaseName,
 	)
 
-	assert.NilError(t, err, "Error running: nuocmd get value --key ", backupName)
+	assert.NilError(t, err, "Error running: nuodocker get current-backup  ")
 	assert.Check(t, backupset != "")
 }
 
-func restoreDatabase(t *testing.T, namespaceName string) {
+func restoreDatabase(t *testing.T, namespaceName string, podName string, databaseOptions *helm.Options, restart string) {
```
vegichan: `restart` is always true, so all of that `restore` code is unused. Also GoLang has a `bool` type, no need to compare a string to `"true"`

stable/database/files/nuobackup
```diff
@@ -0,0 +1,161 @@
+#!/bin/sh
+#
+# simple script to automate backup
+#
+
+# args are:
+#   --type     - backup type: ( full | incremental | journal | get-latest )
```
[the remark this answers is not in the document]
NikTJ777: yup - fixed.

Good catch :)

test/testlib/minikube_utilities.go
vegichan: Got it. That makes sense! Thanks. Resolving.

test/minikube/minikube_base_database_test.go
NikTJ777: I was making the function reasonably general purpose.
So restart is only "true" if the caller sets it so.

I passed in the string "true" rather than the bool to make the assignment to "restore.autoRestart" simpler and cleaner - because Go has no trinary operator.

vegichan: I am not a fan of function that are general purpose today, but might break in the future. You are also only ever checking true and nothing else.

`restore()` is not called when it is not true.

NikTJ777: Having said that about converting from bool to string being less simple - I can see a way to do it now.

So I might change the method signature - the purist in me, and in us all... :)

NikTJ777: And having said _all_ of the above - I am not a big fan of bool parameters to methods, because they are so easily misunderstood.

vegichan: That is what `enums` are for ;)

NikTJ777: The code:
```
else {
   restore()
}
```
calls restore() when the value is not true...

NikTJ777: But I am happy to remove that complication until it is needed.

NikTJ777: > That is what enums are for ;)
I went on a jagg to replace all uses of string args and Hashtable keys in Java with Enums - wherever possible - so as to enlist the compiler in detecting typos, rather than the runtime.

The result was generally good - but various, frequently-used, pieces of code became clumsy as a result...

PR #42, opened 2020-01-13
Backup & restore, permissions secrets fixes

acabrele:
- Add restore credentials missing from hotcopy SM statefulset
- Fix CHOWN
- Move image pull secrets for proper indenting when they are included

### DIFF
--- a/stable/database/templates/statefulset.yaml
+++ b/stable/database/templates/statefulset.yaml
@@ -179,7 +179,6 @@ spec:
         {{- end }}
         terminationMessagePath: /dev/termination-log
         terminationMessagePolicy: File
-{{- include "nuodb.imagePullSecrets" . | indent 6 }}
         readinessProbe:
           initialDelaySeconds: 30
           periodSeconds: 15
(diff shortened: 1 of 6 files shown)


stable/database/templates/statefulset.yaml
```diff
@@ -313,7 +313,7 @@ spec:
       initContainers:
       - name: init-disk
         image: {{ template "init.image" . }}
-        imagePullPolicy: {{ default "" .Values.busybox.imagepullPolicy | quote }}
+        imagePullPolicy: {{ default "" .Values.busyboximagepullPolicy | quote }}
```
NikTJ777: I don't this line is correct.

The values in the values file are `nuodb.image.pullPolicy` and `busybox.image.pullPolicy`
There is also a template in `_helpers.yaml` called `nuodb.imagePullSecrets`

This new line seems to be a confusion of the 2.

I presume the original error is the missing of the period between `image` and `pullPolicy`.

So perhaps: `.Values.busybox.imagepullPolicy`
should be: `.Values.busybox.image.pullPolicy`

PR #45, opened 2020-01-14
Ac/backup restore fixes

acabrele:
- Fix unquoted strings produced by autoRestore and autoImport sources,  which cause helm template / kubectl apply (and ArgoCD Sync) to fail.

For some reason helm install is unaffected.

### DIFF
--- a/stable/database/templates/configmap.yaml
+++ b/stable/database/templates/configmap.yaml
@@ -67,8 +67,8 @@ data:
   NUODB_RESTORE_REQUEST_PREFIX: {{ default "/nuodb/nuosm/database" .Values.nuodb.requestKey }}
   NUODB_LATEST_BACKUP_PREFIX: {{ default "nuodb-backup/last_created" .Values.nuodb.latestPrefix }}
   NUODB_BACKUP_KEY: {{ default "/nuodb/nuobackup" .Values.nuodb.latestKey }}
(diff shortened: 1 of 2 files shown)


stable/database/templates/configmap.yaml
```diff
@@ -67,8 +67,8 @@ data:
   NUODB_RESTORE_REQUEST_PREFIX: {{ default "/nuodb/nuosm/database" .Values.nuodb.requestKey }}
   NUODB_LATEST_BACKUP_PREFIX: {{ default "nuodb-backup/last_created" .Values.nuodb.latestPrefix }}
   NUODB_BACKUP_KEY: {{ default "/nuodb/nuobackup" .Values.nuodb.latestKey }}
-  NUODB_AUTO_IMPORT: {{ .Values.database.autoImport.source }}
-  NUODB_AUTO_RESTORE: {{ .Values.database.autoRestore.source }}
+  NUODB_AUTO_IMPORT: {{ default "" .Values.database.autoImport.source | quote}}
```
sivanov-nuodb: Most of the time {{default}} is used like this
```drink: {{ .Values.favorite.drink | default "tea" | quote }}```
This a style specific so will leave this to you to decide.

acabrele: I noticed this too, I kept it in line with what we were already doing...

PR #47, opened 2020-01-15
Bump NuoDB version to 4.0.4

vegichan:


vegichan:
we will see how the tests go

PR #51, opened 2020-01-24
Add cluster domain support

acabrele:
- New values entries for specifying the cluster domain and name of the local and entrypoint k8s clusters
- Template additions to set defaults and define the entrypoint admin name
- Admin Statefulset changes to always include the fully qualified service names
- Doc updates
- New template test for new values

When using multi cluster solutions, it might be required to specify different cluster domains for each cluster that will be connected, enabling DNS resolution to be forwarded to the remote cluster.

Changes tested on a contrail and calico based multicluster.

I think we can baseline multi-cluster CNI requirements as

- Unique k8s cluster domain per cluster (e.g cluster1.local, cluster2.local)

- Unique Pod CIDR ranges per cluster & pod IP routing

- Unique service CIDR ranger per cluster & service IP routing

- CoreDNS config updated to forward DNS requests for the other clusters domains appropriately 

- Helm charts set correctly with the new values cluster.entrypointDomain and cluster.entrypointName

It is also likely the CNI solution will require a site to site VPN to exist between the clouds


stable/admin/README.md
```diff
@@ -50,6 +50,8 @@ The following tables list the configurable parameters for the `cloud` option:
 | `provider` | Cloud provider; permissible values include: `azure`, `amazon`, or `google` |`nil`|
 | `zones` | List of availability zones to deploy to |`[]`|
 | `clusterName` | logical name of the cluster. Useful in multi-cluster configs | `cluster0` |
+| `clusterDomain` | Kubernetes domain as specified during kubernetes set up. Useful in multi-cluster configs | `cluster.local` |
```
NikTJ777: Is it worth creating a `cluster` section, and having distinct values under that?

Eg:
```
cloud:
   cluster:
      name: cluster0
      domain: cluster.local
      entryPointDomain: cluster.local
```

stable/admin/templates/statefulset.yaml
```diff
@@ -75,8 +75,8 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
```
NikTJ777: I suggest it would be better to do all the concatenation in the {{template}} in the _helpers.yaml file.

stable/admin/values.yaml
```diff
@@ -18,7 +18,13 @@ cloud:
   # cluster name is used to make resources unique in multi-cluster configurations.
   # If the NuoDB domain spans 2 or more physical clusters, then each cluster must have a unique clusterName.
   # The default is fine for single-cluster domains.
-  clusteName: cluster0
+  clusterName: cluster0
```
NikTJ777: As per first comment, I would suggest that we have enough `cluster` values that it now warrants a separate `cluster` section.

stable/admin/README.md
acabrele: Why the heck not... Done!

stable/admin/templates/statefulset.yaml
acabrele: I've done this but kept the cluster domain definition separate still, with the thought it might be needed in the future on its own

stable/admin/templates/_helpers.tpl
```diff
@@ -144,3 +144,40 @@ Import ENV vars from configMaps
 {{- define "admin.envFrom" }}
 envFrom: [{{- range $index, $map := .Values.admin.envFrom.configMapRef }}{{if gt $index 0}},{{end}} configMapRef: { name: {{$map}} } {{ end }}]
 {{- end -}}
+
+{{/*
+Define the cluster domains
+*/}}
+{{- define "cluster.domain" -}}
+{{- .Values.cloud.cluster.domain | default "cluster.local" }}
```
adriansuarez: I'm not sure what the syntax is for this template language, but I see both of the following used interchangeably:
```
{{- ... -}}
```

```
{{ ... }}
```

This one has `{{- ... }}`. Is this a typo? Are all of these valid?

adriansuarez: Dirk Butson just explained it to me; `{{-` means strip whitespace before and `-}}` means strip whitespace after.

stable/admin/templates/_helpers.tpl
```diff
@@ -144,3 +144,40 @@ Import ENV vars from configMaps
...
+{{- if contains $name .Release.Name -}}
+{{- printf "%s-%s-%s" .Release.Name .Values.admin.domain $cluster | trunc 50 | trimSuffix "-" -}}
+{{- else -}}
+{{- printf "%s-%s-%s-%s" .Release.Name .Values.admin.domain $cluster $name | trunc 50 | trimSuffix "-" -}}
+{{- end -}}
+{{- end -}}
+{{- end -}}
+
+{{- define "nuodb.domainEntrypoint" -}}
+{{ include "admin.entrypointFullname" . }}-0.{{ .Values.admin.domain }}.$(NAMESPACE).svc.{{ include "cluster.entrypointDomain" . }}
```
adriansuarez: So the `cluster<ordinal>` tag will be present in every FQDN? Was this already the case using the old variable `clusterName`?

adriansuarez: I guess this is already the case, but without the ordinal:

```
$ kubectl exec -it edgy-peahen-nuodb-cluster0-admin-0 -- hostname -f
edgy-peahen-nuodb-cluster0-admin-0.nuodb.default.svc.cluster.local
```

acabrele: cluster domain is configured in the k8s install so will always be there, defaulting to cluster.local but requiring changing for multicluster so that DNS can forward correctly

PR #52, opened 2020-01-28
Add optional persistent volumes for logs

acabrele:
- New values entries to define log storage
- Stateful set changes to provision persistent log volumes
- Add mounting of the nuoadmin script
- Code to retain previous crash outputs to avoid overwriting
- Change to always deploy local-storage class, not only if we are not using a cloud
- Readme updates

Note that this change does not include persistent logging for TEs

### DIFF
--- a/stable/admin/README.md
+++ b/stable/admin/README.md
@@ -185,6 +185,10 @@ The following tables list the configurable parameters for the `admin` option of 
 | `persistence.accessModes` | Volume access modes enabled (must match capabilities of the storage class) | `ReadWriteMany` |
 | `persistence.size` | Amount of disk space allocated for admin RAFT state | `10Gi` |
 | `persistence.storageClass` | Storage class for volume backing admin RAFT state | `-` |
+| `logPersistence.enabled` | Whether to enable persistent storage for logs | `20Gi` |
+| `logPersistence.accessModes` | Volume access modes enabled (must match capabilities of the storage class) | `ReadWriteOnce` |
+| `logPersistence.size` | Amount of disk space allocated for log storage | `20Gi` |
+| `logPersistence.storageClass` | Storage class for volume backing log storage | `-` |
 | `envFrom` | Import ENV vars from one or more configMaps | `[]` |
 | `options` | Set optons to be passed to nuoadmin as arguments | `{}` |
 | `securityContext.capabilities` | add capabilities to the container | `[]` |
--- a/stable/admin/README.md
+++ b/stable/admin/README.md
@@ -189,7 +189,7 @@ The following tables list the configurable parameters for the `admin` option of 
 | `logPersistence.overwriteBackoff.copies` | How many copies of the crash directory to keep within windowMinutes | `3` |
 | `logPersistence.overwriteBackoff.windowMinutes` | The window within which to keep the number of crash copies | `120` |
 | `logPersistence.accessModes` | Volume access modes enabled (must match capabilities of the storage class) | `ReadWriteOnce` |
-| `logPersistence.size` | Amount of disk space allocated for log storage | `20Gi` |
+| `logPersistence.size` | Amount of disk space allocated for log storage | `60Gi` |
 | `logPersistence.storageClass` | Storage class for volume backing log storage | `-` |
 | `envFrom` | Import ENV vars from one or more configMaps | `[]` |
 | `options` | Set optons to be passed to nuoadmin as arguments | `{}` |
--- a/stable/admin/README.md
+++ b/stable/admin/README.md
@@ -182,7 +182,6 @@ The following tables list the configurable parameters for the `admin` option of 
 | `tolerations` | Tolerations for NuoDB Admin | `[]` |
 | `configFilesPath` | Directory path where `configFiles.*` are found | `/etc/nuodb/` |
 | `configFiles.*` | See below. | `{}` |
-| `persistence.enabled` | Whether or not persistent storage is enabled for admin RAFT state | `false` |
 | `persistence.accessModes` | Volume access modes enabled (must match capabilities of the storage class) | `ReadWriteMany` |
 | `persistence.size` | Amount of disk space allocated for admin RAFT state | `10Gi` |
 | `persistence.storageClass` | Storage class for volume backing admin RAFT state | `-` |
@@ -240,8 +239,7 @@ Verify the Helm chart:
 
 ```bash
 helm install nuodb/admin -n admin \
-    --set persistence.enabled=true \
-    --set persistence.storageClass=nuodb-admin \
+    --set persistence.storageClass=standard-storage \
     --debug --dry-run
 ```
 
@@ -249,8 +247,7 @@ Deploy the administration tier using volumes of the specified storage class:
 
 ```bash
 helm install nuodb/admin -n admin \
-    --set persistence.enabled=true \
-    --set persistence.storageClass=nuodb-admin
+    --set persistence.storageClass=standard-storage
 ```
 
   **Tip**: Be patient, it will take approximately 55 seconds to deploy.
--- a/stable/admin/README.md
+++ b/stable/admin/README.md
@@ -185,7 +185,7 @@ The following tables list the configurable parameters for the `admin` option of 
 | `persistence.accessModes` | Volume access modes enabled (must match capabilities of the storage class) | `ReadWriteMany` |
 | `persistence.size` | Amount of disk space allocated for admin RAFT state | `10Gi` |
 | `persistence.storageClass` | Storage class for volume backing admin RAFT state | `-` |
-| `logPersistence.enabled` | Whether to enable persistent storage for logs | `20Gi` |
+| `logPersistence.enabled` | Whether to enable persistent storage for logs | `false` |
 | `logPersistence.accessModes` | Volume access modes enabled (must match capabilities of the storage class) | `ReadWriteOnce` |
 | `logPersistence.size` | Amount of disk space allocated for log storage | `20Gi` |
 | `logPersistence.storageClass` | Storage class for volume backing log storage | `-` |
--- a/stable/admin/README.md
+++ b/stable/admin/README.md
@@ -186,6 +186,8 @@ The following tables list the configurable parameters for the `admin` option of 
 | `persistence.size` | Amount of disk space allocated for admin RAFT state | `10Gi` |
 | `persistence.storageClass` | Storage class for volume backing admin RAFT state | `-` |
 | `logPersistence.enabled` | Whether to enable persistent storage for logs | `false` |
+| `logPersistence.overwriteBackoff.copies` | How many copies of the crash directory to keep within windowMinutes | `3` |
+| `logPersistence.overwriteBackoff.windowMinutes` | The window within which to keep the number of crash copies | `120` |
 | `logPersistence.accessModes` | Volume access modes enabled (must match capabilities of the storage class) | `ReadWriteOnce` |
 | `logPersistence.size` | Amount of disk space allocated for log storage | `20Gi` |
 | `logPersistence.storageClass` | Storage class for volume backing log storage | `-` |
--- a/stable/admin/files/nuoadmin
+++ b/stable/admin/files/nuoadmin
@@ -26,8 +26,8 @@ export NUODB_LOGDIR NUODB_CFGDIR NUODB_VARDIR NUODB_BINDIR NUODB_RUNDIR
 
 # attempt to retain the previous crash directory (within the configured window to avoid filling the disk)
 if [ -f $NUODB_CRASHDIR/* ]; then
-  crashcount=$(find $NUODB_LOGDIR/crash-* -type d -cmin -10 ! -path . | wc -l)
-  if [ $crashcount -lt 2 ]; then
+  crashcount=$(find $NUODB_LOGDIR/crash-* -type d -cmin -$OVERWRITE_WINDOW ! -path . | wc -l)
+  if [ $crashcount -lt $OVERWRITE_COPIES ]; then
     crashbackupdir="$NUODB_LOGDIR/crash-$( date +%Y%m%dT%H%M%S )/"
     mkdir $crashbackupdir
     mv $NUODB_CRASHDIR/* $crashbackupdir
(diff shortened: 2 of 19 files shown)


stable/admin/README.md
```diff
@@ -182,10 +182,13 @@ The following tables list the configurable parameters for the `admin` option of
 | `tolerations` | Tolerations for NuoDB Admin | `[]` |
 | `configFilesPath` | Directory path where `configFiles.*` are found | `/etc/nuodb/` |
 | `configFiles.*` | See below. | `{}` |
-| `persistence.enabled` | Whether or not persistent storage is enabled for admin RAFT state | `false` |
 | `persistence.accessModes` | Volume access modes enabled (must match capabilities of the storage class) | `ReadWriteMany` |
 | `persistence.size` | Amount of disk space allocated for admin RAFT state | `10Gi` |
 | `persistence.storageClass` | Storage class for volume backing admin RAFT state | `-` |
+| `logPersistence.enabled` | Whether to enable persistent storage for logs | `20Gi` |
```
vegichan: `enabled` sounds like a boolean option. But this takes a file size string. How do you disable this?

stable/admin/README.md
```diff
@@ -182,10 +182,13 @@ The following tables list the configurable parameters for the `admin` option of
 | `tolerations` | Tolerations for NuoDB Admin | `[]` |
 | `configFilesPath` | Directory path where `configFiles.*` are found | `/etc/nuodb/` |
 | `configFiles.*` | See below. | `{}` |
-| `persistence.enabled` | Whether or not persistent storage is enabled for admin RAFT state | `false` |
```
vegichan: so RAFT now has to be persistent in all cases?

stable/admin/files/nuoadmin
```diff
@@ -0,0 +1,37 @@
...
+    if nuodocker --api-server "${NUODB_ALTERNATIVE_ENTRYPOINT}:8888" check servers ; then
+        # Admin server exists at alternative entrypoint. Use it instead
+        NUODB_DOMAIN_ENTRYPOINT="${NUODB_ALTERNATIVE_ENTRYPOINT}"
+    fi
+fi
+
+# if endpoint is not specified, default to bootstrap server
+if [ "${NUODB_DOMAIN_ENTRYPOINT}" = "" ] ; then
+    NUODB_DOMAIN_ENTRYPOINT="${NUODB_BOOTSTRAP_SERVERID}"
+fi
```
vegichan: all of this should be handled by the Admin. It seems like you are reinventing the wheel.

stable/database/README.md
```diff
@@ -238,6 +238,10 @@ The following tables list the configurable parameters of the `database` chart an
 | `persistence.storageClass` | Storage class for volume backing database archive storage | `-` |
 | `configFilesPath` | Directory path where `configFiles.*` are found | `/etc/nuodb/` |
 | `configFiles.*` | See below. | `{}` |
+| `sm.logPersistence.enabled` | Whether to enable persistent storage for logs | `20Gi` |
```
vegichan: same comment on the `boolean` issue

stable/database/README.md
```diff
@@ -315,17 +319,15 @@ Verify the Helm chart:
 
 ```bash
 helm install nuodb/database -n database \
-    --set sm.persistence.enabled=true \
-    --set sm.persistence.storageClass=nuodb-archive \
```
vegichan: was this just wrong?

stable/database/templates/statefulset.yaml
```diff
@@ -195,8 +195,10 @@ spec:
         configMap:
           name: {{ template "database.fullname" . }}-configuration
       {{- end }}
+      {{- if not .Values.database.sm.logPersistence.enabled }}
       - name: log-volume
         emptyDir: {}
+      {{- end }}
```
vegichan: does the SM log anything to this dir?

stable/storage-class/templates/storageclass.yaml
```diff
@@ -73,11 +73,11 @@ parameters:
 {{- include "storageClass.allowVolumeExpansion" . }}
 reclaimPolicy: Retain
 volumeBindingMode: Immediate
-{{- else }}
+{{- end }}
+---
```
vegichan: this was an unrelated bug, correct?

stable/admin/README.md
acabrele: copy n paste error, will fix

stable/admin/README.md
acabrele: i removed this because (I think) we had removed the actual option to disable some time ago - I can't see any references to admin.persistence.enabled...

stable/admin/files/nuoadmin
acabrele: i'm only mounting the existing nuoadmin file like we do nuosm, nuote etc to allow the additions to it... OK or not?

stable/database/README.md
acabrele: same copy-n-paste, will fix

stable/database/README.md
acabrele: Removed as I think we removed optional non-persistence long ago, cannot see any references to sm.persistence.enabled

stable/admin/files/nuoadmin
vegichan: This was one very confusing part of the Admin training session. We should sit down with Adrian and decide how entry points should work. This is what the `peer` property in Admin is for. The training says that you can set it to the load balancer.

stable/database/templates/statefulset.yaml
acabrele: log volume is still there but replaced with the persistent volume (still /var/log/nuodb)

acabrele: I can't see logs but the crashdir is inside

stable/storage-class/templates/storageclass.yaml
acabrele: I wouldn't say it was a bug, I just think that local storage shouldn't be excluded when running on cloud (e.g local ssd)

vegichan:
can you please add a `integration/template_database_test.go` test case that verifies that setting these values does the expected thing? It might be an overkill to write an E2E tests for this, but we should at least validate that the HELM template parses correctly.

stable/admin/files/nuoadmin
acabrele: Hmm, this comment seems unrelated to the code... but like I say this is _the_ nuoadmin file taken from the image, not changed regarding entrypoint.  I added the part that take a copy of the crash directory if it was populated, to avoid overwriting (and when persistent logging allowed it to be there after the crash)...

vegichan: oh. I missed that. Resolving. We can take the discussion offline.

acabrele:
> can you please add a `integration/template_database_test.go` test case that verifies that setting these values does the expected thing? It might be an overkill to write an E2E tests for this, but we should at least validate that the HELM template parses correctly.

Done!

stable/database/templates/statefulset.yaml
sivanov-nuodb: I believe that engines will store core files inside only.

stable/admin/files/nuoadmin
```diff
@@ -0,0 +1,40 @@
...
+
+export NUODB_LOGDIR NUODB_CFGDIR NUODB_VARDIR NUODB_BINDIR NUODB_RUNDIR
+
+# attempt to retain the previous crash directory (within the configured window to avoid filling the disk)
+if [ -f $NUODB_CRASHDIR/* ]; then
+  crashcount=$(find $NUODB_LOGDIR/crash-* -type d -cmin -$OVERWRITE_WINDOW ! -path . | wc -l)
```
sivanov-nuodb: @acabrele Have you got an example of how the log directory content will look like after engine crash multiple times?
I thought that the core files are located in `$NUODB_LOGDIR/crash/core.nuodb.*` or `$NUODB_CRASHDIR`. Did you mention something about the core file overwriting because of the same name?

acabrele: If logPersistence is not enabled this log-volume will be created as it used to be, so the core would still go there, just still be lost on crashloop.

So the overwrite happens inside $NUODB_LOGDIR/crash/.  This moves any contents that exist to $NUODB_LOGDIR/crash-_date_/ allowing a new core to create in $NUODB_LOGDIR/crash/ without overwriting if the process ID was the same.  I've noticed in a crash loop the PID does change.  The issue would occur when someone deletes or scales down/up the container in an attempt to fix an issue.  In that case the re-start is likely to used the same PID as the original crash and then overwrite the interesting one.

stable/admin/templates/statefulset.yaml
```diff
@@ -198,3 +209,28 @@ spec:
       resources:
         requests:
           storage: {{ .Values.admin.persistence.size }}
+{{- if .Values.admin.logPersistence.enabled }}
+  - metadata:
+      name: log-volume
```
sivanov-nuodb: How many volumes is this creating per admin deployment? I see the log-volume mounted also in the init container. Is it reused?

stable/admin/files/nuoadmin
acabrele: For ReadWriteMany, I needed to prioritise the contrail stuff so will come back to that.  Hoping the NuoDB solution will be along soon to address it too.  Ultimately if it can't support GCP it is not useful for the immediate requirement.

stable/admin/templates/statefulset.yaml
acabrele: That logic is unchanged, I guess it is just re-used between them...

stable/admin/files/nuoadmin
NikTJ777: ... I thought we used to put a date-time stamp on core file names.
- Is that no longer the case?

acabrele: This is the example from WeLab
/var/log/nuodb/crash/core.nuodb.sm-database-wlab-aws0-t24-1.63

stable/admin/values.yaml
```diff
@@ -18,7 +18,7 @@ cloud:
   # cluster name is used to make resources unique in multi-cluster configurations.
   # If the NuoDB domain spans 2 or more physical clusters, then each cluster must have a unique clusterName.
   # The default is fine for single-cluster domains.
-  clusteName: cluster0
+  clusterName: cluster0
```
NikTJ777: So - previous changes indicate that you haven't merged this branch with master yet,

I am puzzled by this change.
Did we back away fro the new structure of having `cloud.cluster.name`?

Or is this change just a result of not having merged with master?

Or is this change an accidental result of a merge?

stable/admin/values.yaml
```diff
@@ -80,6 +80,20 @@ admin:
       - ReadWriteOnce
     # storageClass: "-"
 
+  ## Enable persistent log volumes to retain logs when an external logging solution is not used.
+  logPersistence:
+    enabled: false
+    overwriteBackoff:
+      # Copies of the crash directory are taken to avoid overwrites of root crash.
+      # This controls the window within which no more copies of the crash directory will be taken to avoid the disk filling.
+      # Default will retain 3 copies within the last 120 minutes, after which they will resume.
```
NikTJ777: I don't understand what "they will resume" is intended to mean - but I assume "after which the directories will be overwritten and/or cleaned up"?

stable/admin/templates/statefulset.yaml
NikTJ777: Every container in a pod sees the same disks.

One common usage of an initContainer is to precondition the volumes that will be used by the main container.

The SM initContainer ensures that the archive mount has the correct permissions, for example.

stable/admin/values.yaml
acabrele: The other branch is not yet merged with master, it's still on 4.0.4 will tidy it up when I do that one

stable/admin/values.yaml
acabrele: No this won't ever overwrite or clear up the backups, since we don't want to risk losing the root dump.  It will just slow down disk filling up with a rapid crashloop.  This expects disk space is managed for persistent volumes in the same way it should be for archives and backup volumes...

stable/database/README.md
```diff
@@ -238,6 +238,12 @@ The following tables list the configurable parameters of the `database` chart an
 | `persistence.storageClass` | Storage class for volume backing database archive storage | `-` |
 | `configFilesPath` | Directory path where `configFiles.*` are found | `/etc/nuodb/` |
 | `configFiles.*` | See below. | `{}` |
+| `sm.logPersistence.enabled` | Whether to enable persistent storage for logs | `false` |
+| `sm.logPersistence.overwriteBackoff.copies` | How many copies of the crash directory to keep within windowMinutes | `3` |
+| `sm.logPersistence.overwriteBackoff.windowMinutes` | The window within which to keep the number of crash copies | `120` |
+| `sm.logPersistence.accessModes` | Volume access modes enabled (must match capabilities of the storage class) | `ReadWriteOnce` |
+| `sm.logPersistence.size` | Amount of disk space allocated for log storage | `20Gi` |
```
NikTJ777: If we are expecting core files to be saved, then I think 20Gi will usually be too small.

I would suggest a minimum default would be 60Gi.
Customers can increase that if they have increased the process MEM and/or increased the crashBackoff.copies.

stable/database/values.yaml
```diff
@@ -158,6 +158,20 @@ database:
...
+    ## Enable persistent log volumes to retain logs when an external logging solution is not used.
+    logPersistence:
+      enabled: false
+      overwriteBackoff:
+        # Copies of the crash directory are taken to avoid overwrites of root crash.
+        # This controls the window within which no more copies of the crash directory will be taken to avoid the disk filling.
+        # Default will retain 3 copies within the last 120 minutes, after which they will resume.
+        copies: 3
+        windowMinutes: 120
+      size: 20Gi
```
NikTJ777: Suggest 60Gi be the minimum, if we are expecting core files.

stable/admin/values.yaml
acabrele: updated for clarity

stable/database/README.md
acabrele: Updated

PR #55, opened 2020-01-29
Add option to dump cores if specific test fails

sivanov-nuodb:
Improve the debuggability for hung database engines.
- Add a way to `get diagnostics-info` including code dumps from running processes

*Use case*
If a diagnose is needed to be collected before the database teardown following code should be added after `testlib.StartDatabase()`:

```
testlib.AddTeardown(testlib.TEARDOWN_DATABASE, func() { testlib.GetDiagnoseOnTestFailure(t, namespaceName, admin0) })
```


test/testlib/minikube_utilities.go
```diff
@@ -315,6 +316,32 @@ func AwaitDatabaseUp(t *testing.T, namespace string, podName string, databaseNam
...
+		k8s.RunKubectl(t, options, "exec", podName, "--", "nuocmd", "get", "diagnose-info",
+			"--include-cores", "--output-dir", "/tmp")
+		diagnoseFile, err := k8s.RunKubectlAndGetOutputE(t, options, "exec", podName, "--", "bash", "-c", "ls -1 /tmp | grep diagnose-")
+		assert.NilError(t, err, "Can not find diagnose archive")
+
+		k8s.RunKubectl(t, options, "cp", podName+":/tmp/"+diagnoseFile, filepath.Join(targetDirPath, diagnoseFile))
+		if shouldPrintToStdout() {
+			// The file can be recovered via xxd -r -p a.hex a.bin
+			output, _ := exec.Command("hexdump", "-ve", `16/1 "%02x " "\n"`, filepath.Join(targetDirPath, diagnoseFile)).CombinedOutput()
+			t.Logf("%s:\n%s", diagnoseFile, string(output))
```
vegichan: I am a bit worried that this dump will be bigger than the allowed Travis log size, but we can deal with that if it happens.

sivanov-nuodb: Yes ... I hope that we won't need cores all the time. The good thing is that everything goes to the `results` directory at least for local test execution. At some point, there was a talk about S3 storage so maybe we can do it in future if needed.

PR #57, opened 2020-01-30
allow a user to override admin services suffixes

kmabda:
allow a user to have more controler over the name of the clusterip and
load balancer admin services name by overriding
admin.serviceSuffix.balancer and admin.serviceSuffix.clusterip if
desired, default for these suffixes is blancer and clusterip
respectively

Signed-off-by: Abdallah Chatila <abdallah@kaloom.com>


vegichan:
Abdallah, thank you for making this change.

The new values have to be reflected in the [Admin Chart README](https://github.com/nuodb/nuodb-helm-charts/blob/master/stable/admin/README.md). See the _admin.*_ section

kmabda:
Updated the Admin Chart README file.

Thanks Martin

acabrele:
For consistency, we should look to make the same change to the services for the TEs in the database chart

PR #63, opened 2020-02-07
Allow integration with NuoDB CI/CD

vegichan:
Stop using `dep ensure` as part of the travis build. Use `go get` instead. This locks down the packages that are being used.

Run `dep ensure` and update the `Gopkg.lock` file with a configuration that is known to work. Work around a version problem with `gnostic`.

Remove remaining uses of `github.com/stretchr/testify/assert` from the tests. Depend on `gotest`.

Add a way to inject nuodb version into the tests via `InjectTestVersion.yaml`. Fix tests that can not handle that correctly.


vegichan:
I decided to put the `vendor` directory into the repo. Following the guidelines here:
https://github.com/golang/dep/blob/master/docs/FAQ.md#should-i-commit-my-vendor-directory
it seems to be the best way forward.

PR #65, opened 2020-02-13
Enable eviction of unresponsive processes and bump version back up to 4.0.4

adriansuarez:


vegichan:
This change will not be sufficient to guarantee that the tests pass. It will also require a new await in the test framework to guarantee that the Admin has noticed that all engine processes actually exited.

adriansuarez:
Updated my branch to fix a test that was asserting on `nuoadmin` command-line and added a `nuocmd check database` invocation to the database teardown.

stable/admin/values.yaml
```diff
@@ -109,7 +109,16 @@ admin:
...
+  options:
+      # this high reconnect timeout is necessary because an admin can be
+      # rescheduled on a different Node and a reconnecting engine will not
+      # connect to it until the stale DNS entry in its local cache expires,
+      # which has a default TTL of 30 seconds
+      pendingReconnectTimeout: 60000
+      # this is necessary because killed engine Pods may not generate a TCP_RST
+      # packet to close the admin connection, leaving a database process entry
+      # in the domain state that has no running Pod associated with it
+      processLivenessCheckSec: 15
```
adriansuarez: Now that we are not relying on this to make the tests pass, this should be bumped to something higher like 30 seconds.

test/testlib/nuodb_database_utilities.go
```diff
@@ -55,6 +56,14 @@ func GetExtractedOptions(options *helm.Options) (opt ExtractedOptions) {
 	return
 }
 
+func EnsureDatabaseNotRunning(t *testing.T, adminPod string, opt ExtractedOptions, kubectlOptions *k8s.KubectlOptions) {
+	// invoke shutdown database; this may fail if the database is already NOT_RUNNING, which is okay
+	k8s.RunKubectlAndGetOutputE(t, kubectlOptions, "exec", adminPod, "--", "nuocmd", "shutdown", "database", "--db-name", opt.DbName)
+	// wait for all database processes to exit
+	output, err := k8s.RunKubectlAndGetOutputE(t, kubectlOptions, "exec", adminPod, "--", "nuocmd", "check", "database", "--db-name", opt.DbName, "--num-processes", "0", "--timeout", "30")
+	assert.NilError(t, err, output)
```
vegichan: use `k8s.RunKubectl` for both. You do not need the output, since you are not parsing it. You also do not need the error message (`RunKubectlE`), since you are calling NilError on it, without additional parsing.

`RunKubectl` is this under the hood:
`require.NoError(t, RunKubectlE(t, options, args...))`

adriansuarez: It is good to have the output of `nuocmd check database` so that the assert generates a useful message. I'll update the `nuocmd shutdown database` invocation to use `k8s.RunKubectlE()`.

vegichan: you can see the output in the logs regardless of which function you use

adriansuarez: Before I added the output to the assert I had no clue why the test was failing. Even with the knowledge that the command logging is somewhere else in this massive ream of log messages, I still have to look for it. It saves human time to just have the message in the assert. Is there any reason not to do that in `k8s.RunKubectl()`?

vegichan: Thats a good argument for using `k8s.RunKubectlE()`. I buy it.

Again, we do not own the code for `k8s.RunKubectl()`

adriansuarez: Ah, okay. I'll just make the change you suggested and then we can think about adding our own helper functions to generate better assert messages.

PR #66, opened 2020-02-14
Remove OpenShift specific TE DeploymentConfig

vegichan:
NuoDB does not use any of the DeploymentConfig specific features. This part of the product comes from Kubernetes 1.9 days from before Deployments were K8s GA. Remove this.


butson:
Reading up on the differences between Deployments and DeploymentConfig.  I would not be surprised if some customers would rather use DeploymentConfig.   It has problems, specifically when a node is taken out of service the pod running on that node will not be replaced on another node - but instead wait for the node to come back in service.   However, DeploymentConfig supports rollback where a broken update to the configuration can be rolled back to previous state.  while Deployment does not support this.

acabrele:
I don't fully understand the customer implications.  However I did read that deployment configs use replication controllers and deployments use replica sets.  Replica sets supercede replication controllers and the docs here suggest not to use them anymore https://kubernetes.io/docs/concepts/workloads/controllers/replicationcontroller/.  I'd vote to move them to incubator then they are available if needed?

PR #70, opened 2020-02-21
Add privileged:true for OpenShift environments

vegichan:
Add a new flag `openshift.enabled` to all Charts. When enabled, all Kubernetes objects in that chart will be created with security context `privileged`. This is required in OpenShift to mount hostPath and execute some of the operations that a DB requires.

Add a script to start a minishift cluster. This can not be added to Travis as minishift tries to start a VM and does not work with vm-driver=none as compared to minikube. There steps can be followed by anyone who wants to run the exising test suite against minishift.

Refactor tests to support both minikube and minishift. They can also be run against a real cluster.

Add verifyTeardown to all tests. Some tests were not calling all the expected teardowns. Fix all the leaks found this way.

Decreases the DB startup timeout back to 300. Travis has a 10min timeout which kills the whole build.

### DIFF
--- a/stable/admin/README.md
+++ b/stable/admin/README.md
@@ -147,17 +147,8 @@ The following tables list the configurable parameters for the `openshift` option
 | Parameter | Description | Default |
 | ----- | ----------- | ------ |
 | `enabled` | Enable OpenShift features | `false` |
-| `enableDeploymentConfigs` | Prefer DeploymentConfig over Deployment |`false`|
-| `enableRoutes` | Enable OpenShift routes | `true` |
 
-For example, to enable an OpenShift integration, and enable routes:
-
-```yaml
-openshift:
-  enabled: true
-  enableDeploymentConfigs: false
-  enableRoutes: true
-```
+Enabling OpenShift will automatically run all Kubernetes objects with elevated `privileges`.
 
 #### admin.*
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,14 +11,21 @@ go:
 env:
   jobs:
     - TEST_SUITE=basic
+
     - TEST_SUITE=minikube-short REQUIRES_MINIKUBE=true
     - TEST_SUITE=minikube-long REQUIRES_MINIKUBE=true
+
+    # does not work due to KVM/Virtualbox limitations of Travis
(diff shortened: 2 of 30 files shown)


stable/admin/README.md
```diff
@@ -147,17 +147,8 @@ The following tables list the configurable parameters for the `openshift` option
...
 
-For example, to enable an OpenShift integration, and enable routes:
-
-```yaml
-openshift:
-  enabled: true
-  enableDeploymentConfigs: false
-  enableRoutes: true
-```
+Enabling OpenShift will automatically run all Kubernetes objects with elevated `privileges`.
```
acabrele: Can readme include the expected oc commands to make this work (oc adm policy add-scc-to-user...)

butson:
running privilege is against best practices.  I doubt customers will want this.   You should avoid running a container with privileges not make easier to run as.

butson:
It is not a requirement that you run privilege to mount host path.  See SELinux chcon

acabrele:
As per slack chat, do we want to merge this or just leave it as a branch to use when OCP deploy is needed by someone, while we work out a better / alternative way which doesn't need privileged:true.  Even if we do use that somewhere, we should at least have a new service account and include in the charts the serviceAccountName.

butson:
We had worked out OCP and not needing privilege 2-3 years ago when we first
worked with openshift.  You need to under SELinux and chcon.   If you are
going to mount host path it needs to be from a privilege directory'  which
has been marked as such with chcon.  It should not be on the docker
container.

Dirk Butson

Principal Consultant
www.nuodb.com

p: +1 (617) 500-0001 <http://www.nuodb.com/company/office-locations>
t: @nuodb
<https://twitter.com/nuodb>l: linkedin.com/company/NUODB
<https://www.linkedin.com/company/NUODB>

The information contained in this email is confidential and proprietary and
intended for the sole use of the intended recipient. If the reader of this
message is not the intended recipient, you are hereby notified that any
dissemination, distribution or copying of this communication is strictly
prohibited. If you have received this email in error, please notify the
sender and delete all copies.

vegichan:
abandoning this approach. We have a better solution.

PR #73, opened 2020-03-02
Allow TEs to be defined as a StatefulSet

adriansuarez:
Add databases.te.enableStatefulSet option that allows TEs to be defined
as a StatefulSet. Assuming the databases.te.dbServices.enabled is not
false, TEs will acquire FQDNs, which are advertised by the Admin and
also appear in the CN attribute of the TE certificate, allowing hostname
verification to be enabled in SQL clients.

Also update README.md to include instructions on running tests.

### DIFF
--- a/stable/database/templates/statefulset.yaml
+++ b/stable/database/templates/statefulset.yaml
@@ -608,3 +608,185 @@ spec:
   {{- end }}
 {{- end }}
 {{- end }}
+
+{{- if .Values.database.te.enableStatefulSet }}
+---
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  labels:
+    app: {{ template "database.fullname" . }}
+    group: nuodb
+    database: {{ .Values.database.name }}
+    domain: {{ .Values.admin.domain }}
+    chart: {{ template "database.chart" . }}
+    release: {{ .Release.Name | quote }}
+  name: te-{{ template "database.fullname" . }}
+spec:
+  serviceName: {{ .Values.database.name }}
+  replicas: {{ .Values.database.te.replicas }}
+  selector:
+    matchLabels:
+      app: {{ template "database.fullname" . }}
+      component: te
+  strategy:
+    type: RollingUpdate
+  template:
+    metadata:
+      labels:
+        app: {{ template "database.fullname" . }}
+        group: nuodb
+        database: {{ .Values.database.name }}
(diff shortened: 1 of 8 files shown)


stable/database/templates/statefulset.yaml
```diff
@@ -608,3 +608,185 @@ spec:
   {{- end }}
 {{- end }}
 {{- end }}
+
+{{- if .Values.database.te.enableStatefulSet }}
+---
+apiVersion: apps/v1
+kind: StatefulSet
```
adriansuarez: This is identical to the Deployment defined in `stable/database/templates/deployment.yaml` except for the changes I am calling out here.

1) Change Deployment to StatefulSet.

stable/database/templates/statefulset.yaml
```diff
@@ -608,3 +608,185 @@ spec:
...
+  labels:
+    app: {{ template "database.fullname" . }}
+    group: nuodb
+    database: {{ .Values.database.name }}
+    domain: {{ .Values.admin.domain }}
+    chart: {{ template "database.chart" . }}
+    release: {{ .Release.Name | quote }}
+  name: te-{{ template "database.fullname" . }}
+spec:
+  serviceName: {{ .Values.database.name }}
```
adriansuarez: 2) Specify service name.

Maybe this should be conditionalized on `.Values.database.te.dbServices.enabled`.

stable/database/templates/statefulset.yaml
```diff
@@ -608,3 +608,185 @@ spec:
...
+          - "--labels"
+          - "{{- range $opt, $val := . }} {{$opt}} {{$val}} {{- end}}"
+        {{- end }}
+        {{- range $opt, $val := .Values.database.te.otherOptions }}
+          - "--{{$opt}}"
+          - "{{$val}}"
+        {{- end }}
+        {{- if or (not (hasKey .Values.database.te.dbServices "enabled")) .Values.database.te.dbServices.enabled }}
+          - "--resolve-address"
+        {{- end}}
```
adriansuarez: 3) Explicitly advertise the FQDN of the TE.

adriansuarez:
I will abandon this review and keep these changes on the shelf (in `asuarez/te-fqdn`). I agree with @vegichan that we should not increase the testing matrix by adding more variants (whether StatefulSet is better than Deployment is a discussion for another time).

PR #75, opened 2020-03-04
Add SCC for NuoDB Helm in OpenShift

vegichan:
- Add `deploy/nuodb-scc.yaml` as an example scc for a deployment to use.
- Change documentation and explain how to deploy nuodb in OCP. See `stable/README.md`
- Minor changes to documentation.
- Rewrite unit tests and run them against OCP 3.11, 4.2 and 4.3. These can not be automated in Travis.


scripts/ci/install_deps.sh
```diff
@@ -1,34 +1,69 @@
...
 
-# get the image to speed up tests
-docker pull nuodb/nuodb-ce:"${NUODB_VERSION}"
-docker pull nuodb/nuodb-ce:latest
+  # disable THP to match minikube
+  kubectl create namespace nuodb
+  kubectl -n nuodb create serviceaccount nuodb
+  oc apply -f deploy/nuodb-scc.yaml -n nuodb
+  oc adm policy add-scc-to-user nuodb-scc system:serviceaccount:nuodb:nuodb -n nuodb
+  oc adm policy add-scc-to-user nuodb-scc system:serviceaccount:nuodb:default -n nuodb
```
vegichan: @acabrele I specify the user here, not in the nuodb-scc.yaml file

stable/README.md
```diff
@@ -60,41 +60,29 @@ export HELM_HOME=`pwd`/.helm
 export KUBECONFIG=`pwd`/.kube/config
 ```
 
-### Create a Project
-
-We're limiting the Tiller server to manage infrastructure in the project (namespace) where it is installed. So the TILLER_NAMESPACE equals the namespace where NuoDB will be installed.
```
sivanov-nuodb: There are two approaches that we normally use with vanilla K8s:
- Namespace scoped Tiller installation that should work for all charts by StorageClass
- System scoped Tiller installation that was used for StorageClass chart (see `test/scripts/prepare.sh`)

I look like we are removing the steps for namespace scoped tiller installation with these changes. Are we saying that we are going to support only system scoped Tiller installation?

stable/README.md
```diff
@@ -105,41 +93,49 @@ Client: &version.Version{SemVer:"v2.14.1", GitCommit:"618447cbf203d147601b4b9bd7
...
-$ oc get pods
-NAME                      READY     STATUS    RESTARTS   AGE
-nodejs-example-1-build    1/1       Running   0          1m
-tiller-86c4495fcc-rhq97   1/1       Running   0          15m
+kubectl -n nuodb create serviceaccount nuodb
 ```
 
-Now that we've seen Helm deploy an application successfully, we're confident in its installation and configuration, lets now move on to deploying NuoDB. But first, lets clean up Node.js:
+Next, you will want to give your new service account the correct SecurityContextConstraints to run NuoDB.
+You can find the recommended SecurityContextConstraints in `deploy/nuodb-scc.yaml`
```
sivanov-nuodb: Maybe include a link to the file?
```([deploy/nuodb-scc.yaml](deploy/nuodb-scc.yaml)```

File: stable/README.md [hunk 4]
```diff
@@ -105,41 +93,49 @@ Client: &version.Version{SemVer:"v2.14.1", GitCommit:"618447cbf203d147601b4b9bd7
...
 
 ```bash
-$ helm del --purge nodejs-ex
+oc apply -f deploy/nuodb-scc.yaml -n nuodb
+oc adm policy add-scc-to-user nuodb-scc system:serviceaccount:nuodb:nuodb -n nuodb
```
sivanov-nuodb: My understanding is that we don't use `nuodb` service account currently.

stable/README.md
```diff
@@ -105,41 +93,49 @@ Client: &version.Version{SemVer:"v2.14.1", GitCommit:"618447cbf203d147601b4b9bd7
...
 
-## Deploying NuoDB using Helm Charts
+Here is a short list of NuoDB charts and their privilege requirements.
+
+| Charts | Privilege | Short Explanation |
+| ----- | ----------- | ------ |
+| transparent-hugepage| allowHostDirVolumePlugin: true | To mount hostPath and disable THP on host|
+| transparent-hugepage| volumes.hostPath | To mount hostPath and disable THP on host|
+| transparent-hugepage| seLinuxContext.* | To mount hostPath and disable THP on host|
+| admin, database| allowedCapabilities.FOWNER | To change directory ownership in PV to the nuodb process|
```
sivanov-nuodb: According to http://man7.org/linux/man-pages/man7/capabilities.7.html this capability will:
```
Bypass permission checks on operations that normally require
                the filesystem UID of the process to match the UID of the
                file (e.g., chmod(2), utime(2)), excluding those operations
```
My understanding is that we need this only because we execute `chmod 770` in the init containers to prep the volume directories which doesn't fully correspond to what is written in the README.

According to https://stackoverflow.com/questions/43544370/kubernetes-how-to-set-volumemount-user-group-and-file-permissions there are several ways to make sure that the container processes can access files in a volume. Of of the ways is specifying `fsGroup`. It looks like we went down the road with init container which needs the `FOWNER` capability in OCP. I guess that if we decide to change the templates in the charts this will be done in another PR.

stable/admin/README.md
```diff
@@ -283,7 +254,7 @@ Verify the pods are running:
 ```bash
 $ kubectl get pods
 NAME                           READY     STATUS    RESTARTS   AGE
-admin-nuodb-0                         1/1       Running   0          29m
+admin-nuodb-0                  1/1       Running   0          29m
```
sivanov-nuodb: The pod name is different nowadays. We added cluster id to the name.

stable/database/README.md
```diff
@@ -413,21 +389,12 @@ $ nuocmd delete database --db-name demo
 $ nuocmd show domain
 ```
 
-Then you must also clear the PVCs:
```
sivanov-nuodb: The PVCs are not automatically cleaned, aren't they?

test/minikube/minikube_tls_admin_test.go
```diff
@@ -34,15 +34,16 @@ func verifyKeystore(t *testing.T, namespace string, podName string, keystore str
 }
 
 func TestKubernetesTLS(t *testing.T) {
+	if testlib.IsOpenShiftEnvironment(t) {
```
sivanov-nuodb: Do we still have these problems with the new SCC?

scripts/ci/install_deps.sh
```diff
@@ -1,34 +1,69 @@
...
+  kubectl version
 
-# get the image to speed up tests
-docker pull nuodb/nuodb-ce:"${NUODB_VERSION}"
-docker pull nuodb/nuodb-ce:latest
+  # disable THP to match minikube
+  kubectl create namespace nuodb
+  kubectl -n nuodb create serviceaccount nuodb
+  oc apply -f deploy/nuodb-scc.yaml -n nuodb
+  oc adm policy add-scc-to-user nuodb-scc system:serviceaccount:nuodb:nuodb -n nuodb
```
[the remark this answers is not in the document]
acabrele: Yes this is a mixture of the two solutions, no need to create the nuodb SA if we are not also adding the possibility to define the serviceAccountName in the charts / values

scripts/ci/install_deps.sh
acabrele: OK, but it's an extra step for demos.  Maybe when we add serviceAccountName in values it could be set in the SCC.

stable/README.md
acabrele: @Stanimir - this does work, I just checked my notes and I did try a few combinations (setting runAsUser to 0 and fsGroup to 1000 both work).  At the time I moved onto allowing FOWNER in the SCC and setting specifically on the init container as I was trying to limit the changes to where they were needed (and I thought that would be preferable to running as root).  You can't set fsGroup for the init container only and as noted in chat we don't expose the ability to set runAsUser or any security context separately/differently in the init container and container.  We also don't currently have the securityContext.fsGroup logic in the admin chart, only the DB chart.

So this is definitely an option as part of the wider changes in a later PR (Martin raise a Jira already), I'll make sure to add it to the doc too.

stable/database/README.md
acabrele: They aren't, and neither are the PV's...

test/minikube/minikube_tls_admin_test.go
acabrele: Wasn't this a 3.11 specific issue?

stable/database/values.yaml
```diff
@@ -328,5 +328,3 @@ openshift:
   enabled: false
   # generally avoid deployment configs, favor deployments, so leave this as false
   enableDeploymentConfigs: false
-  # if openshift is enabled, then also permit routes by default
-  enableRoutes: true
```
acabrele: Do we care about the stuff in incubator still having refs to routes and routes.yaml files present

stable/README.md
acabrele: Is it relevant for the purposes of our repo - what we are giving is guidance on an option of how to set up helm, not supporting it?  In reality if a customer is using Helm they will likely already have it installed with the correct permissions, following some form of internal guidance... This will also go away when we go to 3.0

scripts/ci/install_deps.sh
vegichan: it is hard to do, if you cant predict the namespace. It is <namespace>:default and <namespace>:<service account>.

I moved it out so that it is not tied to any particular namespace

stable/README.md
vegichan: For OpenShift I want to have Tiller running globally. The THP should also be running just once, not once for every project. So I changed the wording in this doc to reflect that. I basically made the opposite recommendation.

stable/README.md
vegichan: We do. See above. `kubectl -n nuodb create serviceaccount nuodb`

stable/admin/README.md
vegichan: I am not going to fix all the docs with this review. There are many things that need to be changed.

stable/database/README.md
vegichan: PVCs are automatically cleared out when you delete the statefulset. The PVs are deleted based on the reclaim policy. This line was incorrect.

test/minikube/minikube_tls_admin_test.go
vegichan: it might be specific to 3.11. I am still investigating. It has nothing to do with the SCC. It is it a bug in 3.11. I have not retested in 4.3 yet.

stable/database/values.yaml
vegichan: the routes still exist in the incubator. I removed it from here, since there are no routes in the database.

test/minikube/minikube_tls_admin_test.go
sivanov-nuodb: Maybe we can add a comment in the test why we are skipping them for OCP.

scripts/ci/install_deps.sh
acabrele: Something to do later on, it can be tied to values namespace / serviceAccountName

stable/README.md
acabrele: You are creating it, but what I think Stanimir meant is that we don't then use it (at the moment).  To use it, the charts would need to supply serviceAccountName.  For the moment we are relying on the SCC being applied to default...

stable/database/README.md
acabrele: Yes our reclaim is retain...

acabrele: I don't see this behavior - the PVC is always retained and the PV deleted based on reclaim policy happens when the PVC is deleted, not the statefulset.  I just tried again in OCP, deleting with Helm and via the UI...  In Rancher I've always had to clear down the PVC's and PV's after deleting workloads.

If this wasn't the case, the PV wouldn't get re-used on redeployment of the same statefulset (e.g if you delete the PVC but not the PV, re-deploy would create new PVC and PV)

stable/README.md
vegichan: done

stable/database/README.md
vegichan: Ok. I will add it back.

scripts/ci/install_deps.sh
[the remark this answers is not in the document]
adriansuarez: I have an open PR to automatically create a service account when installing the helm charts, which is used for all the Pods. Maybe this SCC thing can be added to that service account in a declarative fashion (i.e. specified in the YAML) rather than having to do it with an explicit command.

deploy/nuodb-scc.yaml
```diff
@@ -0,0 +1,44 @@
...
+supplementalGroups:
+  type: RunAsAny
+volumes:
+  - configMap
+  - downwardAPI
+  - emptyDir
+  - hostPath
+  - persistentVolumeClaim
+  - projected
+  - secret
```
adriansuarez: From the OpenShift docs (https://docs.openshift.com/enterprise/3.0/architecture/additional_concepts/authorization.html#security-context-constraints), it looks like a declarative way to add the SCC to the service account might be something like the following:
```
users: 
- system:serviceaccount:<the namespace>:{{ .Values.nuodb.serviceAccount }}
```

acabrele: This was what was being done but Martin moved it to the doc for this as it's not deployed by the charts (else you could use the namespace passed in values / helm)

vegichan: I do not automatically deploy the SCC. The user has to do it. Because of that, I can't pass the namespace in from HELM, there is no template to fill out.

PR #77, opened 2020-03-07
Autoconfigure role and service account

adriansuarez:
Add nuodb.serviceAccount parameter to values files, and automatically
create a service account, role, and role-binding that allows NuoDB Pods
to inspect Kubernetes state.

This is needed by the Admin so that it can clean up objects in the NuoDB
domain state that are associated with resources that are known to be
defunct in Kubernetes. For example, remove a database process object
(start ID) associated with a non-running Pod or container, remove an
archive object (archive ID) associated with a non-existent PVC.

### DIFF
--- a/stable/admin/README.md
+++ b/stable/admin/README.md
@@ -114,6 +114,8 @@ The following tables list the configurable parameters for the `nuodb` option:
 | `image.tag` | NuoDB container image tag | `latest` |
 | `image.pullPolicy` | NuoDB container pull policy |`Always`|
 | `image.pullSecrets` | Specify docker-registry secret names as an array | [] (does not add image pull secrets to deployed pods) |
+| `serviceAccount` | The name of the service account used by NuoDB Pods | `nuodb` |
+| `addRoleBinding` | Whether to add role and role-binding giving `serviceAccount` access to Kubernetes APIs (Pods, PersistentVolumes, PersistentVolumeClaims, StatefulSets) | `true` |
 
 The `registry` option can be used to connect to private image repositories, such as Artifactory.
 
--- a/stable/admin/templates/role.yaml
+++ b/stable/admin/templates/role.yaml
@@ -1,0 +1,18 @@
+{{- if or (not (hasKey .Values.nuodb "addRoleBinding")) .Values.nuodb.addRoleBinding }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: nuodb-kube-inspector
+rules:
+- apiGroups:
+  - ""
+  - apps
+  resources:
+  - persistentvolumeclaims
+  - persistentvolumes
+  - pods
+  - statefulsets
+  verbs:
+  - get
+  - list
+{{- end }}
--- a/stable/admin/templates/rolebinding.yaml
+++ b/stable/admin/templates/rolebinding.yaml
@@ -1,0 +1,13 @@
+{{- if or (not (hasKey .Values.nuodb "addRoleBinding")) .Values.nuodb.addRoleBinding }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: nuodb-kube-inspector
+roleRef:
+  kind: Role
+  name: nuodb-kube-inspector
+  apiGroup: rbac.authorization.k8s.io
+subjects:
+- kind: ServiceAccount
+  name: {{ default "nuodb" .Values.nuodb.serviceAccount }}
+{{- end }}
--- a/stable/admin/templates/serviceaccount.yaml
+++ b/stable/admin/templates/serviceaccount.yaml
@@ -1,0 +1,4 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ default "nuodb" .Values.nuodb.serviceAccount }}
--- a/stable/admin/templates/statefulset.yaml
+++ b/stable/admin/templates/statefulset.yaml
@@ -27,6 +27,7 @@ spec:
         domain: {{ .Values.admin.domain }}
         group: nuodb
     spec:
+      serviceAccountName: {{ default "nuodb" .Values.nuodb.serviceAccount }}
       {{- with .Values.admin.nodeSelector }}
       nodeSelector:
 {{ toYaml . | trim | indent 8 }}
@@ -71,6 +72,10 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.name
+        - name: POD_UID
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.uid
         - name: NAMESPACE
           valueFrom:
             fieldRef:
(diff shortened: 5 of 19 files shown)


stable/admin/README.md
```diff
@@ -114,6 +114,8 @@ The following tables list the configurable parameters for the `nuodb` option:
 | `image.tag` | NuoDB container image tag | `latest` |
 | `image.pullPolicy` | NuoDB container pull policy |`Always`|
 | `image.pullSecrets` | Specify docker-registry secret names as an array | [] (does not add image pull secrets to deployed pods) |
+| `serviceAccount` | The name of the service account used by NuoDB Pods | `nuodb` |
+| `addRoleBinding` | Whether to add role and role-binding giving `serviceAccount` access to Kubernetes APIs (Pods, PersistentVolumes, PersistentVolumeClaims, StatefulSets) | `true` |
```
sivanov-nuodb: I suggest making this specific to the inspector role binding or make it a list of roles that we would like to attach to NuoDB service account. This way we could extend it in future with all roles needed for the NuoDB pods to run.

stable/admin/templates/rolebinding.yaml
```diff
@@ -0,0 +1,13 @@
...
+kind: RoleBinding
+metadata:
+  name: nuodb-kube-inspector
+roleRef:
+  kind: Role
+  name: nuodb-kube-inspector
+  apiGroup: rbac.authorization.k8s.io
+subjects:
+- kind: ServiceAccount
+  name: {{ default "nuodb" .Values.nuodb.serviceAccount }}
```
sivanov-nuodb: Is there a need to use `default` when rendering `.Values.nuodb.serviceAccount`? It has already defaulted in `values.yaml`.

stable/admin/templates/role.yaml
```diff
@@ -0,0 +1,18 @@
+{{- if or (not (hasKey .Values.nuodb "addRoleBinding")) .Values.nuodb.addRoleBinding }}
```
vegichan: this is a weird check, are there two ways to set it?

test/minikube/minikube_tls_admin_test.go
```diff
@@ -45,7 +45,7 @@ func TestKubernetesTLS(t *testing.T) {
 	defer k8s.DeleteNamespace(t, kubectlOptions, namespaceName)
 
 	defer testlib.Teardown(testlib.TEARDOWN_SECRETS)
-	
+
```
vegichan: did you somehow configure an IDE to clean these up automatically?

vegichan:
high level question: Do the SM/TEs still need to read their K8s state? I thought that we moved away from that. Only the admin should have access

stable/admin/templates/role.yaml
adriansuarez: I put this in because I want roles to be configured even if someone deletes the `addRoleBinding` field from the values file.

test/minikube/minikube_tls_admin_test.go
adriansuarez: It happens automatically with Spacemacs (Emacs distribution) with the go layer.

adriansuarez:
> high level question: Do the SM/TEs still need to read their K8s state? I thought that we moved away from that. Only the admin should have access

We want SMs and TEs to be able to include their own container ID as a process label. To do this, `nuodocker start sm/te` would use the Kubernetes API to inspect its own Pod.

stable/admin/templates/rolebinding.yaml
adriansuarez: I do this in case someone deletes the field from the values file. I'm not sure what is "best practice" in Helm, but if we expect people to edit `values.yaml` then we should be resilient to them deleting fields and explicitly inject the correct default.

stable/admin/README.md
adriansuarez: This is to support functionality that I intend to add to the Admin. I don't expect the end user to care about which APIs the NuoDB has access to; we should just enable all the ones that we use in order to do clean up actions like what I described in the summary.

stable/admin/templates/rolebinding.yaml
sivanov-nuodb: Normally the users are creating their own values files and specify them when invoking `helm install`. The `default` doesn't harm ... so I'm fine with it.

stable/admin/README.md
sivanov-nuodb: What I meant is that the service account is one and multiple roles can be bind to it. This is why I found `addRoleBinding` not specific enough to the inspector role. This is just a minor suggestion  from user point of view if you decide to change it.

test/integration/template_resources_test.go
```diff
@@ -630,3 +630,74 @@ func TestDatabaseNoBackupDisabledDaemonSet(t *testing.T) {
...
+		"      serviceAccountName: default": 1,
+	}
+	assertExpectedLines(t, &optionsMap, "admin", templateNames, &expectedLines)
+}
+
+func TestDatabaseServiceAccount(t *testing.T) {
+	optionsMap := map[string]string{}
+	templateNames := []string{
+		"templates/deployment.yaml",
+		"templates/statefulset.yaml",
```
vegichan: there is also a deploymentconfig and a daemonset. Unless I can finally get my review in and burn deployment configs :)

adriansuarez: I saw that you were getting rid of deploymentconfig and didn't want to generate merge conflicts. I don't know what the status is of daemonset.

stable/admin/templates/statefulset.yaml
```diff
@@ -71,6 +72,10 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.name
+        - name: POD_UID
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.uid
```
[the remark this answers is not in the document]
adriansuarez: Sorry, I wasn't clear in the summary/commit-message. The POD_UID is unrelated to the RBAC changes, but I thought it would be nice to have. We do not need RBAC in order to use Downward APIs (we already do it to get the Pod name).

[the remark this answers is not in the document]
adriansuarez: I noticed that as well. I just thought the Pod UID was a "nice to have", so I included it. In a Fisheye/Crucible review, I have a `nuodocker start sm/te` change to determine the container ID, which I do not think gets recycled. Assuming that Kubernetes always invoked `docker run` to create a new container as opposed to `docker start`, then every container ID should generate at most one start ID in NuoDB.

stable/admin/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: ServiceAccount
```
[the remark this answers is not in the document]
adriansuarez: Not sure what the implications are of automatically creating a role/SA, so I conditionalized it. Just about everything else seems to be conditionalized by some variable in the values files, so I figured I might as well follow that pattern.

With that said, it might be a good idea to reduce the number of permutations of parameters and define more things "by convention".

PR #80, opened 2020-03-18
Ntj/restore improvements

NikTJ777:
Primary work is to improve the `nuorestore` script to cause the sM startup to exit in the event of clearly fatal errors in the restore.

A secondary change is to add logging output to the `restore` chart to better guide the user's expectations.

Finally, some associated changes removing any dependence in the `restore` chart on any values in the `database` chart.

### DIFF
--- a/stable/admin/templates/configmap.yaml
+++ b/stable/admin/templates/configmap.yaml
@@ -42,3 +42,16 @@ metadata:
   name: {{ template "admin.fullname" . }}-nuoadmin
 data:
 {{ (.Files.Glob "files/nuoadmin").AsConfig | indent 2 }}
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  labels:
+    app: {{ template "admin.fullname" . }}
+    group: nuodb
+    domain: {{ .Values.admin.domain }}
+    chart: {{ template "admin.chart" . }}
+    release: {{ .Release.Name | quote }}
+  name: domain-{{ .Values.admin.domain }}-env
+data:
+  NUOADMIN_AP0: {{ template "admin.fullname" .}}-0
--- a/stable/database/README.md
+++ b/stable/database/README.md
@@ -227,6 +227,12 @@ The following tables list the configurable parameters of the `database` chart an
 | `configFilesPath` | Directory path where `configFiles.*` are found | `/etc/nuodb/` |
 | `configFiles.*` | See below. | `{}` |
 | `podAnnotations` | Annotations to pass through to the SM an TE pods | `nil` |
+| `autoImport.*` | Enable and configure the automatic initialising of the initial database state | `disabled` |
+| `autoImport.source` | The source - typically a URL - of the database copy to import | `""` |
+| `autoImport.credentials` | Authentication for the download of `source` in the form `user`:`password` | '""'|
+| `autoImport.stripLevels` | The number of levels to strip off pathnames when unpacking a TAR file of an archive | `1` |
+| `autoImport.type` | Type of content in `source`. One of `stream` -> exact copy of an archive; or `backupset` -> a NuoDB hotcopy backupset | 'backupset' |
+| `autoRestore.*` | Enable and configure the automatic reinitialising of a single archive in a running database - see the options in `autoImport` | `disabled` |
 | `sm.logPersistence.enabled` | Whether to enable persistent storage for logs | `false` |
 | `sm.logPersistence.overwriteBackoff.copies` | How many copies of the crash directory to keep within windowMinutes | `3` |
 | `sm.logPersistence.overwriteBackoff.windowMinutes` | The window within which to keep the number of crash copies | `120` |
@@ -301,7 +307,20 @@ Any file located in `database.configFilesPath` can be replaced; the YAML key cor
 | ----- | ----------- | ------ |
 | `nuodb.config` | [NuoDB database options][6] | `nil` |
 
-#### database.serviceSuffix
+#### restore.*
+
+The following tables list the configurable parameters of the restore chart and their default values.
+
+| Parameter | Description | Default |
+| ----- | ----------- | ------ |
+| `restore.type` | What type of restore to perform: [ "database" | "archive" ]. A "database" restore restarts the entire database at a previous state. An "archive" restore restores/repairs a SINGLE archive in a RUNNING database. | `"database"` |
+| `restore.target` | Where to restore `TO` | `{{ .Values.database.name }}` |
+| `restore.source` | Where to restore `FROM` [ backupset | url | `:latest` | `:group-latest` ] | `:latest` |
+| `restore.credentials` | Credentials to use for a URL source (user:password) | `""` |
+| `restore.autoRestart` | Whether to automatically restart the database and trigger the restore (true/false). Only valid for a "database" restore | `true` |
+
+
+### Running
 
 The purpose of this section is to allow customisation of the names of the clusterIP and balancer database services (load-balancers).
 
--- a/stable/database/files/nuosm
+++ b/stable/database/files/nuosm
@@ -97,10 +97,13 @@ function wrapLogfile() {
     tail -n $(( lines / 2 )) $LOGFILE > ${LOGFILE}-new
     rm -rf $LOGFILE
     mv $LOGFILE-new $LOGFILE
+    log "(nuosm) log file wrapped around"
   fi
-
-  log "(nuosm) log file wrapped around"
 }
+
+#=======================================
+# function to get a value from the KV store
+#
 
 #=======================================
 # function to perform archive restore
--- a/stable/database/files/nuosm
+++ b/stable/database/files/nuosm
@@ -1,11 +1,9 @@
-#!/bin/sh
+#!/bin/bash
 
-[ -n "$NUOSM_LINT" ] && set -n
+[ "$NUODB_DEBUG" = "verbose" ] && set -x
+[ "$NUOSM_VALIDATE" = "true" ] && set -e
 
 . ${NUODB_HOME}/etc/nuodb_setup.sh
-
-NUODB_DEBUG="false"
-#[ -n "$NUODB_DEBUG" ] && set -x
 
 : ${NUODB_ARCHIVEDIR:=/var/opt/nuodb/archive}
 : ${NUODB_BACKUPDIR:=/var/opt/nuodb/backup}
@@ -14,9 +12,20 @@ NUODB_DEBUG="false"
 
 : ${NUODB_IMPORT_CREDENTIALS:=:}
 
+[ -n "$NUODOCKER_KEYSTORE_PASSWORD" ] && ap_protocol="https" || ap_protocol="http"
+
+ERR_ADMIN_UNAVAILABLE=71
+ERR_BACKUPSET_NOT_FOUND=72
+ERR_NO_SPACE=73
+ERR_INCORRECT_BACKUP=74
+ERR_RESTORE_FAILED=75
+
 startup_key="/nuodb/nuosm/startup"
 
-DB_DIR=${NUODB_ARCHIVEDIR}/${NUODB_DOMAIN}/${DB_NAME}
+DB_PARENTDIR=${NUODB_ARCHIVEDIR}/${NUODB_DOMAIN}
+DB_DIR=${DB_PARENTDIR}/${DB_NAME}
+
+LOGFILE=${NUODB_LOGDIR:=/var/log/nuodb}/nuosm.log
 
 NUODB_BINDIR=$NUODB_HOME/bin
 [ -z "$NUOCMD" ] && NUOCMD="$NUODB_BINDIR/nuocmd --api-server $NUOCMD_API_SERVER"
@@ -38,62 +47,126 @@ export NUOCMD DB_NAME
 # function - report an error and exit
 #
 function die() {
-  # cleanup
-  $NUOCMD set value --key $first_req --value '' --expected-value $HOSTNAME
-
   retval=$1
   shift
+  [ -n "$wotIdid" ] && echo "While $wotIdid"
   echo "$@"
+
+  error="$@"
+  log "While $wotIdid: $error"
+
+  # cleanup
+  $NUOCMD set value --key $first_req --value '' --expected-value $HOSTNAME
+  $NUOCMD set value --key $startup_key/$DB_NAME --value '' --expected-value $HOSTNAME
+
+  # release my archive if it is locked
+  locked_archive=$( $NUOCMD show archives --db-name $DB_NAME --removed --removed-archive-format "archive-id: {id}" | sed -En "/^archive-id: / {N; /$HOSTNAME/ s/^archive-id: ([0-9]+).*$/\1/; T; p}" | head -n 1 )
+  [ -n "$locked_archive" ] && $NUOCMD create archive --db-name $DB_NAME --archive-path $DB_DIR --is-external --restored --archive-id $locked_archive
+
   exit $retval
 }
 
 #=======================================
-# function - perform archive restore
+# function - log messages
+#
+function log() {
+  if [ -n "$NUODB_DEBUG" ]; then
+    local err="$@"
+    echo "$err" | tee -a $LOGFILE
+    #curl -k -X POST -E /etc/nuodb/keys/nuocmd.pem -H "Content-type: application/json" $ap_protocol://$NUOADMIN_AP0:8888/api/1/diagnostics/log -d "{\"message\": \"nuosm $err\"}"
+  fi
+}
+
+#=======================================
+# function - trace workflow
+#
+function trace() {
+  wotIdid="$@"
+  [ -n "$1" -a -n "$NUODB_DEBUG" ] && log "trace: $wotIdid"
+}
+
+#=======================================
+# function - wrap a log file around so it doesn't grow infinitely
+#
+function wrapLogfile() {
+  logsize=$( du -sb $LOGFILE | grep -o '^ *[0-9]\+' )
+  maxlog=5000000
+  log "logsize=$logsize; maxlog=$maxlog"
+  if [ ${logsize:=0} -gt $maxlog ]; then
+    lines=$(wc -l $LOGFILE)
+    tail -n $(( lines / 2 )) $LOGFILE > ${LOGFILE}-new
+    rm -rf $LOGFILE
+    mv $LOGFILE-new $LOGFILE
+  fi
+
+  log "(nuosm) log file wrapped around"
+}
+
+#=======================================
+# function to perform archive restore
+#
+# If the restore is successful:
+# * the archive dir contents will have been copied in;
+# * the archive dir metadata will have been reset;
+# * a new corresponding archive object will have been created in Raft.
+#
+# On error, any copy of the replaced archive is retained.
 #
 function perform_restore() {
 
   retval=0
   error=
 
+  log "Restoring $restore_source; existing archives: $( ls -l $DB_PARENTDIR )"
+
   # bail out early if the restore will obviously fail
-  if [ -z "$( echo $restore_source | grep '^[a-z]\+:/[^ ]\+')" ]; then
-    [ -d "$NUODB_BACKUPDIR/$restore_source" ] || error="Backupset $restore_source cannot be found in $NUODB_BACKUPDIR"
+  if [ -z "$( echo $restore_source | grep '^[a-z]\+:/[^ ]\+')" -a ! -d "$NUODB_BACKUPDIR/$restore_source" ]; then
+      error="Backupset $restore_source cannot be found in $NUODB_BACKUPDIR"
+      retval=$ERR_BACKUPSET_NOT_FOUND
+      return $retval
   fi
 
   # work out available space
   archSize="$(du -s $DB_DIR | grep -o '^ *[0-9]\+')"
   archSpace="$(df --output=avail $DB_DIR | grep -o ' *[0-9]\+')"
 
-  if [ $(( archSize * 2 < archSpace)) ]; then
+  if [ $(( archSpace > archSize * 2)) ]; then
     saveName=$DB_DIR-save-$( date +%Y%m%dT%H%M%S )
+    undo="mv $saveName $DB_DIR"
     mv $DB_DIR $saveName
      
     retval=$?
     if [ $retval -ne 0 ]; then
-      mv $saveName $DB_DIR
+      $undo
       error="Error moving archive in preparation for restore"
+      return $retval
     fi
   else
     tarfile=$DB_DIR-$( date +%Y%m%dT%H%M%S ).tar.gz
-    tar czf $tarfile $DB_DIR
+    tar czf $tarfile -C $DB_PARENTDIR $DB_NAME
 
     retval=$?
     if [ $retval -ne 0 ]; then
       rm -rf $tarfile
       error="Restore: unable to save existing archive to TAR file"
+      return $retval
     fi
 
     archSpace="$(df --output=avail $DB_DIR | grep -o ' *[0-9]\+')"
     if [ $(( archSize + 1024000 > archSpace )) ]; then
       rm -rf $tarfile
-      retval=-1
+      retval=$ERR_NO_SPACE
       error="Insufficient space for restore after archive has been saved to TAR."
+      return $retval
     fi
 
+    undo="tar xf $tarfile -C $DB_PARENTDIR"
     rm -rf $DB_DIR
   fi
 
   mkdir $DB_DIR
+
+  log "(restore) recreated $DB_DIR; atoms=$( ls -l $DB_DIR/*.atm | wc -l)"
 
   # find the backup set and backup id
   # backupset=$(ls -tl $BACKUPDIR | head -n 1)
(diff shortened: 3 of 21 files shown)


stable/restore/values.yaml
```diff
@@ -56,15 +56,25 @@ admin:
...
 
-database:
-  # name
-  # NuoDB Database name.  must consist of lowercase alphanumeric
-  # characters '[a-z0-9]+'
-  name:
-
 restore:
+
+  ## Provide a name in place of prometheus-operator for `app:` labels
```
sivanov-nuodb: "prometheus-operator" ?

stable/restore/values.yaml
```diff
@@ -56,15 +56,25 @@ admin:
...
+  # restore type: [ database | archive ]
+  # default value is "database"
+  #
+  # A database restore restores ALL archives of the database - ie restarts the database at a previous state
+  # An archive restore restores/repairs a single archive in a RUNNING database
+  type: "database"
+
   # name of the existing database being restored 
-  target: ""
+  target: "demo"
```
sivanov-nuodb: This doesn't correspond to the README.md text. We have default of `{{ .Values.database.name }}` in there.

stable/restore/templates/_helpers.tpl
```diff
@@ -133,14 +134,18 @@ Import ENV vars from configMaps
...
 {{- define "restore.envFrom" }}
-envFrom: [ configMapRef: { name: {{ .Values.database.name }}-restore } {{- range $map := .Values.restore.envFrom.configMapRef }}, configMapRef: { name: {{$map}} } {{- end }} ]
+envFrom: [ configMapRef: { name: {{ include "restore.target" . }}-restore } {{- range $map := .Values.restore.envFrom.configMapRef }}, configMapRef: { name: {{$map}} } {{- end }} ]
 {{- end -}}
 
 {{/*
 Return the restore target.
 */}}
 {{- define "restore.target" -}}
+{{- if .Values.database -}}
```
sivanov-nuodb: It looks like you make it possible to inject `database.name` as `restore.target`. Is that a good idea having the fact that you removed the `database.name` from the chart values?

stable/database/README.md
```diff
@@ -307,6 +307,19 @@ The following tables list the configurable parameters for the `database` option
 | ----- | ----------- | ------ |
 | `nuodb.config` | [NuoDB database options][6] | `nil` |
 
+#### restore.*
```
sivanov-nuodb: It looks like this section is specific to restore chart. Why we are documenting it in the database chart readme?

NikTJ777: We need to decide on this point.

A number of the chart READMEs seem to be a copy of the others - so that causes sections not covered directly by a specific chart to be documented in the README for that chart.

It is certainly true that all values are _valid_ in any chart, but there is a very strong case for only documenting values for a chart if that value is used by that chart.

stable/restore/templates/_helpers.tpl
NikTJ777: This allows a user to include `database.name` in the restore chart, and have that used automatically as the `restore.target` default.

The template effectively checks whether the user has defined the `database.name` in their `values.yaml`, so it is certainly safe.

I am happy to have a discussion on whether that is a good idea...

stable/database/README.md
vegichan: I was thinking about it this way:
- if a chart uses a value, it should be documented in the README.

These documents should be fairly standalone. When you wanna install via Rancher, you can scroll in the top window and get a quick looksie at this doc and it really helps to have everything there.

stable/database/files/nuosm
```diff
@@ -39,6 +39,11 @@ export NUOCMD DB_NAME
 function die() {
   # cleanup
   $NUOCMD set value --key $first_req --value '' --expected-value $HOSTNAME
+  $NUOCMD set value --key $startup_key/$DB_NAME --value '' --expected-value $HOSTNAME
```
vegichan: this comment applies to the file, not to this line.

When I was doing the 24x7 training, I noticed that if the admin is whacked, this script exits with a really cryptic message. The SM itself or nuodocker has more reasonable logging and can be more helpful.

That said, I would like to add some early sanity checks to this file before trying to set values in the KV store. Maybe it is work for another day. 

It is impossible to debug any failures in this script, unless you are REALLY familiar with nuocmd.

stable/restore/templates/configmap.yaml
```diff
@@ -2,11 +2,11 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   labels:
-    app: {{ template "database.fullname" . }}
+    app: {{ template "restore.fullname" . }}
```
vegichan: good catch

stable/restore/values.yaml
```diff
@@ -56,15 +56,25 @@ admin:
...
+  ##
+  #fullnameOverride: ""
+
+  # restore type: [ database | archive ]
+  # default value is "database"
+  #
+  # A database restore restores ALL archives of the database - ie restarts the database at a previous state
+  # An archive restore restores/repairs a single archive in a RUNNING database
+  type: "database"
+
```
vegichan: are all of these valid? It is sometimes hard to tell whether stuff is commented out because it is there for guidance, or whether it is left behind

stable/restore/values.yaml
```diff
@@ -56,15 +56,25 @@ admin:
...
+  ## Provide a name in place of prometheus-operator for `app:` labels
+  ##
+  #nameOverride: ""
+
+  ## Provide a name to substitute for the full names of resources
+  ##
+  #fullnameOverride: ""
+
+  # restore type: [ database | archive ]
+  # default value is "database"
```
vegichan: it is clear what the default is from line 74 :)

stable/restore/templates/_helpers.tpl
```diff
@@ -18,17 +18,18 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
 Create a default fully qualified app name.
 We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
 */}}
-{{- define "database.fullname" -}}
+{{- define "restore.fullname" -}}
 {{- $domain := default "domain" .Values.admin.domain -}}
 {{- $cluster := default "cluster0" .Values.cloud.clusterName -}}
```
acabrele: You'll hit merge conflict here, now cluster.name

stable/database/README.md
acabrele: In this one re are also missing doco on the autoImport and autoRestore options

stable/database/files/nuosm
```diff
@@ -104,6 +114,13 @@ function perform_restore() {
     echo "curl -k --user '**' $restore_source | tar xzf - --strip-components $strip_levels -C $DB_DIR"
     curl -k --user "$restore_credentials" $restore_source | tar xzf - --strip-components $strip_levels -C $DB_DIR
 
+    retval=$?
```
acabrele: We are missing handling of the restore type being a backupset _or_ stream in this download - I know that was not the purpose of this PR but we need to cover it off at some point!

stable/database/files/nuosm
NikTJ777: I have added a test for the admin layer readiness, and exit immediately with an explanatory message that includes whatever error `nuocmd` returned.

Your comments regarding familiarity with `nuocmd` are reasonable, and confirm that `nuosm` relies as much as possible on the existing (relatively low-level) API of `nuocmd`.

I have added some additional diagnostics to the error message.

I am also adding a new file of SM startup tests that verify the correct operation of the script in a number of failure scenarios.

stable/restore/values.yaml
NikTJ777: Yes, I believe that `nameOverride` and `fullNameOverride` are both valid for this chart.

And both `database` and `archive` restore types are valid and supported.

stable/database/files/nuosm
NikTJ777: I've merged the stream/backupset restore code into `perform_restore()` to ensure all restores are handled consistently.

stable/database/README.md
NikTJ777: Yup - good catch.

README updated - will be included in the next push.

stable/database/files/nuosm
```diff
@@ -189,74 +371,102 @@ if [ "$restore_source" = ":latest" ]; then
...
+      [ -n "$error" ] && log "WARNING: Error performing SEED restore: $error"
     fi
   fi
 
   # wait until it's my turn to startup
+  trace "waiting until it's my turn to start up"
+  retry=0
   until owner=$( $NUOCMD get value --key $startup_key/$DB_NAME ); [ "$owner" = "$HOSTNAME" -o "$NUODB_SEQUENCE_SYNC" = "false" ] ; do
 
+    [ "$owner" = "" -a $retry -gt 0 ] && die $ERR_RESTORE_FAILED "Fatal error in database RESTORE - initial SM has exited with error - $HOSTNAME aborting also"
```
sivanov-nuodb, line 461: What happens if previous SM fails to start after it has been started by `nuodocker`? Possible reasons could be no raft leader, misconfiguration, invalid archive, journal recovery problems, etc. Similar problem is reported in HELM-185.
We will have an owner as `$startup_key/$DB_NAME` is never cleaned but just "passed" between SMs. It is cleaned only in `die` function which is explicitly called if restore scripts fail somewhere.

As discussed before, maybe we should add a timeout while waiting for the last obtained `owner_id` to go into RUNNING. If this timeout, then we can `die` and wait for k8s to restart the pod.

stable/database/files/nuosm
```diff
@@ -1,22 +1,31 @@
-#!/bin/sh
+#!/bin/bash
 
-[ -n "$NUOSM_LINT" ] && set -n
+[ "$NUODB_DEBUG" = "verbose" ] && set -x
+[ "$NUOSM_VALIDATE" = "true" ] && set -e
```
sivanov-nuodb, line 4: If we exit immediately, then we are not clearing some important metadata as `first_req` and `$startup_key/$DB_NAME`. Would that cause some problems?

stable/database/files/nuosm
```diff
@@ -1,22 +1,31 @@
-#!/bin/sh
+#!/bin/bash
 
-[ -n "$NUOSM_LINT" ] && set -n
+[ "$NUODB_DEBUG" = "verbose" ] && set -x
```
sivanov-nuodb, line 3: I couldn't find if this is set by default or not. It might be better to have all debug messages printed by default for easier troubleshooting.

stable/admin/templates/configmap.yaml
```diff
@@ -41,4 +41,17 @@ metadata:
     release: {{ .Release.Name | quote }}
   name: {{ template "admin.fullname" . }}-nuoadmin
 data:
-{{ (.Files.Glob "files/nuoadmin").AsConfig | indent 2 }}
\ No newline at end of file
+{{ (.Files.Glob "files/nuoadmin").AsConfig | indent 2 }}
+---
+apiVersion: v1
```
vegichan, line 46: whats this configmap for?

test/integration/script_lint_test.go
```diff
@@ -0,0 +1,100 @@
+package integration
```
vegichan, line 1: what does this do?

test/minikube/minikube_base_database_test.go
```diff
@@ -236,9 +232,14 @@ func restoreDatabase(t *testing.T, namespaceName string, podName string, databas
 	restore := func() {
 		testlib.InjectTestValues(t, options)
 		helm.Install(t, options, testlib.RESTORE_HELM_CHART_PATH, restName)
-		testlib.AddTeardown(testlib.TEARDOWN_RESTORE, func() { helm.Delete(t, options, restName, true) })
+		defer helm.Delete(t, options, restName, true)
```
vegichan, line 235: whats the difference between deleting it here and deleting it later?

test/minikube/minikube_sm_test.go
```diff
@@ -0,0 +1,291 @@
...
+	"github.com/stretchr/testify/assert"
+
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/gruntwork-io/terratest/modules/helm"
+	"github.com/gruntwork-io/terratest/modules/k8s"
+	"github.com/gruntwork-io/terratest/modules/random"
+)
+
+func TestKubernetesStartSM(t *testing.T) {
```
vegichan, line 27: please dont add tests/code that never worked!

test/minikube/minikube_sm_test.go
```diff
@@ -0,0 +1,291 @@
...
+				"backup.persistence.enabled":            "true",
+				"backup.persistence.size":               "1Gi",
+				"database.env[0].name":                  "NUODB_DEBUG",
+				"database.env[0].value":                 "debug",
+			},
+		}
+
+		databaseChartName := testlib.StartDatabase(t, namespaceName, admin0, &databaseOptions)
+
+		fmt.Println("startDatabase returned...")
```
vegichan, line 58: t.log

test/minikube/minikube_sm_test.go
```diff
@@ -0,0 +1,291 @@
...
+
+		opt := testlib.GetExtractedOptions(&databaseOptions)
+		smPodTemplate := fmt.Sprintf("sm-%s-nuodb-%s-%s", databaseChartName, opt.ClusterName, opt.DbName)
+		sm0 := testlib.GetPodName(t, namespaceName, smPodTemplate)
+
+		fmt.Println("extracted databaseOptions; sm name is ", sm0)
```
vegichan, line 64: t.log

test/minikube/minikube_sm_test.go
```diff
@@ -0,0 +1,291 @@
...
+
+		kubectlOptions := k8s.NewKubectlOptions("", "", namespaceName)
+
+		testlib.AddDiagnosticTeardown(
+			testlib.TEARDOWN_DATABASE,
+			t,
+			func() { testlib.GetFile(t, namespaceName, sm0, "/var/log/nuodb", "nuosm.log") },
```
vegichan, line 71: what does this do?

test/minikube/minikube_sm_test.go
```diff
@@ -0,0 +1,291 @@
...
+		)
+
+		defer func() {
+			status, _ := k8s.RunKubectlAndGetOutputE(t, kubectlOptions, "exec", admin0, "--", "nuocmd", "show", "database", "--db-name", opt.DbName)
+			fmt.Println("database status", status)
```
vegichan, line 76: t.log. But you dont really need it. RunKubectlAndGetOutput prints everything to the log anyways. You will be printing everything twice.

test/minikube/minikube_sm_test.go
```diff
@@ -0,0 +1,291 @@
...
+		}()
+
+		t.Run("verifyRestore", func(t *testing.T) { verifyRestorePasses(t, namespaceName, sm0, "archive", ":latest") })
+		t.Run("verifyNonexstentRestore", func(t *testing.T) { verifyRestoreFails(t, namespaceName, sm0, "archive", "urgle-furgle") })
+
+		// have to wait for the SM to be running again...
+		t.Run("fixArchive1", func(t *testing.T) { healDatabase(t, namespaceName, sm0, admin0, &databaseOptions) })
+		t.Run("verifynonexistentRestoreURL", func(t *testing.T) { verifyRestoreFails(t, namespaceName, sm0, "archive", "urgle://furgle/gurgle") })
+		t.Run("fixArchive2", func(t *testing.T) { healDatabase(t, namespaceName, sm0, admin0, &databaseOptions) })
+		// t.Run("verifyEmptyRestore", func(t *testing.T) {
```
vegichan, line 86: no commented out code in master

test/minikube/minikube_sm_test.go
```diff
@@ -0,0 +1,291 @@
...
+				"backup.persistence.size":               "1Gi",
+				"database.autoRestore.source":           ":latest",
+				"database.env[0].name":                  "NUODB_DEBUG",
+				"database.env[0].value":                 "debug",
+			},
+		}
+
+		databaseChartName := testlib.StartDatabase(t, namespaceName, admin0, &databaseOptions)
+
+		fmt.Println("startDatabase returned...")
```
vegichan, line 162: t.log

test/minikube/minikube_sm_test.go
```diff
@@ -0,0 +1,291 @@
...
+		restart := func() {
+			k8s.RunKubectl(t, kubectlOptions, "exec", sm0, "--", "rm", "-rf", "/var/opt/nuodb/archive/nuodb/demo")
+
+			result, _ := testlib.RunSQL(t, namespaceName, admin0, "demo", "create table user.newtable (id bigint)")
+			t.Log(result)
+		}
+
+		defer func() {
+			status, _ := k8s.RunKubectlAndGetOutputE(t, kubectlOptions, "exec", admin0, "--", "nuocmd", "show", "database", "--db-name", opt.DbName)
+			fmt.Println("database status on EXIT", status)
```
vegichan, line 185: t.log. Probably not required at all

test/minikube/minikube_sm_test.go
```diff
@@ -0,0 +1,291 @@
...
+
+		t.Run("verifyRestartOnLostArchive", func(t *testing.T) {
+			testlib.AwaitDatabaseRestart(t, namespaceName, admin0, "demo", &databaseOptions, restart)
+
+			status, _ := k8s.RunKubectlAndGetOutputE(t, kubectlOptions, "exec", sm0, "--", "ls", "-l", "/var/opt/nuodb/archive/nuodb/demo")
+			t.Log("after archive delete: ", status)
+		})
+	})
+
+	/*
```
vegichan, line 196: no commented out code in master

test/minikube/minikube_sm_test.go
```diff
@@ -0,0 +1,291 @@
...
+
+}
+
+func healDatabase(t *testing.T, namespaceName string, podName string, adminName string, databaseOptions *helm.Options) {
+	opts := testlib.GetExtractedOptions(databaseOptions)
+	kubectlOptions := k8s.NewKubectlOptions("", "", namespaceName)
+
+	restoreArchive(t, namespaceName, podName, "archive", ":latest")
+	err := k8s.RunKubectlE(t, kubectlOptions, "exec", adminName, "--", "nuocmd", "shutdown", "database", "--db-name", opts.DbName)
+	assert.NoError(t, err, "'nuocmd shutdown database' failed")
```
vegichan, line 244: you probably want `require.NoError`. assert will not terminate the test here and will go into a 5 min wait on the next line. I am assuming that you want to throw an exception here and terminate the test.

test/testlib/minikube_utilities.go
```diff
@@ -756,6 +788,35 @@ func GetAdminEventLog(t *testing.T, namespace string, podName string) {
 	)
 }
 
+func GetFile(t *testing.T, namespace string, podname string, path string, filename string) error {
```
vegichan, line 791: is this for some logs that are not published by k8s? And you have to retrieve them from disk?

test/testlib/minikube_utilities.go
```diff
@@ -544,13 +544,27 @@ func GetDatabaseIncarnation(t *testing.T, namespace string, podName string, data
...
+// TODO remove databaseNae param - and retrieve it from extracted databaseOptions
 func AwaitDatabaseRestart(t *testing.T, namespace string, podName string, databaseName string, databaseOptions *helm.Options, restart func()) {
 	incarnation := GetDatabaseIncarnation(t, namespace, podName, databaseName)
 
 	restart()
 
 	Await(t, func() bool {
-		return GetDatabaseIncarnation(t, namespace, podName, databaseName).Major > incarnation.Major
+		newIncarnation := GetDatabaseIncarnation(t, namespace, podName, databaseName)
+		t.Log("incarnation ->", newIncarnation)
```
vegichan, line 566: do you really need this?

```
TestKubernetesRestoreDatabase 2020-11-06T06:32:38Z logger.go:66: Running command kubectl with args [--namespace testkubernetesrestoredatabase-nrbqi3 exec admin-nrbqi3-nuodb-cluster0-0 -- nuocmd --show-json get databases]
TestKubernetesRestoreDatabase 2020-11-06T06:32:41Z logger.go:66: {
TestKubernetesRestoreDatabase 2020-11-06T06:32:41Z logger.go:66:   "incarnation": {
TestKubernetesRestoreDatabase 2020-11-06T06:32:41Z logger.go:66:     "major": 2, 
TestKubernetesRestoreDatabase 2020-11-06T06:32:41Z logger.go:66:     "minor": 0
TestKubernetesRestoreDatabase 2020-11-06T06:32:41Z logger.go:66:   }, 
TestKubernetesRestoreDatabase 2020-11-06T06:32:41Z logger.go:66:   "name": "demo", 
TestKubernetesRestoreDatabase 2020-11-06T06:32:41Z logger.go:66:   "state": "AWAITING_ARCHIVE_HISTORIES_MSG", 
TestKubernetesRestoreDatabase 2020-11-06T06:32:41Z logger.go:66:   "uri": "https://localhost:8888/api/1/databases/demo"
TestKubernetesRestoreDatabase 2020-11-06T06:32:41Z logger.go:66: }
    minikube_utilities.go:566: incarnation -> &{2 0}
```

test/testlib/minikube_utilities.go
```diff
@@ -544,13 +544,27 @@ func GetDatabaseIncarnation(t *testing.T, namespace string, podName string, data
...
+
+	var restartCount int32
+	for _, status := range pod.Status.ContainerStatuses {
+		restartCount += status.RestartCount
+	}
+
+	return restartCount
+}
+
+// TODO remove databaseNae param - and retrieve it from extracted databaseOptions
```
vegichan, line 558: typo

test/testlib/minikube_utilities.go
```diff
@@ -544,13 +544,27 @@ func GetDatabaseIncarnation(t *testing.T, namespace string, podName string, data
 	return nil
 }
 
+func GetRestartCount(t *testing.T, podName string, options *k8s.KubectlOptions) int32 {
```
vegichan, line 547: we already have `GetPodRestartCount`. How is this different?

stable/admin/templates/configmap.yaml
NikTJ777: This config map allows any container to find the address of the zeroeth (admin...-0) pod - hereafter AP-0

So...
I want to push logging messages to the admin layer, using the PUT diagnostics/log endpoint.
However, those log lines are only stored by the admin that received them. So if you send them via the k8s service, the messages will get scattered randomly across all APs in the domain.

This configmap allows the container to address the AP-0 container directly, so that all the log messages are sent to the same AP.
This solution is not perfect:
* PRO: The AP-0 is the first to be started, and the last to be shut down, so it is more likely to be there than any other AP.
* PRO: All such messages can ALWAYS be found in the AP-0 logs;
* CON: if the AP-0 has some problem, it is possible for the admin layer to be up, but the AP-0 to be unavailable.
   ... However, this is also true if we used a different algorithm that identified the AP that this engine pod used as its peer

The only other viable algorithm I could think of is to ping the admin ipcluster service, and capture the returned address.
However, that returns an address and not a hostname, so if the AP restarts during the process, then the logging would fail and be lost.

test/integration/script_lint_test.go
NikTJ777: It performs a "lint" test of all scripts in the helm charts.

The "lint" test is performed by running each found script with the '-n' option.

test/minikube/minikube_base_database_test.go
NikTJ777: Well, clearly we want the 'helm.delete', and 'kubectl delete' to occur directly one after the other.
This code avoids the dependence on the testlib, which is a good thing.
I agree I should have added a comment at the time as to why I used 'defer' rather than a teardown.

My memory is that I chose what I felt was the simplest solution.
In this case, defer works fine, so there was no need to use a teardown.

test/minikube/minikube_sm_test.go
NikTJ777: In my defence, I did say on slack that I was going to perform a cleanup to this PR prior to re-requesting peer review.

This test is particularly difficult in an environment where there is only ever 1 SM.
Causing an error in an SM causes the database to shut down if there is only 1 SM - and it turns out there are so many ways to get to a state where the code tries to clean up, and the database is in some conflicting state - such as NOT_RUNNING, or similar.


I _will_ invest another 2-3 hours trying to make this test work by making the nested tests independent, so there is no need to heal the database.

Otherwise, I will remove this test.
I have a copy on disk, and I can try getting it to work as a separate exercise.

test/minikube/minikube_sm_test.go
NikTJ777: I used Println because it put the message somewhere where I could more easily find it.

I am happy to change/remove all Println calls prior to requesting the final review.

test/minikube/minikube_sm_test.go
NikTJ777: I added logging to nuosm to write to a separate file.

Basically, I needed to see the series of events over multiple SM restarts.
The existing code was only copying specific files, and this was the simplest way to add a new log file in a different location.

It is also possible to push all the nuosm logging to the AP layer, and then to diagnose issues by grepping in the AP-0 log.

I believed that the generic ability to capture any file was useful.

test/minikube/minikube_sm_test.go
NikTJ777: I was adding context, so I could trace the evolution of the database state.

test/minikube/minikube_sm_test.go
NikTJ777: Yup - I will clean all commented code prior to requesting the new review.

It is really good to have a list here in the PR.

test/minikube/minikube_sm_test.go
NikTJ777: Great catch - thank you.

I thought I had found and changed all the assert -> require in the new code...

test/testlib/minikube_utilities.go
NikTJ777: Yes, and Yes.

I am happy to discuss moving this logging to the AP layer, and hence the AP-0 logfile.

For my immediate diagnostics, having a separate file made it quicker to understand the behaviour.

test/testlib/minikube_utilities.go
NikTJ777: I was getting a timeout on the incarnation number changing - so I added logging to understand the value at each attempt, rather than just the final value.

I will be removing this line, now that the test is passing reliably.

test/testlib/minikube_utilities.go
NikTJ777: :)  TODO fix typo in TODO  :)

stable/admin/templates/configmap.yaml
vegichan: I believe that @db2882 and @adriansuarez want to unify what nuodocker does and send all operations to a single admin instance.

test/integration/script_lint_test.go
```diff
@@ -0,0 +1,100 @@
...
+		testlib.DATABASE_HELM_CHART_PATH,
+		testlib.RESTORE_HELM_CHART_PATH,
+		testlib.THP_HELM_CHART_PATH,
+		testlib.YCSB_HELM_CHART_PATH,
+	}
+
+	shebang, err := regexp.Compile(shebangPattern)
+	if err != nil {
+		t.Error("Cannot compile patterh:", shebangPattern)
+	}
```
vegichan, line 30: require.NoError(...

test/integration/script_lint_test.go
```diff
@@ -0,0 +1,100 @@
...
+				continue
+			}
+
+			fmt.Println("groups=", groups)
+			cmd := exec.Command(groups[1], "-n", scriptPath)
+
+			out, err := cmd.CombinedOutput()
+			if err != nil {
+				errorList = append(errorList, fmt.Sprintln(scriptPath, "LINT failed:", err, "\n", string(out)))
+			}
```
vegichan, line 97: just replace this and the collection/defer/reporting piece with `assert.NoError(t, err, <your-error-message>)`

test/integration/script_lint_test.go
```diff
@@ -0,0 +1,100 @@
...
+	"regexp"
+	"strings"
+	"testing"
+
+	"github.com/nuodb/nuodb-helm-charts/test/testlib"
+)
+
+const shebangPattern = "[#][!][ \t]*/[^/]+/(sh|bash)"
+
+func TestLintScripts(t *testing.T) {
```
vegichan, line 18: please unify all the logging here (t.log, fmt, log, ...) t.log(

test/testlib/minikube_utilities.go
NikTJ777: It is different in that it takes the options from the caller directly.
However, my code doesn't use that to get different behaviour.

I can check the history to see if I just missed the existing GetPodRestartCount, or whether it simply did not exist when I wrote mine, and has been implemented since.

test/integration/script_lint_test.go
NikTJ777: Sure.

And there is a typo in there as well!

Double value!! :)

test/integration/script_lint_test.go
NikTJ777: Sure - that is more elegant :)

test/minikube/minikube_base_database_test.go
vegichan: I do not follow. We always try to minimize the number of changes and the code churn. Every change should be motivated by a real improvement.

So my question is:
- why is the delete required now and why was it not required before? Deleting a chart deletes all k8s objects from that chart.

test/minikube/minikube_sm_test.go
vegichan: oh. So I am reviewing this too early! Sorry!

stable/restore/values.yaml
NikTJ777: Yup - an ***old*** copy-n-paste error.

The "prometheus-operator" string is in the stable/database/values.yaml file as well.

Presumably, the original database chart was copied from the public prometheus operator...

I'll fix it here - we should have a separate PR to fix it in database.

stable/restore/values.yaml
NikTJ777: Thanks - fixing...

test/testlib/constants.go
```diff
@@ -45,3 +50,15 @@ const MINIMAL_VIABLE_ENGINE_MEMORY = "500Mi"
 const K8S_EVENT_LOG_FILE = "kubernetes_event.log"
 
 const ALWAYS_RUN_DIAGNOSTIC_TEARDOWNS = "ALWAYS_RUN_DIAGNOSTIC_TEARDOWNS"
+
+// Ok-Ko enum
+type OkKo int
```
adriansuarez, line 55: Can we name this something more descriptive, like `Result`?

test/testlib/minikube_utilities.go
```diff
@@ -756,6 +788,35 @@ func GetAdminEventLog(t *testing.T, namespace string, podName string) {
...
+	pwd, err := os.Getwd()
+	require.NoError(t, err)
+
+	fromPath := fmt.Sprintf("%s/%s:%s/%s", namespace, podname, path, filename)
+	dirPath := filepath.Join(pwd, RESULT_DIR, namespace, podname)
+	toPath := filepath.Join(dirPath, filename)
+
+	_ = os.MkdirAll(dirPath, 0700)
+
+	f, err := os.Create(toPath)
```
adriansuarez, line 801: We have to explicitly create the file before running `kubectl cp` with that file as the destination? Or is this just to truncate the file? Either way, I'm not sure why we would defer the call to `f.Close()` rather than executing eagerly.

vegichan:
This has been superceded by #174

PR #81, opened 2020-03-23
skip doc builds

vegichan:
Skip travis builds on a doc only change
- the first build of a branch is always build
- subsequent builds are only build if the change contains a non-doc change


vegichan:
we have tested this in the operator repo. works as expected

PR #82, opened 2020-03-26
Update README.md and add helper functions script

adriansuarez:
- Update README.md to include instructions on running tests.
- Add script containing helper functions to enable local testing with
Minikube.


README.md
```diff
@@ -78,3 +78,13 @@ The charts in the `Incubator` directory are those that do not meet these criteri
...
 This project is still under active development, so you might run into [issues](https://github.com/nuodb/nuodb-helm-charts/issues). If you do, please don't be shy about letting us know, or better yet, contribute a fix or feature.
+
+## Testing
+
+To run the tests in this project, Go 1.12 is required. This project uses GOPATH to manage dependencies. The easiest way to clone the repository and run a test suite (e.g. `test/integration`) is as follows:
+
+```
+go get github.com/nuodb/nuodb-helm-charts
+cd $(go env GOPATH)/src/github.com/nuodb/nuodb-helm-charts
+go test ./test/integration
```
vegichan: also `-v` and `-timeout 60`.

This will only run 1/3 of the tests. And skip all the minikube tests.

vegichan:
Look into
```
func InjectTestVersion(t *testing.T, options *helm.Options) {
	dat, err := ioutil.ReadFile(INJECT_FILE)```

`const INJECT_FILE = "../../versionInject.yaml"`

It might already do part of what you are trying to achieve

File: scripts/helpers [hunk 2]
```diff
@@ -0,0 +1,55 @@
...
+    if [ "$NUODB_IMAGE_REPO" != "" ] && [ "$NUODB_IMAGE_TAG" != "" ]; then
+        helm install "$@" \
+            --set nuodb.image.repository="$NUODB_IMAGE_REPO" \
+            --set nuodb.image.tag="$NUODB_IMAGE_TAG" \
+            --set nuodb.image.pullPolicy=IfNotPresent \
+            --set busybox.image.pullPolicy=IfNotPresent
+    else
+        helm install "$@" \
+            --set nuodb.image.pullPolicy=IfNotPresent \
+            --set busybox.image.pullPolicy=IfNotPresent
```
vegichan: I was considering making this change to the repo in general.

Most other helm charts I've seen use IfNotPresent

adriansuarez:
> Look into
> 
> ```
> func InjectTestVersion(t *testing.T, options *helm.Options) {
> 	dat, err := ioutil.ReadFile(INJECT_FILE)
> ```
> 
> `const INJECT_FILE = "../../versionInject.yaml"`
> 
> It might already do part of what you are trying to achieve

I'm not trying to run tests in Go with a custom image (yet). What I'm trying to do is manually test a NuoDB image in my local repository. For development, I use the `build/docker/scripts/update-image.sh` script to update a specific file in an existing NuoDB image. I'm trying to create a similar workflow with the Helm charts in Minikube.

adriansuarez:
FYI @vegichan, it looks like the CI build on the branch failed because I did a force push: https://travis-ci.org/github/nuodb/nuodb-helm-charts/jobs/667420325?utm_medium=notification&utm_source=email

```
$ set -e
  # fail loudly when force-pushed
  MODIFIED_FILES=$(git diff --name-only $TRAVIS_COMMIT_RANGE)
  # waiting for native solution https://github.com/travis-ci/travis-ci/issues/6301
  if [ -z "${MODIFIED_FILES}" ]; then
    # $TRAVIS_COMMIT_RANGE will be empty for builds triggered by the initial commit of a new branch.
    echo "No changes found, can not determine what to skip"
  elif ! echo ${MODIFIED_FILES} | grep -qvE '(\.md$)'; then
    echo "Only docs were updated, stopping build process."
    exit
  fi
  
fatal: ambiguous argument '9cd87d74ee65...20c588ed0df9': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

README.md
adriansuarez: What is the 60 second timeout? It definitely takes longer than 60 seconds to run the minikube tests. Also, does that command run both the short and long variants?

vegichan: the default is 15min. It needs to be bumped to 60/120 min. See https://github.com/nuodb/nuodb-helm-charts/blob/master/scripts/ci/determine_and_run_tests.sh

scripts/helpers
acabrele: I've had comments about this too, I think this would be a good change to make.

PR #87, opened 2020-04-07
Add test exercising domain resync

adriansuarez:
Add test that exercises unreleased domain resync functionality,
specifically SM scale-down and PVC deletion. For now, this test can only
be run internally, and it also starts multiple SMs, which means that it
requires an Enterprise license. We do not want to publish an Enterprise
license in GitHub, so these tests might remain internal only even after
the domain resync functionality is released.


vegichan:
should this get merged into "nuodb/test" rather than master?

test/minikube/minikube_domain_resync_test.go
```diff
@@ -0,0 +1,193 @@
+// +build enterprise
```
vegichan, line 1: got it!

test/minikube/minikube_domain_resync_test.go
```diff
@@ -0,0 +1,193 @@
...
+			return nil, processes
+		}
+
+		if err != nil {
+			return
+		}
+
+		processes = append(processes, obj)
+	}
+}
```
vegichan, line 85: I would break this out into its own class. Same as the other JSON/YAML to Go parsers

test/minikube/minikube_domain_resync_test.go
```diff
@@ -0,0 +1,193 @@
...
+
+	helmChartReleaseName, namespaceName := testlib.StartAdmin(t, &helm.Options{}, 1, "")
+
+	admin0 := fmt.Sprintf("%s-nuodb-cluster0-0", helmChartReleaseName)
+
+	defer testlib.Teardown(testlib.TEARDOWN_DATABASE) // ensure resources allocated in called functions are released when this function exits
+
+	testlib.StartDatabase(t, namespaceName, admin0, &helm.Options{
+		SetValues: map[string]string{
+			"database.sm.resources.requests.cpu":    "0.25",
```
vegichan, line 124: other tests (see Green/Blue basic test) use "250m"

test/minikube/minikube_domain_resync_test.go
```diff
@@ -0,0 +1,193 @@
...
+	for _, archiveId := range originalArchiveVolumeClaims {
+		originalArchiveId = archiveId
+	}
+	assert.Assert(t, originalArchiveId != "")
+
+	// update replica count
+	options := k8s.NewKubectlOptions("", "")
+	options.Namespace = namespaceName
+	output, err := k8s.RunKubectlAndGetOutputE(t, options, "get", "statefulset", "-o", "jsonpath={.items[*].metadata.name}")
+	assert.NilError(t, err, output)
```
vegichan, line 143: I am not a huge fan of these "output parsing" get functions. I prefer using the K8s API to list/grab the objects. There is the `StatefulSetList` class which might be a good start.

test/minikube/minikube_domain_resync_test.go
```diff
@@ -0,0 +1,193 @@
...
+
+	// by default the hotcopy SM replica count is 1 and regular SM count is 0
+	// scale regular SM replica count up to 1
+	smStatefulSet := ""
+	for _, statefulSet := range statefulSets {
+		if strings.HasPrefix(statefulSet, "sm-") && !strings.Contains(statefulSet, "hotcopy") {
+			k8s.RunKubectl(t, options, "scale", "statefulset", statefulSet, "--replicas=1")
+			smStatefulSet = statefulSet
+		}
+	}
```
vegichan, line 156: you can achieve this by giving Helm the options for `nuodb.database.sm.noHotCopy.Replicas` or similar around lines 126

vegichan:
amazing functionality!

test/minikube/minikube_domain_resync_test.go
adriansuarez: Makes sense. That could cut down test time a bit. But it would make it harder to figure out which PVC corresponds to the scaled down StatefulSet (hotcopy). Currently the test causes PVCs to be created in an orderly fashion so that we know which is which.

test/minikube/minikube_domain_resync_test.go
adriansuarez: I swiped this from the existing `Unmarshal` function, hence the bogus `processes` variable name. You're saying promote this so that it can be used elsewhere?

test/minikube/minikube_domain_resync_test.go
adriansuarez: I wasn't able to find a `GetStatefulSet()` function, but maybe it is because my Golang code navigation in Emacs was busted.

test/minikube/minikube_domain_resync_test.go
```diff
@@ -0,0 +1,186 @@
...
+	for _, archiveId := range originalArchiveVolumeClaims {
+		originalArchiveId = archiveId
+	}
+	assert.Assert(t, originalArchiveId != -1)
+
+	// update replica count
+	options := k8s.NewKubectlOptions("", "")
+	options.Namespace = namespaceName
+
+	statefulSets := getStatefulSets(t, namespaceName).Items
```
vegichan, line 136: I think that with Helm you have a guarantee that all 3 will exist. Now that I am rewriting the tests for the Helm Operator, I noticed that sometimes the test is faster than the operator and the k8s objects are not yet created. I had to change a bunch of these into Awaits.

vegichan:
I am assuming that we will see this pass in NuoDB enterprise CI/CD eventually

db2882:
I'm not too familiar with Go nor helm testing so not sure I'm the best reviewer, but looks okay to me.

PR #89, opened 2020-04-08
Conditionalize license tests on environment variable

adriansuarez:
Conditionalize tests that exercise licensing on the NUODB_LICENSE
environment variable, which is set to ENTERPRISE if an internal build is
used that has Enterprise as its default license.


vegichan:
thanks

PR #94, opened 2020-04-13
Fix YCSB for non-cloud environments

vegichan:
Due to the `cloud.sh` script, it was not possible to run the YCSB driver in non-cloud environments (bamboo, minikube). This was causing issues in the WIP Helm Operator as it was impossible to scale the the YCSB workload past 0 replicas.

We removed the corresponding `cloud.sh` auto-tagging from the database pods in Helm Chart release 2.0.0 and as such, keeping it in YCSB was serving no purpose.

Add a simple test for YCSB.


test/minikube/minikube_e2e_demo_test.go
```diff
@@ -0,0 +1,63 @@
...
+	"fmt"
+	"github.com/gruntwork-io/terratest/modules/k8s"
+	"github.com/nuodb/nuodb-helm-charts/test/testlib"
+	"testing"
+	"time"
+
+	"github.com/gruntwork-io/terratest/modules/helm"
+)
+
+const YCSB_CONTOLLER_NAME = "ycsb-load"
```
adriansuarez, line 15: Typo: CONTOLLER

vegichan: fixed. thanks

incubator/demo-ycsb/files/ycsb_gen.sh
```diff
@@ -75,24 +75,14 @@ if [ -n "${LB_POLICY}" ]; then
...
-
 #check for nuo.properties file overrides
 if [ "$PEER_ADDRESS" != "" ] && [ "$DB_NAME" != "" ]; then
     AGENT_DB_NAME_VALUE="-p db.url=jdbc:com.nuodb://${PEER_ADDRESS}/${DB_NAME}?schema=${db_schema}"
 
     if [ "$TE_DIRECT" == "true" ]; then
         AGENT_DB_NAME_VALUE="${AGENT_DB_NAME_VALUE}&direct=true"
     else
-        AGENT_DB_NAME_VALUE="${AGENT_DB_NAME_VALUE}${LOADBALANCER_QUERY}${LOADBALANCER_POLICY}${LBVAR_POD_NAME}${LBVAR_NODE_NAME}${LBVAR_NODE_ZONE}"
+        AGENT_DB_NAME_VALUE="${AGENT_DB_NAME_VALUE}${LOADBALANCER_QUERY}${LOADBALANCER_POLICY}"
```
NikTJ777, line 85: So does this mean we cannot pass any variable on the URL to represent which ZONE (region) the YCBS client is running in - or any other "locality"?

Joe used to rely on this for his demo - can his demo now be modified to still show load-balancing across ZONE/region?

butson:
That should not have been true.   Cloud.sh should only be a helper if the data was not passed down.   And if it was not on a cloud platform or in a platform that could not detect what cloud it was on,  things just did not get set.

As far as the helm charts and how they set locality that is up to you.   Unfortunately the way the product works the client needs to tell the broker where it is in order for the broker to find out where it is.   And k8s does not have a downward api to pass down labels from the node.   The operator should have a means to get that info

incubator/demo-ycsb/files/ycsb_gen.sh
vegichan: the TEs are not labeled. So using labels to select them can not work.

What this code was doing, is that it was passing down some strings, nothing would match them, so it would fall back to `ANY`.

NikTJ777: So what you're saying is that this was already broken?

vegichan: yes.

> We removed the corresponding cloud.sh auto-tagging from the database pods in Helm Chart release 2.0.0 and as such, keeping it in YCSB was serving no purpose.

NikTJ777: So do we have any plan going forward to support ZONE and/or region load-balancing using the New Admin LB system?

- I can see a pretty easy way to support it using k8s balancers, and direct=true.

PR #96, opened 2020-04-15
Re-enable 3-admin tests

vegichan:
With many bugs fixed, we should be now able to re-enable all 3-admin tests.

Also increase the timeout for short/long. We have now reached 35-40 min per test family.


vegichan:
After a few tests were run, this can not be merged just yet. Abandoning

PR #100, opened 2020-04-19
Add test of membership resync based on Admin StatefulSet

adriansuarez:


test/minikube/minikube_domain_resync_test.go
```diff
@@ -101,6 +101,55 @@ func checkArchives(t *testing.T, namespaceName string, adminPod string, numExpec
 	return
 }
 
+func TestAdminScaleDown(t *testing.T) {
```
vegichan, line 104: should this not be skipped if this is not the ENTERPRISE test suite?

adriansuarez: Yes. The file has tag "enterprise".

test/minikube/minikube_domain_resync_test.go
```diff
@@ -101,6 +101,55 @@ func checkArchives(t *testing.T, namespaceName string, adminPod string, numExpec
...
+			"nuocmd", "show", "domain", "--server-format", "{id} {connected_state}")
+		return strings.Contains(output, admin1+" Disconnected")
+	}, 300*time.Second)
+
+	// wait for scaled-down Admin Pod to be deleted
+	testlib.AwaitNoPods(t, namespaceName, admin1)
+
+	// commit a Raft command to confirm that remaining Admin has consensus
+	k8s.RunKubectl(t, options, "exec", admin0, "--",
+		"nuocmd", "set", "value", "--key", "testKey", "--value", "testValue", "--unconditional")
```
vegichan, line 136: this should not be the case without the K8s aware module. What is the guarantee that the module processed the leaving admin in time? Would this not be flaky?

test/minikube/minikube_domain_resync_test.go
```diff
@@ -101,6 +101,55 @@ func checkArchives(t *testing.T, namespaceName string, adminPod string, numExpec
...
+
+	// admin1 is still in membership, though it is excluded from consensus;
+	// delete PVC to cause it to be completely removed from the membership;
+	// this should allow the Admin health-check to succeed
```
vegichan, line 140: I do not fully understand. My questions are:
1) should there not be an await or something to make sure that the k8s aware action was taken?
2) the comment says that health-check succeeds. Why would it fail before? If it fails, how is it possible to commit a command on line 135?

test/minikube/minikube_domain_resync_test.go
adriansuarez: The conditions for a scaled-down Admin being excluded from consensus are that (1) the StatefulSet replica count is lower than or equal to the ordinal in the server ID and (2) the Pod with that server ID (the server ID and Pod name should be identical for this behavior to be enabled) is deleted. We checked earlier in the test that Pod was deleted. Even if the surviving Admin did not receive the "Pod deleted" event by the time we invoke `nuocmd set value`, all blocking Raft commands are issued with a timeout (the defaults vary, but a typical timeout is 5 seconds. So eventually the surviving member would notice that it has consensus to commit the command and would notify the client that the command is committed.

test/minikube/minikube_domain_resync_test.go
adriansuarez: Take a look at the corresponding Fisheye/Crucible to see how the consensus override works. The consensus override is not triggered periodically or by event notifications. The Admin process that is trying to become Raft leader or replicate a command (if it already considers itself leader) will check the stateful K8s client for the conditions I mentioned above, every time that it is evaluating that there is a quorum.

test/minikube/minikube_domain_resync_test.go
vegichan: can you actively check whether a server is excluded from membership? Does it show in "show domain" or any other REST API call?

PR #101, opened 2020-04-21
Add regression testing for multi-admin deployments

adriansuarez:
This basically the PR @vegichan posted a week or two ago, but with the `verifyPodKill` variants disabled, since they are currently flaky.


adriansuarez:
Actually, I'm going to abandon this PR since it is with respect to a branch other than master.

PR #110, opened 2020-05-11
Remove version from release label + typo fixes

acabrele:
- _release_ labels included chart version info making rolling upgrades of stateful sets impossible.  This change removes the version in all labels, avoiding arbitrary changes to artefacts caused only by the chart version being upgraded, and removing the issue in statefusets.

- Minor typos corrected

- Indentation on admin options corrected

### DIFF
--- a/stable/database/values.yaml
+++ b/stable/database/values.yaml
@@ -201,7 +201,7 @@ database:
       replicas: 1
 
       # Deadline for starting a hotcopy job - if a job hasn't started in this time - give up
-      deadine: 1800
+      deadline: 1800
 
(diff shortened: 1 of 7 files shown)


stable/database/values.yaml
```diff
@@ -201,7 +201,7 @@ database:
       replicas: 1
 
       # Deadline for starting a hotcopy job - if a job hasn't started in this time - give up
-      deadine: 1800
+      deadline: 1800
```
vegichan, line 204: well spotted

vegichan:
what about the `restore`, `storageClass` and `thp`  charts?

stable/database/values.yaml
acabrele: ta!

PR #114, opened 2020-05-22
Allow a ReadWriteMany PVC for persistent log volumes in Transaction Engine Deployments

vegichan:
Similarly to Storage Engine StatefulSets and Admin Process StatefulSets, allow persistence of the log directory.

Deployments require a ReadWriteMany PV type that might not be supported on all major K8s platforms today.


vegichan:
forgot to add the doc change. will do!

stable/database/templates/deployment.yaml
```diff
@@ -105,7 +105,7 @@ spec:
           subPath: {{ $key }}
         {{- end -}}
         {{- end }}
-        - name: logdir
+        - name: log-te-volume
           mountPath: /var/log/nuodb
```
NikTJ777, line 109: I don't see how collisions by multiple TEs in the Deployment are avoided.
Would 3 different TEs write to 3 different log files - or all share the same log file?

vegichan: The only thing that a TE ever writes to disk is a core file. We don't write any other log information. The log goes to `stdout` instead. Core files have a file format dependant on both PID (not so unique) and `hostname`. If the TE crashes multiple times on the same host, with the same PID, the core generator just appends a counter at the end.

NikTJ777: Ok - thanks!

acabrele:
So the assumption being made here is that the customer will already have deployed a storage class supporting a RWM volume, and we are only implementing the ability to use it rather than also create the SC itself...  In this way we have made the deployment consistent for any underlying k8s rather than implementing the AWS / Azure RWM volumes only.

stable/database/templates/persistentvolumeclaim.yaml
```diff
@@ -37,4 +37,33 @@ spec:
...
+    domain: {{ .Values.admin.domain }}
+    chart: {{ template "database.chart" . }}
+    release: {{ .Release.Name | quote }}
+spec:
+  accessModes:
+  {{- range .Values.database.te.logPersistence.accessModes }}
+    - {{ . }}
+  {{- end }}
+  {{- if .Values.database.te.logPersistence.storageClass }}
+  {{- if (eq "-" .Values.database.te.logPersistence.storageClass) }}
```
acabrele, line 60: Do we need to allow anything other than RWM?

vegichan: I did it this way to make it more similar to SM/Admins. Put the value in the values file, rather than hide it here.

Of course, as far as we know, we can only accept RWM.

PR #116, opened 2020-06-03
Clean out unused files

vegichan:
We have accumulated many old files over the years. Delete all of them.


vegichan:
> I looked to see if any of the images were used in the README.
> 
> The logo image in the README now renders in my browser as broken - but that is not any of these images being deleted.
> 
> Overall, looks like a nice cleanup.

That is caused by the nuodb webpages migrating resources during the page rehaul. We should fix that too!

PR #119, opened 2020-06-12
Add readiness probe timeouts in values

acabrele:
I have found, depending on the type of cluster being used and the resources set for the containers, that some deployments require longer readiness timeouts.  If a readiness probe timeout is hit too many times, we can also hit a zombie thread reaping issue.

This change 

- adds a readinessTimeoutSeconds to option to the values file
- add to readme
- defaults to the existing settings for 1s in admin and 5s in engines

Related to DB-31293


vegichan:
should we make the admin timeout probes match the engine ones and set them to 5s?

acabrele:
> should we make the admin timeout probes match the engine ones and set them to 5s?

Maybe?  I've never hit the issue before in other deployments so would assume admin is _normally_ OK, Defaulting engines to longer kind of makes sense since multiple checks are being run with all remote processes (vs admin being one command, directly in the exec of the yaml, and its own raft is already local).

PR #122, opened 2020-06-24
LB policy job affinity

kmabda:
in some use cases it's desirable to control where the lb policy
kubernetes job will be scheduled.

use the nodeSelector, affinity or tolerations parameters if they are
defined in the admin chart to control on which kuberentes node the
job will run.


acabrele:
"in some use cases it's desirable to control where the lb policy kubernetes job will be scheduled."

The changes uses the existing admin values to control the job location, but that would have the longer term side effect of also causing the admin statefulset to follow the same rules.  Is that the intention?  Perhaps having new values for the job explicitly is needed?

kmabda:
> The change uses the existing admin values to control the job location, but that would have the longer term side effect of also causing the admin statefulset to follow the same rules. Is that the intention? Perhaps having new values for the job explicitly is needed?

For our use case the intention is to have the job  follow the same rules used to schedule the admin pods.

PR #126, opened 2020-07-13
[DB-30945] Add bootstrapServers label to Admin StatefulSet

adriansuarez:
Add `boostrapServers` label for the Admin StatefulSet that defines the number of servers that should be part of the initial membership of the domain, which will be generated by `nuodocker start admin`. This enables reprovisioning of a fresh admin-0 instance so it rejoins its peers. Previously, a reprovisioned admin-0 would instance consider itself the lone member of a new domain (since there is no Raft state to replay, which is the only place where added peers would be made durable) and would bootstrap a new disjoint domain.

In multi-cluster deployments, this label is only added to the Admin StatefulSet in the entrypoint cluster.

### DIFF
--- a/stable/admin/templates/statefulset.yaml
+++ b/stable/admin/templates/statefulset.yaml
@@ -7,6 +7,9 @@ metadata:
     release: {{ .Release.Name | quote }}
     group: nuodb
     domain: {{ .Values.admin.domain }}
+    {{- if (eq (default "cluster0" .Values.cloud.cluster.name) (default "cluster0" .Values.cloud.cluster.entrypointName)) }}
+    bootstrapServers: {{ default 0 .Values.admin.bootstrapServers | quote }}
+    {{- end }}
   name: {{ template "admin.fullname" . }}
 spec:
   serviceName: {{ .Values.admin.domain }}
--- a/stable/admin/values.yaml
+++ b/stable/admin/values.yaml
@@ -82,6 +82,24 @@ admin:
   # Admin Service
   replicas: 1
 
+  # If specified and set to a value greater than 0, an initial membership
+  # consisting of server IDs <statefulset name>-0, <statefulset name>-1, ...,
+  # <statefulset name>-<n-1> is defined for all Admins. Defining this to a
+  # value larger than 1 allows a bootstrap server (e.g. <statefulset name>-0)
+  # that has to be reprovisioned from scratch to rejoin its peers when it is
+  # restarted, rather than bootstrapping a new domain consisting of itself.
+  #
+  # For multi-cluster configurations, only the Admin StatefulSet in the
+  # entrypoint cluster (i.e. the one with cloud.cluster.name equal to
+  # cloud.cluster.entrypointName) can specify bootstrapServers.
+  #
+  # For backwards-compatibility, the default value is 0 so that it is disabled
+  # by default. Changing the value of bootstrapServers on a running domain,
+  # either by reinstantiating the Helm chart with a new value or by updating
+  # the bootstrapServers label on an existing Admin StatefulSet, is illegal
+  # because the domain can only be bootstrapped once.
+  bootstrapServers: 0
+
   lbPolicy: nearest
   lbQuery: random(first(label(pod ${pod:-}) label(node ${node:-}) label(zone ${zone:-}) any))
 
--- a/stable/admin/values.yaml
+++ b/stable/admin/values.yaml
@@ -93,12 +93,13 @@ admin:
   # entrypoint cluster (i.e. the one with cloud.cluster.name equal to
   # cloud.cluster.entrypointName) can specify bootstrapServers.
   #
-  # For backwards-compatibility, the default value is 0 so that it is disabled
-  # by default. Changing the value of bootstrapServers on a running domain,
-  # either by reinstantiating the Helm chart with a new value or by updating
-  # the bootstrapServers label on an existing Admin StatefulSet, is illegal
-  # because the domain can only be bootstrapped once.
-  bootstrapServers: 0
+  # Changing the value of bootstrapServers on a existing domain, either by
+  # reinstantiating the Helm chart with a new value or by updating the
+  # bootstrapServers label on an existing Admin StatefulSet, is illegal because
+  # the domain can only be bootstrapped once. When upgrading nuodb-helm-charts
+  # from a version that does not support bootstrapServers, bootstrapServers
+  # must be set to 0.
+  bootstrapServers: 1
 
   lbPolicy: nearest
   lbQuery: random(first(label(pod ${pod:-}) label(node ${node:-}) label(zone ${zone:-}) any))
--- a/test/minikube/minikube_domain_resync_test.go
+++ b/test/minikube/minikube_domain_resync_test.go
@@ -3,7 +3,9 @@
 package minikube
 
 import (
+	"encoding/json"
 	"fmt"
+	"io"
 	"strconv"
 	"strings"
 	"testing"
@@ -16,6 +18,7 @@ import (
 	"github.com/gruntwork-io/terratest/modules/helm"
 	"github.com/gruntwork-io/terratest/modules/k8s"
 	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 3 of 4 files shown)


stable/admin/values.yaml
```diff
@@ -82,6 +82,24 @@ admin:
   # Admin Service
   replicas: 1
 
+  # If specified and set to a value greater than 0, an initial membership
```
vegichan, line 85: should the default not be 0? That is the current default. Setting it to greater than 1 should solve the -0 issue

test/minikube/minikube_domain_resync_test.go
```diff
@@ -101,6 +104,86 @@ func checkArchives(t *testing.T, namespaceName string, adminPod string, numExpec
...
+
+	// delete Raft data and Pod for admin-0
+	k8s.RunKubectl(t, options, "exec", admin0, "--",
+		"bash", "-c", "rm $NUODB_VARDIR/raftlog")
+	k8s.RunKubectl(t, options, "delete", "pod", admin0)
+	// wait until the Pod is rescheduled
+	testlib.Await(t, func() bool {
+		newPod, e := k8s.GetPodE(t, options, admin0)
+		return e == nil && newPod.Status.Phase == v1.PodRunning && originalPod.GetUID() != newPod.GetUID()
+	}, 300*time.Second)
```
vegichan, line 168: this is already a utility function. for example
```
adminPod := GetPod(t, namespaceName, admin0)
k8s.RunKubectl(t, kubectlOptions, "delete", "pod", "--selector=domain=nuodb")
testlib.AwaitPodObjectRecreated(t, namespaceName, adminPod, 30*time.Second)
```

stable/admin/templates/statefulset.yaml
```diff
@@ -7,6 +7,9 @@ metadata:
     release: {{ .Release.Name | quote }}
     group: nuodb
     domain: {{ .Values.admin.domain }}
+    {{- if (eq (default "cluster0" .Values.cloud.cluster.name) (default "cluster0" .Values.cloud.cluster.entrypointName)) }}
+    bootstrapServers: {{ default 0 .Values.admin.bootstrapServers | quote }}
+    {{- end }}
```
vegichan, line 12: why not inject this value into the startup call?

Are labels mutable? Can someone upgrading from 2.3.1 use this without recreating their admin statefulset?

stable/admin/values.yaml
```diff
@@ -82,6 +82,24 @@ admin:
...
+  # <statefulset name>-<n-1> is defined for all Admins. Defining this to a
+  # value larger than 1 allows a bootstrap server (e.g. <statefulset name>-0)
+  # that has to be reprovisioned from scratch to rejoin its peers when it is
+  # restarted, rather than bootstrapping a new domain consisting of itself.
+  #
+  # For multi-cluster configurations, only the Admin StatefulSet in the
+  # entrypoint cluster (i.e. the one with cloud.cluster.name equal to
+  # cloud.cluster.entrypointName) can specify bootstrapServers.
+  #
+  # For backwards-compatibility, the default value is 0 so that it is disabled
```
vegichan, line 96: random location. If this value is specified. We should not pass `--bootstrap-server-id` to nuodocker. See `nuoadmin` script.

adriansuarez: We do not pass `--bootstrap-server-id`. I was not able to find any place where the `NUODB_BOOTSTRAP_SERVERID` is defined.

stable/admin/templates/statefulset.yaml
adriansuarez: The label is not mutable, as I mention in the comment. It can only be defined when the domain is bootstrapped for the very first time, because the initialMembership has to be defined consistently among all servers in the initialMembership (to prevent disjoint leaders).

stable/admin/values.yaml
adriansuarez: I see that there is a special `nuoadmin` script (it looks like the `nuoadmin` script in the Docker image with some extra stuff), which specifies `--bootstrap-server-id "$NUODB_BOOTSTRAP_SERVERID"`, but this will expand to `--bootstrap-server-id ""` (empty string) since it is not defined.

stable/admin/templates/statefulset.yaml
adriansuarez: The labels are not mutable as I mention in the comment in the values file.

The default value is 0 to prevent from breaking Helm chart upgrade, but I'm on the fence about whether to define the default as something else and force customers upgrading to remember to explicitly set `bootstrapServers=0`.

stable/admin/values.yaml
adriansuarez: The command-line argument expands to `--bootstrap-server-id ""`, because `NUODB_BOOTSTRAP_SERVERID` is never defined.

stable/admin/values.yaml
adriansuarez: No, the current value is equivalent to 0 because we never define a bootstrap server ID.

stable/admin/values.yaml
vegichan: Good catch. We should remove it then. Wanna do it as part of this review, or should I make a separate one?

stable/admin/values.yaml
vegichan: if we dont define it, it defaults to -0, correct? Which is logically the same as defining 1. Defining 0 is confusing. It seems to indicate that it is OK to generate separate subgroups.

stable/admin/templates/statefulset.yaml
vegichan: So, if we specified the default as 1. Someone upgrading from 2.3.1 would have to override it to 0 to make it work? That sounds like something that can be documented. Also, if you have the value set to either 0 or 1, it basically means that -0 is the bootstrap id, correct?

stable/admin/values.yaml
adriansuarez: 1 and 0 are not the same. 1 means define `admin-0` as the only bootstrap server and is equivalent to defining `NUODB_BOOTSTRAP_SERVERID=admin-0`, which we do not currently do. 0 means there is no `NUODB_BOOTSTRAP_SERVERID`.

vegichan: what happens if there is no bootstrap server?

stable/admin/templates/statefulset.yaml
adriansuarez: We do not define a bootstrap server at all currently. The fact that admin-0 ends up being the server that all others join through is only enforced by the fact that it is `NUODB_DOMAIN_ENTRYPOINT`. If `NUODB_DOMAIN_ENTRYPOINT`, then there would be nothing forcing servers to join through `admin-0` (I have another branch from a while ago that changes `NUODB_DOMAIN_ENTRYPOINT` to the load-balancer, but there were other problems that blocked it, which should be addressed now).

stable/admin/values.yaml
adriansuarez: I don't think we need to change `nuoadmin`. It should accept `NUODB_BOOTSTRAP_SERVERID`. `bootstrapServers=<n>` is just a generalization of `NUODB_BOOTSTRAP_SERVERID` that assumes K8s conventions, and both should be available.

stable/admin/templates/statefulset.yaml
vegichan: I did not know that! I always assumed that we do use bootstrap of -0.

stable/admin/values.yaml
adriansuarez: See my other comment. We have to define `NUODB_DOMAIN_ENTRYPOINT=admin-0` otherwise there is the possibility of disjoint domains being created by concurrently starting admins.

vegichan: My preference would be to stop supporting the broken model of no bootstrapID and hence do not support a model where NUODB_BOOTSTRAP_SERVERID is empty. Hence I would prefer the default of `boostrapServers` to be `1`.

stable/admin/values.yaml
vegichan: I am not talking about `./build/docker/entry/nuoadmin`. I am talking about the overridden file that is specific to K8s.

stable/admin/values.yaml
adriansuarez: Okay, but we have to make it clear to customers that when they upgrade the Helm charts on existing deployments from nuodb-helm-charts <=2.3.1, they have to explicitly set `bootstrapServers=0`.

stable/admin/values.yaml
adriansuarez: Do we really want the two to diverge further? I just looked, and the only difference is the handling of core files, which I'm not sure is even relevant:
```
$ diff -u build/docker/entry/nuoadmin ~/go/src/github.com/nuodb/nuodb-helm-charts/stable/admin/files/nuoadmin --ignore-space
--- build/docker/entry/nuoadmin	2020-01-08 18:00:25.000000000 -0500
+++ /Users/asuarez/go/src/github.com/nuodb/nuodb-helm-charts/stable/admin/files/nuoadmin	2020-03-01 22:27:30.000000000 -0500
@@ -24,6 +24,16 @@
 
 export NUODB_LOGDIR NUODB_CFGDIR NUODB_VARDIR NUODB_BINDIR NUODB_RUNDIR
 
+# attempt to retain the previous crash directory (within the configured window to avoid filling the disk)
+if [ -f $NUODB_CRASHDIR/* ]; then
+  crashcount=$(find $NUODB_LOGDIR/crash-* -type d -cmin -$OVERWRITE_WINDOW ! -path . | wc -l)
+  if [ $crashcount -lt $OVERWRITE_COPIES ]; then
+    crashbackupdir="$NUODB_LOGDIR/crash-$( date +%Y%m%dT%H%M%S )/"
+    mkdir $crashbackupdir
+    mv $NUODB_CRASHDIR/* $crashbackupdir
+  fi
+fi
+
 exec nuodocker start admin \
      --server-id "$NUODB_SERVERID" --alt-address "$NUODB_ALT_ADDRESS" --admin-port "$NUODB_ADMIN_PORT" \
      --domain-entrypoint "$NUODB_DOMAIN_ENTRYPOINT" --bootstrap-server-id "$NUODB_BOOTSTRAP_SERVERID" \
```

stable/admin/values.yaml
vegichan: So that it does not generate the new label and hence can be helm upgraded?

stable/admin/values.yaml
vegichan: good point. You are right.

stable/admin/values.yaml
adriansuarez: The issue with an already bootstrapped domain having the `bootstrapServers` value changed, which causes the Admin to complain about its configuration (nuoadmin.conf) not matching its durable state (raftlog).

stable/admin/values.yaml
```diff
@@ -82,6 +82,25 @@ admin:
...
+  # <statefulset name>-<n-1> is defined for all Admins. Defining this to a
+  # value larger than 1 allows a bootstrap server (e.g. <statefulset name>-0)
+  # that has to be reprovisioned from scratch to rejoin its peers when it is
+  # restarted, rather than bootstrapping a new domain consisting of itself.
+  #
+  # For multi-cluster configurations, only the Admin StatefulSet in the
+  # entrypoint cluster (i.e. the one with cloud.cluster.name equal to
+  # cloud.cluster.entrypointName) can specify bootstrapServers.
+  #
+  # Changing the value of bootstrapServers on a existing domain, either by
```
vegichan, line 96: great explanation! Very clear

test/minikube/minikube_rolling_upgrade_test.go
```diff
@@ -74,9 +73,10 @@ func TestKubernetesUpgradeFullDatabaseMinorVersion(t *testing.T) {
 
 	options := helm.Options{
 		SetValues: map[string]string{
-			"nuodb.image.registry": "docker.io",
+			"nuodb.image.registry":   "docker.io",
```
vegichan, line 76: what editor are you using? I am guessing that we have different settings for intents.

adriansuarez: [Spacemacs](https://github.com/syl20bnr/spacemacs) with the [Golang layer](https://develop.spacemacs.org/layers/+lang/go/README.html).

PR #131, opened 2020-07-28
add support for DIAGNOSTIC teardown funcs.

NikTJ777:
Diagnostic teardowns get called before regular teardowns.
Add further comments on teardown functions.

When troubleshooting various test errors, it can be useful to have diagnostic information published - and/or a Sleep() inserted - at the point the error is detected, but before the teardown functions have removed resources.

This PR adds a second type of teardown func - a diagnostic teardown - that is called *before* all other teardowns in that named teardown list.

To have a diagnostic teardown run in Travis, you should pass in a 'conditional' to AddDiagnosticTeardown() that will be true when the diagnostic teardown should be run - for example:

* t - the teardown will run if the test has already been marked as failed (the t.Failed() test is run at Teaardown() time)
* true - the teardown will run unconditionally - even if ALWAYS_RUN_DIAGNOSTIC_TEARDOWNS = false

@vegichan  - we may wish to change some of the existing teardowns that call GetAppLog() to be diagnostic teardowns.

Some usage examples:

* call sleep for 30 minutes *if* the test has failed
// t *testing.T
testlib.AddDiagnosticTeardown("some name", t, func() { time.Sleep(1800) } )

* save the SM log if we went past incarnation 2
saveSmLog := func() { testlib.GetFile(t, namespaceName, sm0, "/var/log/nuodb", "nuosm.log") }
// ...
majorInc := testlib.getDatabaseIncarnation(...).Major
testlib.AddDiagnosticLog("SM test", majorInc > 2, saveSmLog)


test/testlib/minikube_utilities.go
```diff
@@ -52,27 +91,37 @@ func Teardown(name string) {
...
 		for x := len(list) - 1; x >= 0; x-- {
 			list[x]()
 		}
 	}
 
-	// release all funcs in all lists
-	teardownLists = make(map[string][]func())
-
-	assert.True(t, remaining == 0, "Error - %d teardownLists were left uncleared", remaining)
+	assert.True(t, len(uncleared) == 0, "Error - %d teardownLists were left uncleared: %s", len(uncleared), uncleared)
```
vegichan, line 124: we now have `assert.Zero` or `assert.Equals` :)

test/testlib/minikube_utilities.go
```diff
@@ -29,21 +30,59 @@ import (
...
+ * See: testlib.AddTeardown(); testlib.AddDiagnosticTeardown()
  */
 func Teardown(name string) {
+	// ensure both list and diagnostic list are removed.
+	defer func() { delete(diagnosticTeardownLists, name) }()
+	defer func() { delete(teardownLists, name) }()
+
 	list := teardownLists[name]
-	delete(teardownLists, name)
+	list = append(list, diagnosticTeardownLists[name]...) // append any diagnostic funcs - so they are called FIRST
```
vegichan, line 85: I would only execute diagnostic teardown if the test has failed. That way you can just add them and they will only print something if they are actually required. You would have to use `if t.hasFailed()`

vegichan:
I have used this pattern so far:
```
// Generate diagnose in case this test fails
	testlib.AddTeardown(testlib.TEARDOWN_DATABASE, func() {
		if t.Failed() {
			testlib.GetDiagnoseOnTestFailure(t, namespaceName, admin0)
			testlib.RecoverCoresFromTEs(t, namespaceName, "demo")
		}
	})
```

test/testlib/minikube_utilities.go
NikTJ777: Good catch - thanks!
I thought I had changed them all.

test/testlib/minikube_utilities.go
NikTJ777: Is that a request to change AddDiagnosticTeardown to take a *testing.T argument, and check it before calling the diagnostic teardowns?

I have actually used diagnostic teardowns to dump the exit state to log to understand the sequence of events in situations where pods are restarted - so I would like making the t.Failed() test to be optional.

So I would be happy to take a *testing.T argument, but skip the test if the passed in *testing.T is nil.

test/testlib/minikube_utilities.go
NikTJ777: Actually - I realise that that assert.True() code is in master after the most recent updates which had changed many other of those to assert.Equal().
So I left it intentionally - assuming it had not been changed for a reason. :)

test/testlib/minikube_utilities.go
vegichan: you could override it with an environmental variable? Something like `if t.Failed() || ALWAYS_PRINT_DIAGNOSE`

test/testlib/minikube_utilities.go
vegichan: I probably missed it

test/testlib/minikube_utilities.go
NikTJ777: In the expectation that we would normally only set that variable when running tests locally - and it would always take the if t.Failed() path in Travis?

That sounds fair...

test/testlib/minikube_utilities.go
NikTJ777: I agree the error message is probably better with assert.Equal()...

test/testlib/minikube_utilities.go
NikTJ777: But as far as I can see, I still need to pass the *testing.T var as a parameter, since the *testing.T is not in the closure of the AddDiagnoseTeardown() function.

NikTJ777: Ok, I've added an EnvVar ALWAYS_RUN_DIAGNOSTIC_TEARDOWNS; plus a condition param to AddDiagnosticTeardown() to control whether diagnostic teardowns are run or not.

The condition param is a bit more flexible than just passing a *testing.T - but that also works.

test/testlib/minikube_utilities.go
```diff
@@ -29,21 +30,86 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+/** Lists of the teardown and diagnostic teardown funcs */
 var teardownLists = make(map[string][]func())
+var diagnosticTeardownLists = make(map[string][]func())
+
+/** Exported var - initialised from the EnvVar, but can be reset in code if desired */
+var AlwaysRunDiagnosticTeardowns = strings.EqualFold(os.Getenv(ALWAYS_RUN_DIAGNOSTIC_TEARDOWNS), "true")
```
sivanov-nuodb, line 38: Do we set this to `true` somewhere in `.travis.yml` file or the intuition is to overwrite it only when running tests locally?

test/testlib/minikube_utilities.go
```diff
@@ -29,21 +30,86 @@ import (
...
+			case bool:
+				shouldIdoIt = c
+
+			default:
+				shouldIdoIt = c != nil
+			}
+		}
+
+		if shouldIdoIt {
+			teardownFunc()
```
sivanov-nuodb, line 90: Why we execute the diagnostic function here when it is executed as part of the `Teardown()`? Maybe I'm overlooking something so a quick usage example will help.

vegichan:
@NikTJ777 somehow your branch is colliding with your previous dep->modules change

test/testlib/minikube_utilities.go
vegichan: the idea is to set it locally if you need it, yeah

test/testlib/minikube_utilities.go
```diff
@@ -29,21 +30,86 @@ import (
...
+ * See: testlib.AddTeardown(); testlib.AddDiagnosticTeardown()
  */
 func Teardown(name string) {
+	// ensure both list and diagnostic list are removed.
+	defer func() { delete(diagnosticTeardownLists, name) }()
+	defer func() { delete(teardownLists, name) }()
+
 	list := teardownLists[name]
-	delete(teardownLists, name)
+	list = append(list, diagnosticTeardownLists[name]...) // append any diagnostic funcs - so they are called FIRST
```
vegichan, line 112: my suggestion was to conditionalize the diagnostic teardown here. Something like:
`func Teardown(t *testing.T, name string) {
	list := teardownLists[name]

	delete(teardownLists, name)

	if t.Failed() || AlwaysRunDiagnosticTeardowns {
		list = append(list, diagnosticTeardownLists[name]...) // append any diagnostic funcs - so they are called FIRST
	}

	delete(diagnosticTeardownLists, name)
	
	for _, lmbd := range list {
		lmbd()
	}
}`

test/testlib/minikube_utilities.go
```diff
@@ -63,8 +65,32 @@ func AddTeardown(name string, teardownFunc func()) {
...
  * intermediate states.
  */
-func AddDiagnosticTeardown(name string, teardownFunc func()) {
-	diagnosticTeardownLists[name] = append(diagnosticTeardownLists[name], teardownFunc)
+func AddDiagnosticTeardown(name string, condition interface{}, teardownFunc func()) {
+	tdfunc = func() {
+		shouldIdoIt := strings.EqualFold(AlwaysRunDiagnoseTeardowns, "true")
+
+		if !shouldIdoIt {
+			switch c := condition.(type) {
```
adriansuarez, line 74: This type matching stuff is pretty cool.

test/testlib/minikube_utilities.go
NikTJ777: This call to teadownFunc is in a func() (tdfunc) that is being created in this context so that it has access to the 'condition'  param in its closure.

The tdfunc is then called in teardown - and tdfunc contains the conditional logic to decide whether the original diagnostic teardownFunc is actually called or not.

test/testlib/minikube_utilities.go
NikTJ777: Go has some really elegant syntax for things like that. :)

test/testlib/minikube_utilities.go
NikTJ777: @sivanov-nuodb 
> I'm a bit distracted by 1000+ changes in vendor files. Should these be part of the review?

The changes in 'vendor' are the result of my rebasing to master, which includes a change that removes the 'vendor' dir entirely.
By doing a rebase, I was trying to avoid those changes in master showing up in this subsequent PR.

NikTJ777: There are a few short usage examples in the unit tests:
test/minikube/minikube_test_test.go

test/testlib/minikube_utilities.go
NikTJ777: > somehow your branch is colliding with your previous dep->modules change
Seems I've somehow messed up the rebase of this branch onto master.

If that seems like a real problem, I'll just create (yet another) new branch,and just port these changes across manually - which is what the rebase was supposed to do automatically... :/

test/testlib/minikube_utilities.go
vegichan: I also missed `tdfunc()`. I get how it works now.

NikTJ777: I will add more comments in future...

test/testlib/minikube_utilities.go
NikTJ777: To complete the response - the idea is that the diagnostic teardowns are *not* run in Travis by default.

To have a diagnostic teardown run in Travis, you should pass in a 'conditional' that will be true when the diagnostic teardown should be run - for example:

* t.Failed() - will run if the test has already been marked as failed (the t.Failed() test is run at Teaardown() time)
* true - will run unconditionally - even if ALWAYS_RUN_DIAGNOSTIC_TEARDOWNS = false

test/testlib/minikube_utilities.go
NikTJ777: I have added some additional comments and 2 usage examples in the PR description.

sivanov-nuodb: Ah, it makes sense ... You create a closure/wrapper of the user-defined teardown function. Thanks for the explanation!

PR #133, opened 2020-08-04
Add support for LoadBalancer specification via annotations in admin statefulset and TE deployment

sivanov-nuodb:
**Changes**
* New annotations for Kubernetes Aware Admin (KAA) load-balancer config resync
* Add daemonsets resource to the `nuodb-kube-inspector` role
* Fix for `IsOpenShiftEnvironment` so that it doesn't report that OCP is used when `oc` is connected to regular K8s cluster

**Not part of this PR**
The job-lb-policy job is not removed as we'll need it until we upgrade to NuoDB version that supports KAA for LB config.

**Testing**
* Integration tests that run in GitHub
* Functional tests that run internally

**Example usage**

Helm values:
```
admin:
  lbConfig:
    prefilter: "not(label(region tiebreaker))"
    default: "random(first(label(node ${NODE_NAME:-}) label(zone ${ZONE_NAME:-}) any))"
    policies:
      nearest: "random(first(label(pod ${pod:-}) label(node ${node:-}) label(zone ${zone:-}) any))"
      zone1: "round_robin(first(label(zone zone1) label(zone zone2) any))"
      zone2: "round_robin(first(label(zone zone2) label(zone zone1) any))"
...
database:
  lbConfig:
    prefilter: "not(label(zone DR))"
    default: "random(first(label(node ${NODE_NAME:-}) label(zone ${ZONE_NAME:-}) any))"
```

should be equivalent to:

```
nuocmd set load-balancer-config --is-global \
  --prefilter 'not(label(region tiebreaker))'
 
nuocmd set load-balancer-config --is-global \
  --default 'random(first(label(node node1) label(zone zone1) any))'
 
nuocmd set load-balancer-config --db-name demo \
  --prefilter 'not(label(zone DR))'
 
nuocmd set load-balancer-config --db-name demo \
  --default 'random(first(label(node ${NODE_NAME:-}) label(zone ${ZONE_NAME:-}) any))'
 
nuocmd nuocmd set load-balancer --policy-name nearest \
  --lb-query 'random(first(label(pod ${pod:-}) label(node ${node:-}) label(zone ${zone:-}) any))'
nuocmd nuocmd set load-balancer --policy-name zone1 \
  --lb-query 'round_robin(first(label(zone zone1) label(zone zone2) any))'
nuocmd nuocmd set load-balancer --policy-name zone2 \
  --lb-query 'round_robin(first(label(zone zone2) label(zone zone1) any))'
```


stable/admin/templates/statefulset.yaml
```diff
@@ -1,6 +1,10 @@
 apiVersion: "apps/v1"
 kind: StatefulSet
 metadata:
+  annotations:
+    description: |-
+      NuoAdmin statefulset resource for NuoDB domain administration layer.
```
vegichan, line 6: I am not certain what the NuoDB terminology is for the admin. It might be admin tier. @adriansuarez ?

stable/admin/templates/role.yaml
```diff
@@ -18,6 +18,7 @@ rules:
   resources:
   - deployments
   - statefulsets
+  - daemonsets
```
vegichan, line 21: do you still require this?

test/minikube/minikube_domain_resync_test.go
```diff
@@ -311,6 +377,43 @@ func TestDomainResync(t *testing.T) {
...
+	defer testlib.Teardown(testlib.TEARDOWN_ADMIN)
+
+	options := &helm.Options{
+		SetValues: map[string]string{
+			"admin.lbConfig.prefilter":              "not(label(region tiebreaker))",
+			"admin.lbConfig.default":                "random(first(label(node node1) any))",
+			"admin.lbConfig.policies.zone1":         "round_robin(first(label(zone zone1) any))",
+			"admin.lbConfig.policies.nearest":       "random(first(label(pod ${pod:-}) label(node ${node:-}) label(zone ${zone:-}) any))",
+			"admin.lbConfig.fullSync":               "true",
+			"database.sm.resources.requests.cpu":    "0.25",
```
vegichan, line 392: you can probably give it `MINIMAL_VIABLE_ENGINE_CPU` since you are only starting 2 engines.

test/minikube/minikube_domain_resync_test.go
```diff
@@ -311,6 +377,43 @@ func TestDomainResync(t *testing.T) {
...
+			"database.sm.resources.requests.memory": testlib.MINIMAL_VIABLE_ENGINE_MEMORY,
+			"database.te.resources.requests.cpu":    "0.25",
+			"database.te.resources.requests.memory": testlib.MINIMAL_VIABLE_ENGINE_MEMORY,
+			"database.lbConfig.prefilter":           "not(label(zone DR))",
+			"database.lbConfig.default":             "random(first(label(node ${NODE_NAME:-}) any))",
+			"nuodb.image.tag":                       "4.2-mod",
```
vegichan, line 398: what is this tag? This is probably left over from your manual testing.

test/minikube/minikube_domain_resync_test.go
```diff
@@ -311,6 +377,43 @@ func TestDomainResync(t *testing.T) {
...
+		},
+	}
+
+	helmChartReleaseName, namespaceName := testlib.StartAdmin(t, options, 1, "")
+	admin0 := fmt.Sprintf("%s-nuodb-cluster0-0", helmChartReleaseName)
+	defer testlib.Teardown(testlib.TEARDOWN_DATABASE) // ensure resources allocated in called functions are released when this function exits
+	testlib.StartDatabase(t, namespaceName, admin0, options)
+
+	// Configure one manual policy
```
vegichan, line 407: I would add a test for the resync being `false` and verifying that manual configs don't get overridden.

test/testlib/nuodb_admin_utilities.go
```diff
@@ -130,4 +131,33 @@ func StartAdmin(t *testing.T, options *helm.Options, replicaCount int, namespace
 			helm.Install(t, options, "nuodb/admin ", helmChartReleaseName)
 		}
 	})
-}
\ No newline at end of file
+}
+
+func GetLoadBalancerPolicies(t *testing.T, namespaceName string, adminPod string) (map[string]NuoDBLoadBalancerPolicy, error) {
```
vegichan, line 136: so far we have followed the `TerraTest` naming convention of appending an `E` to the end of a getter function that can return an error. Functions without the `E` assert internally. Hence this function should be called `GetLoadBalancerPoliciesE`

applies to the ones below too

stable/admin/templates/statefulset.yaml
sivanov-nuodb: I'm happy to change that with more appropriate wording.

stable/admin/templates/role.yaml
sivanov-nuodb: We decided to display DaemonSets in `nuocmd --show-json get kubernetes-config` for diagnostic purposes although they are not used directly in KAA resync logic. I would propose to keep that in the role.

test/minikube/minikube_domain_resync_test.go
sivanov-nuodb: Thanks! I had problems with this when running the test on the internal CI.

test/minikube/minikube_domain_resync_test.go
sivanov-nuodb: Will do.

stable/admin/templates/statefulset.yaml
adriansuarez: I would just say "NuoDB Admin layer". Maybe @jleslie85 has an opinion once he is back.

stable/database/values.yaml
```diff
@@ -20,6 +20,7 @@ cloud:
     # If the NuoDB domain spans 2 or more physical clusters, then each cluster must have a unique cluster.name.
     # The default is fine for single-cluster domains.
     name: cluster0
+    entrypointName: cluster0
```
adriansuarez, line 23: Do we need this?

stable/database/values.yaml
```diff
@@ -154,6 +155,12 @@ database:
     # nuodb.config: |-
     #   verbose error,flush,warn
 
+  # Load balancer configuration per database
+  lbConfig:
+    prefilter: ""
+    default: ""
+    fullSync: false
```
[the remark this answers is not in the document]
sivanov-nuodb: Yes, it is only for the admin statefulset. It controls if the KAA module will delete LB configuration that is not backed up by annotations.

stable/database/values.yaml
butson: Check with Nik and Aaron,  this is so that when using multi-clusters you can peer to admin layer already started in a different cluster. I think it should be more generic to peer with already admin.   As in admin started in namespace a and want to start in namespace b.    So make it a fully qualified name.

acabrele: Dirk is right that is what it's for but I think stanimir has added here only because it is referenced in helpers.  namespace a/b scenario (in one cloud) is handled with the existing admin.namespace value I believe (so provide admin.namespace in separate database values file).  What we can't currently configure is peering an admin in one NS2/cloud2 to admin in NS1/cloud1, only NS1/cloud2 to NS1/cloud1.

stable/database/values.yaml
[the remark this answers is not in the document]
acabrele: So should be removed here?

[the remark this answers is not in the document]
sivanov-nuodb: Yes. Thanks so much!

stable/database/values.yaml
sivanov-nuodb: Yes. I added this because it's referenced in the chart. Technically it's not needed because it's defaulted in the templates but I think that it's good practice to have it in the values so users know that it's used.

@acabrele I'm not sure I understand completely the rest of your comment about and is it related to this PR or is more general limitation.

acabrele: More general response to Dirk, no additonal change requested.

butson: Can we deploy databases in one NS and admin in another NS.   I have no problem with admin pretty much always being in NS - nuodb.   However,  I can see projects deploying in another NS using the same admin if possible.  This was all possible a long time ago when we first wrote the helm charts. or whatever OpenShift had back then.

sivanov-nuodb:
> Should the job be suppressed somehow, e.g. by wrapping the entire YAML in some conditional block based on admin.lbConfig.fullSync so that we do not create a job to register a policy just so that KAA will delete the policy as soon as it does resync?

My suggestion would be to remove the job completely along with its variables once we change the NuoDB image version to some that support the KAA LB config.

adriansuarez:
> > Should the job be suppressed somehow, e.g. by wrapping the entire YAML in some conditional block based on admin.lbConfig.fullSync so that we do not create a job to register a policy just so that KAA will delete the policy as soon as it does resync?
> 
> My suggestion would be to remove the job completely along with its variables once we change the NuoDB image version to some that support the KAA LB config.

We cannot just remove the job, because that could break compatibility with versions of the product that do not support the annotations. I think we should conditionalize the job, which we can do now, and then update the `lbConfig` defaults after the published image supports the annotations.

stable/database/values.yaml
acabrele: @butson Yes this is possible, however since we added the serviceAccount, it requires some additional effort (before that, it worked without any modifications).  The issue being simply that the SA and role are created during admin deployment in the admin NS.  The db deployment in a separate DB NS is then missing the SA and role.  So this could be worked around by either setting the SA to be _default_ in values instead of _nuodb_ if security allows (like we used to), or manually deploying the SA/Role in the DB namespace.

PR #134, opened 2020-08-07
Change database container names to "engine"

adriansuarez:
Change database container name to engine for SM and hotcopy SM
StatefulSet, SM DaemonSet, and TE Deployment. This allows us to add
sidecar containers to run in the same Pod while still being able to
detect which container is running the database process.

### DIFF
--- a/stable/database/templates/daemonset.yaml
+++ b/stable/database/templates/daemonset.yaml
@@ -44,7 +44,7 @@ spec:
 {{ toYaml .Values.database.sm.tolerationsDS | trim | indent 8 }}
       {{- end }}          
       containers:
-      - name: sm
+      - name: engine
         image: {{ template "nuodb.image" . }}
         imagePullPolicy: {{ .Values.nuodb.image.pullPolicy }}
     {{- include "database.capabilities" . | indent 8 }}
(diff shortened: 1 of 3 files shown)


acabrele:
Any backward compatibility or upgrade considerations with this (as in, moving from one version of helm without this change to this version).

adriansuarez:
> Any backward compatibility or upgrade considerations with this (as in, moving from one version of helm without this change to this version).

I don't see anywhere that we use the container name. The CI build seems to be green, if we ignore the flakey TestKubernetesTLSRotation.

acabrele:
> > Any backward compatibility or upgrade considerations with this (as in, moving from one version of helm without this change to this version).
> 
> I don't see anywhere that we use the container name. The CI build seems to be green, if we ignore the flakey TestKubernetesTLSRotation.

I was thinking more k8s preventing that change to an existing deployment like we saw before with storage labels.  I just checked and we also have a helm upgrade test now anyway, so all good!

butson:
@A

> I was thinking more k8s preventing that change to an existing deployment like we saw before with storage labels. I just checked and we also have a helm upgrade test now anyway, so all good!

MY

butson:
Sorry for closing your comment Aaron,  I'm git challenged.  I was going to add a message that my quick testing of upgrade was okay.   But then I read your respond that that test passed.   And I ended up with a partition comment I went to delete and ended up closing .   Again sorry.

vegichan:
@butson  you closed the whole PR. We have to make a new one...

PR #136, opened 2020-08-17
Expose bootstrapServers value as Admin statefulset annotation

sivanov-nuodb:
This will expose `boostrapServers` value as `nuodb.com/bootstrap-servers` annotation for the Admin StatefulSet. Newer versions of NuoDB will take into account the value of the annotation and prefer it over the value of the `boostrapServers` label.

### DIFF
--- a/stable/admin/templates/statefulset.yaml
+++ b/stable/admin/templates/statefulset.yaml
@@ -5,6 +5,9 @@ metadata:
     description: |-
       NuoAdmin statefulset resource for NuoDB Admin layer.
 {{- include "admin.loadBalancerConfig" . | indent 4 }}
+    {{- if (eq (default "cluster0" .Values.cloud.cluster.name) (default "cluster0" .Values.cloud.cluster.entrypointName)) }}
+    nuodb.com/bootstrap-servers: {{ default 0 .Values.admin.bootstrapServers | quote }}
+    {{- end }}
   labels:
     app: {{ template "admin.fullname" . }}
     chart: {{ template "admin.chart" . }}
--- a/stable/admin/values.yaml
+++ b/stable/admin/values.yaml
@@ -94,8 +94,9 @@ admin:
   # cloud.cluster.entrypointName) can specify bootstrapServers.
   #
   # Changing the value of bootstrapServers on a existing domain, either by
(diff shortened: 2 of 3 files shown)


vegichan:
as per our conversation, please investigate how to not set the metadata label. Going forward, we would like to use annotations only.

sivanov-nuodb:
> as per our conversation, please investigate how to not set the metadata label. Going forward, we would like to use annotations only.

@vegichan As discussed before our customers can use older version of the charts with newer version of the product and the other way around. I think we should guarantee backward compatibility in both core products and charts. We could perform conditional rendering of the label based on used `nuodb.image.tag` and suppress the label if a new enough version of the product is used. I've tested with https://github.com/Masterminds/semver and looks like it works for our nuodb/nuodb-ce image versioning. I tested it manually with Helm v2.16.10 and v3.1.2.
This is the change that I can implement:

```
diff --git a/stable/admin/templates/statefulset.yaml b/stable/admin/templates/statefulset.yaml
index e6f56cd..5379cc3 100644
--- a/stable/admin/templates/statefulset.yaml
+++ b/stable/admin/templates/statefulset.yaml
@@ -14,9 +14,11 @@ metadata:
     group: nuodb
+    {{- if and (not (semverCompare "~4.0.7-0" .Values.nuodb.image.tag)) (not (semverCompare ">= 4.1.1-0" .Values.nuodb.image.tag)) }}
     {{- if (eq (default "cluster0" .Values.cloud.cluster.name) (default "cluster0" .Values.cloud.cluster.entrypointName)) }}
     bootstrapServers: {{ default 0 .Values.admin.bootstrapServers | quote }}
     {{- end }}
+    {{- end }}
 spec:
   serviceName: {{ .Values.admin.domain }}
```
This will break only if we have tags like `4.0.6.1` which is an invalid semantic version, however, likely we don't have such.
@acabrele , @NikTJ777 I would love to receive more comments on this before making the change.

acabrele:
> @acabrele , @NikTJ777 I would love to receive more comments on this before making the change.

Since the new way has not been released yet, I'd be tempted not to do this right now, and raise a Jira to handle whatever is appropriate when it is released.  I've never been a fan of attempting to predict the future!

acabrele (approved):

PR #138, opened 2020-08-27
dbutson/sidecar

butson:
These are the updates I made for the nuocd sidecars.   Chose to accept or not.  But, at least have something to work from.  I've currently set default values to disable the sidecars so there should be no changes to how default deployments work.  Once we get nuodb/nuocd in place we can changes default values to install the sidecars.


stable/database/values.yaml
```diff
@@ -357,6 +360,23 @@ database:
...
   enabled: false
+
+insights:
+  # disable insights by setting insights.enabled=false or admin.insights="" or admin.insights=null
+  # enabled here as a super values file that works for both admin and database helm charts can set
+  # to false to disable for both helm charts.
+  enabled: false
+  image:
+    registry: docker.io
+    repository: nuodb/nuocd
```
vegichan, line 375: this repo does not exist

butson: yes,  I know that.  Which is why sidecar is disabled for now.  You can enable sidecar and build nuodb/nuocd from nuodb-dashboards-influx and push to a local registry.  Then override image.registry.   If you or someone else can push to github nuodb then go ahead and build and install.

stable/admin/templates/role.yaml
```diff
@@ -9,6 +9,7 @@ rules:
   resources:
   - persistentvolumeclaims
   - pods
+  - configmaps
```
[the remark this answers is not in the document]
butson: This is needed for the sidecar which looks in project for configmaps that control how it is to configure the sidecar.   This allows us to deploy any number of plugins and redeploy without having to modify the running pods.

stable/admin/templates/_helpers.tpl
```diff
@@ -56,6 +56,53 @@ Also, we can't use a single if because lazy evaluation is not an option
...
+    {{- else -}}
+        {{- printf "%s/%s:%s" $registryName $repositoryName $tag -}}
+    {{- end -}}
+{{- else -}}
+    {{- printf "%s/%s:%s" $registryName $repositoryName $tag -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Return the proper insights image name
```
sivanov-nuodb, line 83: The description is the same as for `insights.image` helper. The logic seems to be almost the same.
Maybe you can consider parametrizing it or invoking it with a different scope?
Same applies to `stable/database/templates/_helpers.tpl`.

stable/admin/templates/statefulset.yaml
```diff
@@ -160,8 +160,39 @@ spec:
           mountPath: /etc/nuodb/keys/nuocmd.pem
           subPath: {{ .Values.admin.tlsClientPEM.key }}
         {{- end }}
+{{- if and .Values.insights.enabled .Values.admin.insights }}
```
sivanov-nuodb, line 163: `.Values.admin.insights` is string. I was wondering how empty string will evaluate. It looks like an empty string is `false`.

stable/database/templates/statefulset.yaml
```diff
@@ -495,8 +526,39 @@ spec:
           # the SM becomes unready if it does not start within 15 minutes = 30s + 15s*58
           successThreshold: 2
           timeoutSeconds: {{ default 5 .Values.database.sm.readinessTimeoutSeconds }}
+{{- if and .Values.database.insights .Values.insights.enabled }}
```
sivanov-nuodb, line 529: It looks like this section is repeated 4 times in the database chart. Can we make a template helper from that?

stable/admin/values.yaml
```diff
@@ -112,6 +112,9 @@ admin:
     policies:
       nearest: random(first(label(pod ${pod:-}) label(node ${node:-}) label(zone ${zone:-}) any))
   
+  # Label for insights configuration.
+  insights: "insights-ap-plugin"
```
sivanov-nuodb, line 116: It looks like that this is a specific name of an Insights plugin. Can we document which plugins are available and the names?

stable/database/values.yaml
```diff
@@ -357,6 +360,23 @@ database:
   isManualVolumeProvisioning: false
   isRestore: false
 
+
 openshift:
   # change this to true if you want to deploy using OpenShift proprietary features
   enabled: false
+
+insights:
```
sivanov-nuodb, line 368: We should mention the new values in the `README.md`.

stable/database/templates/statefulset.yaml
```diff
@@ -495,8 +526,39 @@ spec:
...
+          mountPath: /etc/telegraf/telegraf.d/
+        - name: log-volume
+          mountPath: /var/log/nuodb
+      shareProcessNamespace: true
+{{- end }}
 {{- include "nuodb.imagePullSecrets" . | indent 6 }}
       volumes:
+      {{- if and .Values.database.insights .Values.insights.enabled }}
+      - name: insights-config
+        emptyDir: {}
```
sivanov-nuodb, line 560: I remember seeing some configmaps on the demo. How they are configured? Should they be part of the admin chart?
It looks like the `insights-config` is ephemeral storage...

sivanov-nuodb (commented):
Are you planning to open separate PR for the rendering and minikube tests?

stable/database/templates/statefulset.yaml
butson: If you are refering to the if template helper does not add a lot of value.  If you are referring to the whole inclusion,  probably.  I expect it is the same fpr all.

stable/admin/templates/statefulset.yaml
butson: yes, empty is false, which is what I want.  If you want to disable for admin but not for engine you can set the admin.insights to "" and insights.enabled to true.

stable/database/templates/statefulset.yaml
butson: The configmaps are loaded by a different chart, current in nuodb-dashboard-insights.

stable/admin/values.yaml
sivanov-nuodb: Thanks. This is even better. I recommend we document the possible plugins that we have and mention that other custom plugins can be created. Maybe also change the `insights` key to something like `insights-plugin` which would be more descriptive I think.

stable/database/templates/statefulset.yaml
sivanov-nuodb: Do you plan to migrate the chart here? I think that we should mention the dependencies. Users might think that it's enough to set `insights` to `true` but it seems not the case.

butson: It's probably more of a naming issue than anything.    The architecture is one where a collector (nuocd) runs at the endpoints that you are monitoring.   It will push data to a server.   What that server is,  is up to user to configure along with the plugins to push to the server.  Out of the box, we ship with a server that configures plugins to collect and ship to influx.  The term insights is being used at both the collector and our default server.

stable/database/templates/statefulset.yaml
butson: I have a build that puts the whole section into a template.   One of the issues is that it depends upon log-volume which is not named the same between SM and TE.  should change the volume name in the te also to be log-volume instead of log-te-volume.  Then it's good.  I made changes by adding a _sidecar.tpl that has 2 templates.   One for additional volumes and one for additional sidecar containers.   That way there is one place if we add more sidecars.   But I don't know how to push these changes to this PR. Or if we want the changes, or if we want a different PR.

sivanov-nuodb: You can just push the changes to the same branch `dbutson/sidecar` and they will appear in the PR.

stable/database/templates/statefulset.yaml
sivanov-nuodb: Thanks for the explanation! My takeaway is that another chart from `nuodb-dashboard-insights` is needed to get the "server" along with the collector configuration.

sivanov-nuodb (changes requested):
There are some open discussions + questions about the rendering tests.

stable/database/templates/statefulset.yaml
butson: yes,  until you install the chart insights in nuodb-dashboard-insights.  The sidecar is dormant.   Once you install the plugins they will be picked up by the sidecar and it will start collecting the metrics.  If there is an output plugin defined then those metrics are pushed to the output endpoint.

vegichan:
Replaced by 161. Closing

PR #142, opened 2020-08-28
Update issue templates

vegichan:
Do not have a bug template set up. This proposal will create a default template for anyone who files issues against this GitHub repo.

It will show up here:
https://github.com/nuodb/nuodb-helm-charts/issues/new

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -28,3 +28,6 @@ A clear and concise description of what you expected to happen.
 
 **Additional context**
 Add any other context about the problem here.
+
+**External Reference**
+ZenDesk/JIRA numbers if available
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,30 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: ''
+assignees: ''
+
```
NikTJ777, line 7: Should we have an "external reference" field, that can hold either a JIRA number or a ZenDesk tickert ref?

vegichan: I dont think that I can add it to the auto-generated list of labels, but we can definitely add it to the end.

something like:?

External Reference
ZenDesk/JIRA numbers if available

acabrele (approved):

sivanov-nuodb (approved):
Thanks.

.github/ISSUE_TEMPLATE/bug_report.md
NikTJ777: Perfect! :)

PR #146, opened 2020-08-31
Improve tests for OpenShift 4.5.7 CodeReadyContainers 

vegichan:
- CRC does not have THP disabled. Minikube does. In case the test are being run in OCP, disable them automatically.
- the admin readiness probe timeout of 1s is not long enough. Make it 5s. Same as the DB charts
- TLS test work in OCP 4+. Remove the guards.

### DIFF
--- a/test/minikube/minikube_rolling_upgrade_test.go
+++ b/test/minikube/minikube_rolling_upgrade_test.go
@@ -91,7 +91,7 @@ func TestKubernetesUpgradeFullDatabaseMinorVersion(t *testing.T) {
 		SetValues: map[string]string{
 			"database.sm.resources.requests.cpu":    testlib.MINIMAL_VIABLE_ENGINE_CPU,
 			"database.sm.resources.requests.memory": testlib.MINIMAL_VIABLE_ENGINE_MEMORY,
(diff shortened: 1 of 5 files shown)


test/minikube/minikube_rolling_upgrade_test.go
```diff
@@ -140,6 +142,9 @@ func TestKubernetesUpgradeFullDatabaseMinorVersion(t *testing.T) {
 
 		helm.Upgrade(t, &databaseOptions, testlib.DATABASE_HELM_CHART_PATH, databaseHelmChartReleaseName)
 
+		// make sure that we only have 1 TE and not 2
+		testlib.SetDeploymentUpgradeStrategyToRecreate(t, namespaceName, fmt.Sprintf("te-%s-nuodb-cluster0-demo", databaseHelmChartReleaseName))
```
sivanov-nuodb, line 146: I can't map the comment with the upgrade strategy change. Is that so we need to change the strategy before upgrading when we have only one TE replica?

sivanov-nuodb (approved):
Adding CRC is a great improvement to our testing coverage. Thanks!

acabrele (approved):
Nice!

PR #148, opened 2020-09-01
Improve Upgrade/Enterprise Tests

vegichan:
- change all tests to prefer the injected version over the passed version. The version is no longer hardcoded. All tests now upgrade from a pre-defined old version to either the inject file or the current version in the template.
- move existing enterprise tests that depend on 4.0.7 into `long` or `diagnostics` suites
- minor change to the template files to prevent circular includes
- move all upgrade tests into their own `upgrade` suite
- change the minimal required version of NuoDB to 4.0.4 and fix the test to reflect that. 4.0.0 does not work and SMs/TEs will not start.
- remove PRINT_TO_STDOUT and set up an AWS bucket for log collection
- made collection of logs/k8s-logs async so that it keeps pulling until the container exits. This allows us to have complete logs including the last line.
- added admin log collection to all tests that restart the admin. This allows us to debug problems after a pod has been recreated.


vegichan:
when reviewing I suggest to ignore everything in the `test/integration` dir. All those changes are due to the file move.

vegichan:
@sivanov-nuodb this change enables your request to test helm upgrades using the injected unreleased nuodb version (what bamboo is doing)

test/minikube/minikube_domain_resync_test.go
```diff
@@ -295,6 +298,10 @@ func TestAdminScaleDown(t *testing.T) {
 }
 
 func TestDomainResync(t *testing.T) {
+	if os.Getenv("NUODB_LICENSE") != "ENTERPRISE" {
+		t.Skip("Cannot test resync without the Enterprise Edition")
```
adriansuarez, line 302: I think the actual reason we need EE is that this test starts multiple SMs. Is this now the only test that only runs internally?

vegichan: yeah. Its the only one left. If we rewrote it in a way where it does no need 2 SMs, we could make it run in travis.

vegichan:
the tests cant go into long. It expires after 50 min. I will move upgrade into its own plan.

test/minikube/minikube_domain_resync_test.go
adriansuarez: That is impossible because we are exercising SM scaledown and we cannot delete the only archive object in a database.

vegichan: that's what I thought. It will hence remain the only test that we only run in bamboo. I am sure that we will accumulate more over time. I will resolve the comment.

sivanov-nuodb (approved):
Thanks for making these changes!

adriansuarez (approved):

PR #149, opened 2020-09-07
Add the ability to pass through custom annotations to be applied to pods

acabrele:
Recent work to understand Vault integration options found an easy and effective approach was to use their sidecar method, however we don't have any way to provision pods with custom annotations in place which drive the vault sidecar integration (it would otherwise be a post deployment task).  Many other integrations use annotations to define and drive their behaviour, so making this generic opens up our integration options in the future.

- Added database.podAnnotations for TE and SM pods
- Added admin.podAnnotations for Admin pods
- Updated readme
- Added new test for various formats of supplied values including those used by Vault

Note the name podAnnotations was used to differentiate - it may be in the future we need to also have, for example, serviceAnnotations to enable some 3rd party CNIs.


stable/admin/values.yaml
```diff
@@ -223,3 +223,7 @@ admin:
 
   #Some clusters require longer readiness probe timeouts
   readinessTimeoutSeconds: 1
+
+  #These annotations will pass throgh to the pod as supplied, useful for integrating 3rd party products such as Vault.
+  podAnnotations: {}
```
sivanov-nuodb, line 228: As the annotations go to the controller K8s resource (i.e statefulset or deployment) maybe the variable name could be something like `customAnnotations` or `userDefinedAnnotations` or just `annotations`. This is just a suggestion.

stable/database/values.yaml
```diff
@@ -351,6 +351,10 @@ database:
     #Some clusters require longer readiness probe timeouts
     readinessTimeoutSeconds: 5
   
+  #These annotations will pass throgh to the pod as supplied, useful for integrating 3rd party products such as Vault.
+  annotations: {}
```
sivanov-nuodb, line 355: It seems not consistent with the variable used in the template.

test/integration/template_admin_test.go
```diff
@@ -365,3 +366,36 @@ func TestGlobalLoadBalancerConfigRendersOnlyOnEntryPointCluster(t *testing.T) {
...
+	helmChartPath := testlib.ADMIN_HELM_CHART_PATH
+
+	options := &helm.Options{
+		SetValues: map[string]string{
+			"admin.podAnnotations.key1": "value1",
+			"admin.podAnnotations.key2": "value2",
+			"admin.podAnnotations.key3\\.key3": "value3",
+			"admin.podAnnotations.key4\\.key4/key4": "value4",
+			"admin.podAnnotations.key5\\.key5/key5": "value5/value5",
+			"admin.podAnnotations.vault\\.hashicorp\\.com/agent-inject": `"true"`,
```
sivanov-nuodb, line 381: Some time ago I noticed a problem when using `toYaml` with bool values represented as strings (https://github.com/helm/helm/issues/4262). I don't know if this breaks somehow Vault in case it is boolean `true` instead of the string representation.

sivanov-nuodb (changes requested):

stable/admin/values.yaml
acabrele: I don't mind, but we need a way to seperate them from service annotations, for example... hence I moved away from just annotations...

stable/database/values.yaml
acabrele: Thanks, fixed

test/integration/template_admin_test.go
acabrele: I tested generating with both string and bool which both generate fine.  However I did find an issue using bool in the tests and had to switch to string for it to pass.

sivanov-nuodb (approved):
Thanks.

vegichan:
the PR failed with a render error:
```
Running basic
scripts/ci/determine_and_run_tests.sh: line 6: ./test/test_suite.sh: No such file or directory
# github.com/nuodb/nuodb-helm-charts/test/integration [github.com/nuodb/nuodb-helm-charts/test/integration.test]
test/integration/template_admin_test.go:392:22: undefined: SplitAndRenderStatefulSet
FAIL	github.com/nuodb/nuodb-helm-charts/test/integration [build failed]
FAIL
The command "scripts/ci/determine_and_run_tests.sh" exited with 2.```

**Comment by vegichan** [comment]
With changes like these, we should start thinking about documenting how the feature works from beginning to end. The single line doc change won't be sufficient. We already have a few readmes in https://github.com/nuodb/nuodb-helm-charts/tree/master/docs. I suggest that we add a doc "How to integrate NuoDB helm charts with Vault" there as well.

File: stable/database/values.yaml [hunk 4]
```diff
@@ -351,6 +351,10 @@ database:
     #Some clusters require longer readiness probe timeouts
     readinessTimeoutSeconds: 5
   
+  #These annotations will pass throgh to the pod as supplied, useful for integrating 3rd party products such as Vault.
```
vegichan, line 354: nit: we usually have a space between `#` and the next word.

vegichan (approved):
good changeset

stable/database/values.yaml
acabrele: Thanks, fixed

acabrele:
> With changes like these, we should start thinking about documenting how the feature works from beginning to end. The single line doc change won't be sufficient. We already have a few readmes in https://github.com/nuodb/nuodb-helm-charts/tree/master/docs. I suggest that we add a doc "How to integrate NuoDB helm charts with Vault" there as well.

This is a step in the direction to make that a nice document to read.  I agree we need it, although this change is not specific to it.  Can we defer that for when the Vault story is complete (namely enabling environment variables / passwords from files).

vegichan:
> > With changes like these, we should start thinking about documenting how the feature works from beginning to end. The single line doc change won't be sufficient. We already have a few readmes in https://github.com/nuodb/nuodb-helm-charts/tree/master/docs. I suggest that we add a doc "How to integrate NuoDB helm charts with Vault" there as well.
> 
> This is a step in the direction to make that a nice document to read. I agree we need it, although this change is not specific to it. Can we defer that for when the Vault story is complete (namely enabling environment variables / passwords from files).

Yeah. I am not blocking this PR :) I already approved it

Issue #140, opened 2020-08-28
Forbidden!Configured service account doesn't have access. Service account may have been revoked. daemonsets.apps is forbidden


vegichan:
NuoDB version: 4.0.7
NuoDB Helm charts: 2.4.0

Failure Description: NuoDB admin KAA module does not start

Log:
```
2020-08-27T14:48:33.553+0000 WARN  io.fabric8.kubernetes.client.informers.cache.Controller informer-controller-DaemonSet Reflector list-watching job exiting because the thread-pool is shutting down
java.util.concurrent.RejectedExecutionException: Error while starting ReflectorRunnable watch
        at io.fabric8.kubernetes.client.informers.cache.Reflector.listAndWatch(Reflector.java:85)
        at io.fabric8.kubernetes.client.informers.cache.Controller.run(Controller.java:112)
        at java.lang.Thread.run(Thread.java:748)
Caused by: java.util.concurrent.RejectedExecutionException: Error while doing ReflectorRunnable list
        at io.fabric8.kubernetes.client.informers.cache.Reflector.getList(Reflector.java:73)
        at io.fabric8.kubernetes.client.informers.cache.Reflector.reListAndSync(Reflector.java:94)
        at io.fabric8.kubernetes.client.informers.cache.Reflector.listAndWatch(Reflector.java:80)
        ... 2 common frames omitted
Caused by: io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: GET at: https://10.96.0.1/apis/apps/v1/namespaces/testadminscaledown-riqkzt/daemonsets. Message: Forbidden!Configured service account doesn't have access. Service account may have been revoked. daemonsets.apps is forbidden: User "system:serviceaccount:testadminscaledown-riqkzt:nuodb" cannot list resource "daemonsets" in API group "apps" in the namespace "testadminscaledown-riqkzt".
        at io.fabric8.kubernetes.client.dsl.base.OperationSupport.requestFailure(OperationSupport.java:568)
        at io.fabric8.kubernetes.client.dsl.base.OperationSupport.assertResponseCode(OperationSupport.java:505)
        at io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleResponse(OperationSupport.java:471)
        at io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleResponse(OperationSupport.java:430)
        at io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleResponse(OperationSupport.java:412)
[...]

PR #150, opened 2020-09-08
Start Kubernetes Aware Admin module with limited permissions

sivanov-nuodb:
**Changes**
- Make sure that KAA starts when a limited set of permissions are present in RBAC
- This addresses the backward compatibility issue described in https://github.com/nuodb/nuodb-helm-charts/issues/140

### DIFF
--- a/test/minikube/minikube_domain_resync_test.go
+++ b/test/minikube/minikube_domain_resync_test.go
@@ -10,9 +10,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"os"
-	"path/filepath"
 	"strconv"
 	"strings"
 	"testing"
@@ -24,12 +22,9 @@ import (
 
 	"github.com/nuodb/nuodb-helm-charts/test/testlib"
 
-	"github.com/ghodss/yaml"
-
 	"github.com/gruntwork-io/terratest/modules/helm"
 	"github.com/gruntwork-io/terratest/modules/k8s"
 	appsv1 "k8s.io/api/apps/v1"
-	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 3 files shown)


test/minikube/minikube_domain_resync_test.go
```diff
@@ -465,4 +488,61 @@ func TestLoadBalancerConfigurationResync(t *testing.T) {
...
+	if os.Getenv("NUODB_LICENSE") != "ENTERPRISE" {
+		t.Skip("This test requires NuoDB 4.1.1+")
+	}
+	testlib.AwaitTillerUp(t)
+	defer testlib.VerifyTeardown(t)
+	defer testlib.Teardown(testlib.TEARDOWN_ADMIN)
+
+	modifyKubeInspectorRoleInPlace(t, func(role *rbacv1.Role) {
+		// Remove "daemonsets" from the resources list
+		role.Rules[1].Resources = []string{"statefulsets", "deployments"}
```
adriansuarez, line 503: Also removing `pods`?

test/minikube/minikube_domain_resync_test.go
```diff
@@ -188,6 +194,23 @@ func checkInitialMembership(t assert.TestingT, configJson string, expectedSize i
...
+	testlib.AddTeardown(testlib.TEARDOWN_ADMIN, func() { ioutil.WriteFile(inspectorRoleFile, originalData, 0644) })
+
+	output := helm.RenderTemplate(t, &helm.Options{}, testlib.ADMIN_HELM_CHART_PATH, "release-name", []string{"templates/role.yaml"})
+	roles := testlib.SplitAndRenderRole(t, output, 1)
+	modificationFunc(&roles[0])
+	roleBytes, err := yaml.Marshal(&roles[0])
+	assert.NoError(t, err)
+	err = ioutil.WriteFile(inspectorRoleFile, roleBytes, 0644)
+	assert.NoError(t, err)
+	out, _ := ioutil.ReadFile(inspectorRoleFile)
```
vegichan, line 210: are you overriding the file on disk? I would just create a temp file

test/minikube/minikube_domain_resync_test.go
```diff
@@ -465,4 +488,61 @@ func TestLoadBalancerConfigurationResync(t *testing.T) {
...
 	options.SetValues["admin.lbConfig.policies.manual"] = "random(any)"
 	verifyLoadBalancer(t, namespaceName, admin0, options.SetValues)
-}
\ No newline at end of file
+}
+
+func TestKaaLimitedPermissions(t *testing.T) {
+	if os.Getenv("NUODB_LICENSE") != "ENTERPRISE" {
+		t.Skip("This test requires NuoDB 4.1.1+")
+	}
```
vegichan, line 496: hmm. This works, but I don't like using the license as a version guard. 

My idea was to create a new file and put in the `enterprise` build guard. Once we upgrade to say 4.1.1 as the default NuoDB instance, we can then move the whole file to say `long`.

test/minikube/minikube_domain_resync_test.go
```diff
@@ -465,4 +488,61 @@ func TestLoadBalancerConfigurationResync(t *testing.T) {
...
+	// Verify that KAA will not register informer for PVCs
+	testlib.Await(t, func() bool {
+		return testlib.GetStringOccurrenceInLog(t, namespaceName, admin0,
+			"Informer for resource 'persistentvolumeclaims' not registered", &v12.PodLogOptions{}) == 1
+	}, 30*time.Second)
+
+	// Verify that resources that KAA have permissions for are available
+	adminStatefulSet := fmt.Sprintf("%s-nuodb-cluster0", helmChartReleaseName)
+	teDeployment := fmt.Sprintf("te-%s-nuodb-cluster0-demo", databaseHelmChartReleaseName)
+	config := testlib.GetNuoDBK8sConfigDump(t, namespaceName, admin0)
```
vegichan, line 543: good idea!

vegichan (approved):
besides minor comments, this is a great change. Thanks!

test/minikube/minikube_domain_resync_test.go
sivanov-nuodb: The rule with index 1 is having normally 3 resources. This is removing just the `daemonsets` from the list.
 ```
- apiGroups:
  - apps
  resources:
  - deployments
  - statefulsets
  - daemonsets
  verbs:
  - get
  - list
  - watch
```

test/minikube/minikube_domain_resync_test.go
sivanov-nuodb: As I know helm will send all files in the `templates` directory via the rendering engine. Either we should edit the file in place and revert the change or we should move the original file and revert it later. I preferred the first option as we don't need to write a lot of helper code just to move files around.

test/minikube/minikube_domain_resync_test.go
sivanov-nuodb: That works for me. I will move the test to a new file and tag it as `enterprise`.

adriansuarez (approved):
Looks good.

PR #153, opened 2020-09-15
Add "nuodb" serviceaccount for the admin's LB job

kmabda:
The admin's job may fails to start if RBAC rolebinding are needed.

Use the helm charts nuodb.serviceAccount to set the service account for the admin's job


sivanov-nuodb (commented):
Is there any special scenario/deployment where the job is failing due to using the default service account? Are there any additional RBAC permissions that need to be granted to the service account specifically for this job?

kmabda:
There is no additional RBAC permissions needed for the job, it's the same ones granted to the other nuodb pods.

The nuodb service account was granted permission to use any uid, however since the Job was missing an explicit service account (i.e. was using the "default" service account) it was failing to start.

sivanov-nuodb (approved):
Thanks for the additional explanation and for doing this change!

vegichan (approved):
thank you for the contribution!

PR #154, opened 2020-09-16
render --user option conditionally on a nun-null credentials

NikTJ777:
And put quotes around the URL

See [HELM-207]

When using secure URLs on cloud storage, the authentication token is embedded in the URL.
In this situation, if the --user option is also provided to 'curl', many endpoints return an error stating only 1 authentication method can be specified.

This change causes the --user option to only be rendered into the curl command if the credentials value is null or effectively null ( ':' ).

In addition, double-quotes are used to enclose the URL to protect shell-metacharacters, since embedding the authentication token will usually require '?' and '&' characters in the URL.


acabrele (approved):

vegichan (approved):
thank you!

PR #155, opened 2020-10-07
Remove https references for NUOCMD_API_SERVER and add certificates for upgrade tests

sivanov-nuodb:
**Changes**
Starting with NuoDB 4.2 the default TLS keys won't be shipped with the image. Users should either generate the keys and expose them as secrets or disable SSL.

Self-signed certificates can be generated by using `setup-keys.sh` script which is available in the image. This PR doesn't include any improvements that we might implement like auto-generation of self-signed keys at runtime.

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -244,7 +244,6 @@ github.com/gruntwork-io/gruntwork-cli v0.6.1 h1:sPeSdsn3TU1PJMPUHJRFfrxTBHNF0Y0O
 github.com/gruntwork-io/gruntwork-cli v0.6.1/go.mod h1:jp6Z7NcLF2avpY8v71fBx6hds9eOFPELSuD/VPv7w00=
 github.com/gruntwork-io/terratest v0.28.10 h1:4K3idJK5atavZ7iwY+L+RPCA7lFwPu87dVnoFOaoPzI=
 github.com/gruntwork-io/terratest v0.28.10/go.mod h1:PkVylPuUNmItkfOTwSiFreYA4FkanK8AluBuNeGxQOw=
-github.com/gruntwork-io/terratest v0.30.7 h1:RmvfWTngHv9Mat2AF4++Znte8Znkz585/IzDvUgz84o=
 github.com/hashicorp/golang-lru v0.5.0/go.mod h1:/m3WP610KZHVQ1SGc6re/UDhFvYD7pJ4Ao+sR/qLZy8=
 github.com/hashicorp/golang-lru v0.5.1/go.mod h1:/m3WP610KZHVQ1SGc6re/UDhFvYD7pJ4Ao+sR/qLZy8=
 github.com/hashicorp/golang-lru v0.5.3/go.mod h1:iADmTwqILo4mZ8BN3D2Q6+9jd8WM5uGBxy+E8yxSoD4=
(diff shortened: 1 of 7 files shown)


go.sum
```diff
@@ -244,6 +244,7 @@ github.com/gruntwork-io/gruntwork-cli v0.6.1 h1:sPeSdsn3TU1PJMPUHJRFfrxTBHNF0Y0O
 github.com/gruntwork-io/gruntwork-cli v0.6.1/go.mod h1:jp6Z7NcLF2avpY8v71fBx6hds9eOFPELSuD/VPv7w00=
 github.com/gruntwork-io/terratest v0.28.10 h1:4K3idJK5atavZ7iwY+L+RPCA7lFwPu87dVnoFOaoPzI=
 github.com/gruntwork-io/terratest v0.28.10/go.mod h1:PkVylPuUNmItkfOTwSiFreYA4FkanK8AluBuNeGxQOw=
+github.com/gruntwork-io/terratest v0.30.7 h1:RmvfWTngHv9Mat2AF4++Znte8Znkz585/IzDvUgz84o=
```
vegichan, line 247: where is this change coming from?

vegichan:
this will fix the tests, but what happens if someone attempts to install 4.2 with the current charts. Will they have to provide overrides too?

go.sum
sivanov-nuodb: It looks that I'm running with terratest v0.30.7 locally and `go test` is updating this file as it's tracked in version control.

sivanov-nuodb:
> this will fix the tests, but what happens if someone attempts to install 4.2 with the current charts. Will they have to provide overrides too?

The current plan is that we would like the users to be aware that they switch off TLS. Hence NuoDB 4.2+ will need to be installed with `helm install admin --set admin.options.ssl=false ...`.

test/testlib/minikube_utilities.go
```diff
@@ -219,10 +219,22 @@ func InjectTestVersion(t *testing.T, options *helm.Options) {
 	options.SetValues["nuodb.image.tag"] = image.Nuodb.Image.Tag
 }
 
+func InjectSSLMode(t *testing.T, options *helm.Options) {
```
NikTJ777, line 222: Does this logic also rely on `nuodocker` internally modifying its connection URL from https to http if ssl is disabled?

I don't believe it does that yet - but should shortly.

acabrele:
Should we add to the docs in this change for starting with SSL false or leave that until later

acabrele (approved):

sivanov-nuodb:
After some more internal discussions, we decided to implement an approach that will lower the friction for helm charts users but at the same time keep the default value for SSL to `true`. With 4.2+ in case, the user didn't provide overwrite for SSL and the keystore files in the default locations are missing, the SSL will be downgraded to "false".
This won't require `--set admin.options.ssl=false` and the admin chart installation will be the same as before. Users that have used secrets with the help charts before should be able to upgrade without any additional steps.

@acabrele @vegichan I've removed several occurrences of hardcoded `https://` and added certificates for the upgrade tests. Could you please check the changes again?

acabrele:
> After some more internal discussions, we decided to implement an approach that will lower the friction for helm charts users but at the same time keep the default value for SSL to `true`. With 4.2+ in case, the user didn't provide overwrite for SSL and the keystore files in the default locations are missing, the SSL will be downgraded to "false".
> This won't require `--set admin.options.ssl=false` and the admin chart installation will be the same as before. Users that have used secrets with the help charts before should be able to upgrade without any additional steps.
> 
> @acabrele @vegichan I've removed several occurrences of hardcoded `https://` and added certificates for the upgrade tests. Could you please check the changes again?

I suppose this could introduce the possibility that someone intended to set up TLS, failed in some way, but because the deployment succeeds, assumes that TLS _is_ enabled?

vegichan (approved):
I am impressed by how easy this was to change!

sivanov-nuodb:
@acabrele If the standard secrets I can't think of a way to get missing files without an error during install.

How likely is that Vault fails to mount all files due to runtime error some time after initial deployment?
In such cases SSL mode can be locked down to the decired mode by passing `admin.options.ssl`.

adriansuarez:
> > this will fix the tests, but what happens if someone attempts to install 4.2 with the current charts. Will they have to provide overrides too?
> 
> The current plan is that we would like the users to be aware that they switch off TLS. Hence NuoDB 4.2+ will need to be installed with `helm install admin --set admin.options.ssl=false ...`.

I don't think this comment is still relevant after the automatic downgrade to non-TLS.

Versions 4.2+ of NuoDB do have a dependency on the newest version of the Helm charts due to the removal of the `https://` from URLs to enable downgrade of the client.

adriansuarez (approved):
Looks good.

PR #156, opened 2020-10-08
Move podAnnotations to create on the pod

acabrele:
- podAnnotations were added but created on the pod controller not the pod itself.  Vault requires them to be on the pod, and this fits better with the naming too.
- Added more tests for the DB charts


sivanov-nuodb (approved):
@acabrele thanks for the change and it's good to see more tests!

vegichan (approved):
Good that we caught this before it got released

PR #157, opened 2020-10-13
remove single-quotes from curl_creds

NikTJ777:
Single quotes had been used to protect against shell-expansion of the userid:password string.
However, this value was being resolved from a variable, and does not need protection, and the quotes were actually being added to the value, causing an authentication failure.


vegichan (approved):
thanks for the quick turnaround!

acabrele (approved):

sivanov-nuodb (approved):
Many thanks for fixing this!

PR #158, opened 2020-10-16
Change tests from assert to require

vegichan:
When I modified the test library to use stretcher, I incorrectly decided to use assert.
- `assert` marks the test as failed, but continues
- `requires` fails the test straight away

The correct equivalent of the Helm 2.X tests is actually `requires`. We noticed that some tests failed in a bizarre way, because their execution did not get aborted early.

This is a sweeping change to all minikube tests. In some cases we could use assert instead, but we have never made that differentiation so far, so I did not introduce is.


NikTJ777 (approved):
Great job - well caught...

PR #159, opened 2020-10-28
Test that KAA will not start if addRoleBinding is set to false

sivanov-nuodb:
**Changes**
New test case added to the enterprise suite to verify that KKA won't start if `addRoleBinding` is set to `false` due to mandatory permissions missing.

### DIFF
--- a/test/minikube/minikube_kaa_additions_test.go
+++ b/test/minikube/minikube_kaa_additions_test.go
@@ -123,7 +123,7 @@ func TestKaaRolebindingDisabled(t *testing.T) {
 			"Not registering event listeners: service account unauthorized for resource 'leases'", &v12.PodLogOptions{}) == 1
 	}, 30*time.Second)
 
-	// Verify that no resources are avaialble via KKA
+	// Verify that no resources are avaialble via KAA
 	config := testlib.GetNuoDBK8sConfigDump(t, namespaceName, admin0)
 	require.True(t, len(config.StatefulSets) == 0)
 	require.True(t, len(config.Deployments) == 0)
--- a/test/minikube/minikube_kaa_additions_test.go
+++ b/test/minikube/minikube_kaa_additions_test.go
@@ -108,15 +108,6 @@ func TestKaaRolebindingDisabled(t *testing.T) {
 
 	defer testlib.Teardown(testlib.TEARDOWN_DATABASE)
 
-	testlib.StartDatabase(t, namespaceName, admin0, &helm.Options{
-		SetValues: map[string]string{
-			"database.sm.resources.requests.cpu":    testlib.MINIMAL_VIABLE_ENGINE_CPU,
-			"database.sm.resources.requests.memory": testlib.MINIMAL_VIABLE_ENGINE_MEMORY,
-			"database.te.resources.requests.cpu":    testlib.MINIMAL_VIABLE_ENGINE_CPU,
-			"database.te.resources.requests.memory": testlib.MINIMAL_VIABLE_ENGINE_MEMORY,
-		},
-	})
-
 	// Verify that KAA won't start due to limited mandatory permissions
 	testlib.Await(t, func() bool {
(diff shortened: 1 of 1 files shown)


test/minikube/minikube_kaa_additions_test.go
```diff
@@ -92,3 +92,41 @@ func TestKaaLimitedPermissions(t *testing.T) {
...
+	defer testlib.Teardown(testlib.TEARDOWN_DATABASE)
+
+	testlib.StartDatabase(t, namespaceName, admin0, &helm.Options{
+		SetValues: map[string]string{
+			"database.sm.resources.requests.cpu":    testlib.MINIMAL_VIABLE_ENGINE_CPU,
+			"database.sm.resources.requests.memory": testlib.MINIMAL_VIABLE_ENGINE_MEMORY,
+			"database.te.resources.requests.cpu":    testlib.MINIMAL_VIABLE_ENGINE_CPU,
+			"database.te.resources.requests.memory": testlib.MINIMAL_VIABLE_ENGINE_MEMORY,
+		},
+	})
```
vegichan, line 118: do you need the database here?

test/minikube/minikube_kaa_additions_test.go
```diff
@@ -92,3 +92,41 @@ func TestKaaLimitedPermissions(t *testing.T) {
...
+
+	// Verify that KAA won't start due to limited mandatory permissions
+	testlib.Await(t, func() bool {
+		return testlib.GetStringOccurrenceInLog(t, namespaceName, admin0,
+			"Not registering event listeners: service account unauthorized for resource 'leases'", &v12.PodLogOptions{}) == 1
+	}, 30*time.Second)
+
+	// Verify that no resources are avaialble via KKA
```
vegichan, line 126: typo

vegichan:
if you add the required permissions after the AP was started; Will KAA start?

test/minikube/minikube_kaa_additions_test.go
sivanov-nuodb: Later on, we are checking that there are no K8s resources returned by `nuocmd`. So that we minimize the test time, I will remove the database and skip the deployments check.

test/minikube/minikube_kaa_additions_test.go
sivanov-nuodb: Thanks.

sivanov-nuodb:
> if you add the required permissions after the AP was started; Will KAA start?
The mandatory permissions are evaluated on admin startup. The admin pods need to be restarted if you add the new permissions. The same behavior is seen in 4.0.7.

Note: If you add previously revoked permissions while admin is running, no actions are needed.

vegichan (approved):
Thanks for the updates!

acabrele (approved):

test/minikube/minikube_kaa_additions_test.go
```diff
@@ -92,3 +92,31 @@ func TestKaaLimitedPermissions(t *testing.T) {
 	require.True(t, len(config.Volumes) == 0)
 	require.True(t, len(config.Pods) == 0)
 }
+
+func TestKaaRolebindingDisabled(t *testing.T) {
+	// This test requires NuoDB 4.2+ or 4.1.2+
```
adriansuarez, line 97: This file is a convenient place to put the test for now that the functionality is unreleased and can only be tested internally, but it does not require an Enterprise license, and we should put it in a different file once the 4.1.2/4.2 are released so that this can run in Travis.

Before tests like these pile up that are candidates for Travis, we should create a new convention for testing unreleased functionality that is separate from the `enterprise` build tag. We should put tests like these in a normal test file (not having the `enterprise` tag) and explicitly check an environment variable like `UNRELEASED_FUNCTIONALITY`. And then we can remove the check whenever a relevant release is made.

adriansuarez (approved):
Looks good.

PR #160, opened 2020-11-02
add timeout and retry on wait for starting SM

NikTJ777:
Per helm-185, if the "starting SM" encounters an error and is restarted, then the SMs waiting on it would wait forever.

Changed "nuocmd check process ... --wait-forever"
to: "nuocmd check process ... --timeout 600"

If "check process" returns an error, then print a diagnostic message and retry the enclosing loop.

In the process of retrying, the identity of the "starting SM" is re-evaluated, in case the SM restarted.

### DIFF
--- a/stable/database/files/nuosm
+++ b/stable/database/files/nuosm
@@ -265,7 +265,12 @@ if [ -n "$restore_requested" -a -n "$restore_source" ]; then
     fi
 
     # wait for the starting SM to be RUNNING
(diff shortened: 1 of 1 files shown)


vegichan (approved):
looks good!

acabrele (approved):

sivanov-nuodb (approved):
@NikTJ777 Thanks so much for fixing this!

PR #161, opened 2020-11-03
Add NuoDB Collector support in NuoDB Helm charts

sivanov-nuodb:
**Changes**
- Sidecar changes made by Dirk in https://github.com/nuodb/nuodb-helm-charts/pull/138 have been used as a start point
- Add support for user-defined NuoDB Collector plugins. By default, all plugins shipped with the image in `/etc/telegraf/telegraf.d/static/` directory are used. All dynamic plugins are collected from configMaps and stored in `/etc/telegraf/telegraf.d/dynamic/` directory. 
- Replace `admin.insights` and `database.insights` labels values with `nuodb.com/nuocollector-plugin`. The `nuocollector-config` watcher sidecar will load plugins based on this label value. Insights chart is installing dynamic output plugin with value "insights" which is selected and loaded if available.

Example installation with insights enabled:
```
helm install admin nuodb/admin --set nuocollector.enabled="true" --namespace nuodb
helm install database nuodb/database --set nuocollector.enabled="true" --namespace nuodb
helm install insights nuodb/insights --namespace nuodb
```
Example of user-defined plugins:

```
nuocollector:
  enabled: true
  plugins:
    database:
      file: |-
        [[outputs.file]]
        files = ["/var/log/nuodb/metrics.log"]
        data_format = "influx"
```

**Dependencies**
- This change requires `nuocollector` version 1.0.1 (or the next release version).  Specifically, the changes in PR#9 are needed.

**Testing**
- Integration tests for the sidecars have been added
- Functional tests have been added that use `file` Telegraf plugin to write all stats in `stdout`. This removes the InfluxDB dependency and makes it easier to assert for metrics.
- `inputs.tail` Telegraf plugin have been used in functional tests to verify that metrics are collected for admin pods as well. This plugin is not included in default values as according to Dirk "Telegraf is not designed necessarily to handle logs and it can probably easily break insights with data volume". In the future, we might ship a new plugin designed for admin metrics.


stable/admin/templates/insights-configmap.yaml
```diff
@@ -0,0 +1,28 @@
+
+{{- if $.Values.insights }}
+{{- if $.Values.insights.enabled }}
+{{- if $.Values.insights.plugins }}
```
adriansuarez, line 4: Can we combine these with `and`?

stable/database/templates/deployment.yaml
```diff
@@ -115,7 +115,7 @@ spec:
           subPath: {{ $key }}
         {{- end -}}
         {{- end }}
-        - name: log-te-volume
+        - name: log-volume
```
adriansuarez, line 118: +1

test/integration/template_insights_test.go
```diff
@@ -0,0 +1,240 @@
...
+
+	v1 "k8s.io/api/core/v1"
+
+	"github.com/gruntwork-io/terratest/modules/helm"
+
+	"github.com/nuodb/nuodb-helm-charts/test/testlib"
+)
+
+func checkSidecarContainers(t *testing.T, containers []v1.Container, options *helm.Options, chartPath string) {
+	assert.NotEmpty(t, containers)
```
vegichan, line 18: I am assuming that you want to abandon the test straight away when this assert is false. In stretch, that's actually `require.NotEmpty`, not `assert.`

test/integration/template_insights_test.go
```diff
@@ -0,0 +1,240 @@
...
+	found := false
+	for _, volume := range volumes {
+		if volume.Name == "insights-config" {
+			found = true
+			// Check that empty dir is mounted
+			assert.NotNil(t, volume.EmptyDir)
+		}
+	}
+	if options.SetValues["insights.enabled"] == "true" {
+		assert.True(t, found)
```
vegichan, line 86: assert.True tends to have a fairly incomprehensible failure message. @NikTJ777 always advocates for explicit failure messages. So something like `assert.True(t, found, "insights-config volume not found") might be helpful

test/testlib/nuodb_database_utilities.go
```diff
@@ -143,4 +143,35 @@ func StartDatabase(t *testing.T, namespace string, adminPod string, options *hel
 func SetDeploymentUpgradeStrategyToRecreate(t *testing.T, namespaceName string, deploymentName string) {
 	kubectlOptions := k8s.NewKubectlOptions("", "", namespaceName)
 	k8s.RunKubectl(t, kubectlOptions, "patch", "deployment", deploymentName, "-p", UPGRADE_STRATEGY)
-}
\ No newline at end of file
+}
+
+func PopulateDBWithQuickstart(t *testing.T, namespaceName string, adminPod string, dbName string) {
```
vegichan, line 148: just pointing out that these take `forever` to finish... I wish we had something that inserts some data and take less time... Nothing you can do here.

test/testlib/minikube_utilities.go
```diff
@@ -727,6 +740,15 @@ func getAppLogStream(t *testing.T, namespace string, podName string, podLogOptio
 	client, err := k8s.GetKubernetesClientFromOptionsE(t, options)
 	require.NoError(t, err)
 
+	if podLogOptions.Container == "" {
+		// Select first container if not specified; otherwise the GetLogs method will fail if there are sidecars
```
vegichan, line 744: sounds like the expectation is to get logs from the "engine" container, or "admin" container. Do we have the naming convention in place? Can we select either engine or admin, rather than [0]?

test/minikube/minikube_base_insights_test.go
```diff
@@ -0,0 +1,199 @@
...
+		createOutputFilePlugin(t, namespaceName)
+		defer testlib.Teardown(testlib.TEARDOWN_INSIGHTS)
+		testlib.PopulateDBWithQuickstart(t, namespaceName, admin0, "demo")
+		t.Run("verifyMetricsCollection", func(t *testing.T) {
+			verifyCollectionForAdmin(t, namespaceName, fmt.Sprintf("%s-nuodb-cluster0", adminReleaseName))
+			verifyCollectionForDatabase(t, namespaceName, fmt.Sprintf("%s-nuodb-%s-%s", databaseRealeaseName, "cluster0", "demo"), "demo")
+		})
+	})
+
+	t.Run("startDatabaseMultiTenant", func(t *testing.T) {
```
vegichan, line 158: this sub-test probably takes forever and a half. If this test would be pushing data from multiple sources to the same visualization product, I would say "lets keep it" in order to verify that the visualization product can behave in a correct multi-tenant fashion. But the collector being tested here should not do anything special if there are multiple independent DBs running. So I suggest that we don't have this test.

stable/database/values.yaml
```diff
@@ -361,6 +362,26 @@ database:
...
+
+insights:
+  # disable insights by setting insights.enabled=false or admin.insights="" or admin.insights=null
+  # enabled here as a super values file that works for both admin and database helm charts can set
+  # to false to disable for both helm charts.
+  enabled: false
+  image:
+    registry: docker.io
+    repository: nuodb/nuodb-collector
+    tag: 1.0.1
```
vegichan, line 378: this and other places. Should we hardcode the version of the collector same as we do the DB, or should we use `latest`? I have no strong preference, but I wanna call it out explicitly.

test/minikube/minikube_base_insights_test.go
sivanov-nuodb: Locally it takes ~5 min:
```
--- PASS: TestInsightsMetricsCollection (337.55s)
    --- PASS: TestInsightsMetricsCollection/startDatabaseMultiTenant (271.38s)
        --- PASS: TestInsightsMetricsCollection/startDatabaseMultiTenant/verifyMetricsCollectionForAdmin (24.01s)
        --- PASS: TestInsightsMetricsCollection/startDatabaseMultiTenant/verifyMetricsCollectionForGreen (25.80s)
        --- PASS: TestInsightsMetricsCollection/startDatabaseMultiTenant/verifyMetricsCollectionForBlue (18.02s)
```

The main reason to have this test is to verify that multiple DB deployments can forward metrics to a single destination. I'm fine skipping the test to save test execution time.

stable/admin/templates/insights-configmap.yaml
sivanov-nuodb: Good catch! I've updated this in both templates.

test/integration/template_insights_test.go
sivanov-nuodb: Not sure if we have a convention when using `assert` or `require`. I normally use `assert` in the actual test and `require` in `testlib` utilities.

test/integration/template_insights_test.go
sivanov-nuodb: Thanks for the comment! I agree with you and added messages.

test/testlib/nuodb_database_utilities.go
sivanov-nuodb: Actually it's not that long. Locally it takes ~10sec.

test/testlib/minikube_utilities.go
sivanov-nuodb: In practice, the first element is the main container in the pod (either `admin` or `engine`). It's a shame that Golang doesn't have generic slice/array filter or find methods. If we need to explicitly find the container name, we should add something like:
```
var containerName string
for _, container := range pod.Spec.Containers {
  if container.Name == "admin" || container.Name == "engine" {
    containerName = container.Name
    break
  }
}
if containerName == "" {
  containerName = pod.Spec.Containers[0].Name
}
```
I wished that we can avoid that...

test/integration/template_insights_test.go
vegichan: it is not about convention. `assert` marks the test as failed, but continues. So in this case, it will let marked as failed, but further operations can still be executed. To give you an example:

container := nil
assert.NotNil(t, container)
assert.Greater(t, size(container), 2) // we assume that there are two elements
if container[0]... <- this will panic

test/testlib/minikube_utilities.go
vegichan: I would do this: https://play.golang.org/p/Q0UET49Ez6T

test/minikube/minikube_base_insights_test.go
vegichan: ok with me. resolving

test/testlib/constants.go
```diff
@@ -24,6 +24,7 @@ const TEARDOWN_DATABASE = "database"
 const TEARDOWN_RESTORE = "database"
 const TEARDOWN_SECRETS = "secrets"
 const TEARDOWN_YCSB = "ycsb"
+const TEARDOWN_INSIGHTS = "insights"
```
vegichan, line 27: I will make this comment here, but it applies to a few more places. Can you please purge the term "insights" from places that are actually the collector? That is: variables, constants, function names, etc.

I wonder, is there anything in this review that relates to "Insights" the monitoring stack? Or are all changes here related to the collector? I think that you check log files? You don't output anything to Influx?

If this review has nothing to do with a particular output target, I suggest that we remove `all` mentions of Insights.

test/minikube/minikube_base_insights_test.go
```diff
@@ -0,0 +1,200 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const FILE_PLUGIN_CONFIGMAP = `
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: nuocollector-insights-file
+  labels:
+    nuodb.com/nuocollector-plugin: insights
```
vegichan, line 24: I will comment here, but it might apply to other places. I am not sure whether "insights" here is the right thing. I hope that it is not hardcoded. I assume that we pick up any configmap that has the `nuodb.com/nuocollector-plugin` label. What do we do with the value of the label?

butson: I believe it is hardcoded and we do nothing with the value just the key.  If it was configurable we could change the k8s-sidecar configuration to select based upon LABEL_VALUE allowing us to deploy different dynamic plugins to different components (admin or database) or different helm deployments.

test/integration/template_insights_test.go
sivanov-nuodb: I've updated the line as per your recommendation.

test/testlib/minikube_utilities.go
sivanov-nuodb: Okay, I've added another function as suggested.

test/testlib/constants.go
sivanov-nuodb: "Insights" is used here as we are deploying a dynamic output plugin similar to the one that Insights will deploy. Here we use "outputs.file" for simplicity to avoid installing InfluxDB. One of the things tested here is that a dynamic output plugin similar to the one deployed by Insights chart will be loaded.

I've removed "insights" in all the places where I think it's not relevant.

test/minikube/minikube_base_insights_test.go
sivanov-nuodb: >I assume that we pick up any configmap that has the nuodb.com/nuocollector-plugin label.
Nope. We pick up any configMap that match all of the following criteria:
* in the same namespace
* has a label of `nuodb.com/nuocollector-plugin`
* The value of the label is either `insights` OR specified in `nuocollector.plugins.admin` or `nuocollector.plugins.database`.

To be more specific the configMap selector is:
* for admin plugins: `"nuodb.com/nuocollector-plugin in ({{ template "admin.fullname" $ }}, insights)"`
* for database plugins: `"nuodb.com/nuocollector-plugin in ({{ template "database.fullname" $ }}, insights)"`

>If it was configurable we could change the k8s-sidecar configuration to select based upon LABEL_VALUE allowing us to deploy different dynamic plugins to different components (admin or database) or different helm deployments.

You can still deploy dynamic plugins for different components/deployments --- see https://github.com/nuodb/nuodb-collector#deploying-custom-plugins. I don't really see a reason for making the "insights" label configurable because there is no need for the user to change this. I personally would like users to use the helm charts to deploy new plugins vs creating configMaps by hand.

vegichan (approved):
good to go!

stable/database/README.md
```diff
@@ -310,6 +310,25 @@ The purpose of this section is to allow customisation of the names of the cluste
...
+The purpose of this section is to specify the NuoDB monitoring parameters.
+
+The following tables list the configurable parameters for the `nuocollector` option of the admin chart and their default values.
+
+| Parameter | Description | Default |
+| ----- | ----------- | ------ |
+| `enabled` | Whether to enable NuoDB monitoring using sidecar containers |`false`|
+| `image.registry` | NuoDB Collector container registry | `docker.io` |
+| `image.repository` | NuoDB Collector container image name |`nuodb/nuodb-collector`|
+| `image.tag` | NuoDB Collector container image tag | `1.0.1` |
```
vegichan, line 324: this value is incorrect. The chart uses 1.1.0

stable/admin/README.md
```diff
@@ -215,6 +215,24 @@ The purpose of this section is to allow customisation of the names of the cluste
...
+The purpose of this section is to specify the NuoDB monitoring parameters.
+
+The following tables list the configurable parameters for the `nuocollector` option of the admin chart and their default values.
+
+| Parameter | Description | Default |
+| ----- | ----------- | ------ |
+| `enabled` | Whether to enable NuoDB monitoring using sidecar containers |`false`|
+| `image.registry` | NuoDB Collector container registry | `docker.io` |
+| `image.repository` | NuoDB Collector container image name |`nuodb/nuodb-collector`|
+| `image.tag` | NuoDB Collector container image tag | `1.0.1` |
```
vegichan, line 229: same version issue here

test/minikube/minikube_base_collector_test.go
```diff
@@ -0,0 +1,200 @@
...
+	return false
+}
+
+func verifyCollectionForAdmin(t *testing.T, namespaceName string, app string) {
+	options := k8s.NewKubectlOptions("", "", namespaceName)
+	pods := k8s.ListPods(t, options, metav1.ListOptions{
+		LabelSelector: fmt.Sprintf("app=%s,component=admin", app),
+	})
+	for _, pod := range pods {
+		// Verfy that logfile nuocollector input plugin is collecting and outputing data to stdout
```
vegichan, line 56: typo in `verfy` and `outputing` here and beow.

stable/database/README.md
sivanov-nuodb: Good catch! I forgot to change this one.

test/minikube/minikube_base_collector_test.go
sivanov-nuodb: Thanks

stable/admin/templates/_sidecar.tpl
```diff
@@ -0,0 +1,86 @@
...
+    name: nuocollector-config
+  - name: log-volume
+    mountPath: /var/log/nuodb
+- name: nuocollector-config
+  image: {{ template "nuocollector.watcher" . }}
+  imagePullPolicy: {{ $.Values.nuocollector.watcher.pullPolicy }}
+  env:
+  - name: LABEL
+    value: "nuodb.com/nuocollector-plugin in ({{ template "admin.fullname" $ }}, insights)"
+  - name: FOLDER
```
butson, line 65: We might want to have an optional env variable LABEL_VALUE. and allow different labels for different databases or domains.   Probably not an issue at moment.  we'll see.

stable/admin/values.yaml
```diff
@@ -226,4 +226,20 @@ admin:
 
   # These annotations will pass throgh to the pod as supplied, useful for integrating 3rd party products such as Vault.
   podAnnotations: {}
-    # vault.hashicorp.com/agent-inject: true
\ No newline at end of file
+    # vault.hashicorp.com/agent-inject: true
+
+nuocollector:
+  enabled: false
```
butson, line 232: Are we disabling by default?   I thought we were going to enable once the sidecar container was publicly available.

sivanov-nuodb: If we enable it by default, then this means that on every installation we'll be collecting stats and dropping them on the floor, if `insights` is not installed. I believe it's better if we keep the default to `false` and enable it only if the iser have plans to install Insights or any other monitoring solution.

@jleslie85 , @vegichan What do you think?

vegichan: I prefer to pick the simpler default. Aka: don't enable stuff that people might not need and is not essential. In this case, I prefer `false`.

butson: With the old model of not having default collectors hardcoded into the container.   You could enable at any time.  So by using `enabled=true` you installed nuocollector but no plugins were installed.   So no additonal overhead and you could add the plugins at any time.    With the move to hardcoding the default plugins,  you are stuck to deciding at deployment time that you want to use nuocollector.  If you decide later to enable you have to shutdown  pods and redeploy them.

vegichan: yeah. You change a value, upgrade the helm deployments and you get a rolling upgrade. It's a trivial process.

test/integration/template_collector_test.go
```diff
@@ -0,0 +1,240 @@
...
+
+		for _, obj := range testlib.SplitAndRenderStatefulSet(t, output, 1) {
+			t.Logf("Inspecting database statefulset: %s", obj.Name)
+			checkSpecVolumes(t, obj.Spec.Template.Spec.Volumes, options, helmChartPath)
+			checkSidecarContainers(t, obj.Spec.Template.Spec.Containers, options, helmChartPath)
+		}
+	})
+
+	t.Run("testDatabaseDeploymentSidecars", func(t *testing.T) {
+		// Run RenderTemplate to render the template and inspect database statefulset
```
adriansuarez, line 149: The comment should say "database deployment".

test/integration/template_collector_test.go
```diff
@@ -0,0 +1,240 @@
...
+
+	options := &helm.Options{
+		SetValues: map[string]string{
+			"nuocollector.enabled":            "true",
+			"nuocollector.image.registry":     "docker.io",
+			"nuocollector.image.repository":   "nuodb/nuocd",
+			"nuocollector.image.tag":          "1.0.0",
+			"nuocollector.watcher.registry":   "docker.io",
+			"nuocollector.watcher.repository": "kiwigrid/k8s-sidecar",
+			"nuocollector.watcher.tag":        "latest",
```
adriansuarez, line 184: This would be future work, but it might be good to put the logic to detect the config map inside of nuocollector itself rather that relying on this kiwigrid thing.

test/minikube/minikube_base_collector_test.go
```diff
@@ -0,0 +1,167 @@
...
+			verifyCollectionForDatabase(t, namespaceName, fmt.Sprintf("%s-nuodb-%s-%s", databaseRealeaseName, "cluster0", "demo"), "demo")
+		})
+	})
+
+	t.Run("startDatabaseDaemonSet", func(t *testing.T) {
+		defer testlib.Teardown(testlib.TEARDOWN_DATABASE) // ensure resources allocated in called functions are released when this function exits
+		options.SetValues["database.enableDaemonSet"] = "true"
+		// Start only hotcopy SM daemonset
+		options.SetValues["database.sm.noHotCopy.enablePod"] = "false"
+		databaseRealeaseName := testlib.StartDatabase(t, namespaceName, admin0, &options)
```
adriansuarez, line 152: Nitpick: "release" is mispelled here and above.

adriansuarez (approved):
Looks good.

stable/admin/values.yaml
butson: It's not trivial in production when you have active connections / transactions that will get aborted. and the application logic might not be written to handle reconnect all that well.

NikTJ777: I think our message to users must be: "Insights should be enabled on _all_ databases - and only disabled if there are specific reasons."

So accordingly, I believe our default should reflect that - ie 'insights.enabled: true'.

Since engineering and support also find the output of insights to be very valuable in diagnosing issues, we should not be encouraging customers to deploy databases without insights.

I fully expect everyone in engineering, support - and customer organisations - will want all PROD databases to have insights enabled, therefore the overhead of insights needs to be acceptable for production.
If it isn't, then we need to address that, rather than disabling insights and ignoring the performance problems.

stable/admin/templates/_sidecar.tpl
```diff
@@ -0,0 +1,86 @@
...
+    {{- else -}}
+        {{- printf "%s/%s:%s" $registryName $repositoryName $tag -}}
+    {{- end -}}
+{{- else -}}
+    {{- printf "%s/%s:%s" $registryName $repositoryName $tag -}}
+{{- end -}}
+{{- end -}}
+
+{{- define "nuodb.sidecar" -}}
+{{- if $.Values.nuocollector }}
```
NikTJ777, line 48: Does this '{{ $.Values.nuocollector }}' have different behaviour to '{{ .Values.nuocollector }}' ?

If not, then could we please have all value references be consistent?

Currently, all references to values from the values.yaml file are in the form {{ .Values... }}
and only variables have the '$' prefix - since they must.

stable/admin/values.yaml
butson: We could have a default output to the log that can be override when deploying insights or some other server-side component.  The issue I see with defaulting to stdout or a log file is the data could get quite large.   the telegraf file output does have  a means to limit file size and rotate logs, keeping N number of logs.

stable/admin/templates/_sidecar.tpl
sivanov-nuodb: @NikTJ777 The `$` is mapped to the root scope when template execution begins and it does not change during template execution. It is required when root or parent scope needs to be accessed within a scoped block. In this specific case, I don't think it's needed. The PR is merged already so we may want to address this in subsequent change.

PR #162, opened 2020-11-10
Change GoLang module to v3

vegichan:
This is required to import the module and resolve it to 3.0.0 rather than 2.4.1.

Without it, go get fails like follows:
```
go get -u -v github.com/nuodb/nuodb-helm-charts@v3.0.0
go get github.com/nuodb/nuodb-helm-charts@v3.0.0: github.com/nuodb/nuodb-helm-charts@v3.0.0: invalid version: module contains a go.mod file, so major version must be compatible: should be v0 or v1, not v3
```

With the change, it can be done like this:
```
go get -u -v github.com/nuodb/nuodb-helm-charts/v3
go: downloading github.com/nuodb/nuodb-helm-charts/v3 v3.0.0
go: github.com/nuodb/nuodb-helm-charts/v3 upgrade => v3.0.0
```

We use this module in the insights project.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/nuodb/nuodb-helm-charts
+module github.com/nuodb/nuodb-helm-charts/v3
 
 go 1.14
 
--- a/test/integration/database_marshall_test.go
+++ b/test/integration/database_marshall_test.go
@@ -1,7 +1,7 @@
 package integration
 
 import (
-	"github.com/nuodb/nuodb-helm-charts/test/testlib"
+	"github.com/nuodb/nuodb-helm-charts/v3/test/testlib"
 	"github.com/stretchr/testify/assert"
 
 	"testing"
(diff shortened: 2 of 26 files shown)


sivanov-nuodb (commented):
You may need to update also the import sections to "github.com/nuodb/nuodb-helm-charts/v3/test/testlib". 

Not sure if this is the cause for the test failures:
```
# github.com/nuodb/nuodb-helm-charts/test/testlib
../../../../pkg/mod/github.com/nuodb/nuodb-helm-charts@v2.4.1+incompatible/test/testlib/core_recovery.go:44:41: not enough arguments in call to k8s.NewKubectlOptions
	have (string, string)
	want (string, string, string)

vegichan:
@sivanov-nuodb yeah! I learned that the hard way 👍

sivanov-nuodb (approved):
Thanks for changing this!

PR #166, opened 2020-11-23
[HELM-191] Use single-server readiness probe

adriansuarez:
Previously the readiness probe for Admin Pods checked for full-server
connectivity and that all Admin processes agree on a Raft leader. Since these
are not properties of a particular Admin process, but of the entire domain,
they were not appropriate checks for a readiness probe (e.g. admin-1 could show
as not ready because admin-2 is having connectivity issues).

This change adds a readinessprobe script as a ConfigMap that uses the new
'nuocmd check server' (singular) subcommand that will be available in 4.1.2 and
4.2, and falls back to the 'nuocmd check servers' (plural) subcommand that was
previously used as the readiness probe if the new subcommand is not supported.

### DIFF
--- a/stable/admin/files/readinessprobe
+++ b/stable/admin/files/readinessprobe
@@ -3,8 +3,8 @@
 nuocmd_fallback() {
     # run nuocmd subcommand and suppress parse errors due to the command
     # containing unsupported arguments or the subcommand not being supported at
-    # all by the available version of nuocmd; for all non-parse errors or
-    # successful invocations, emit command output and exit
+    # all by the available version of nuocmd; if the command succeeds or fails
+    # with a non-parse error, emit command output and exit
 
     # capture stdout and stderr and get exit code
     out="$(nuocmd "$@" 2>&1)"
@@ -13,8 +13,8 @@ nuocmd_fallback() {
     # nuocmd returns exit code 2 if there is a parse error due to unrecognized
     # subcommand or arguments
     if [ $ret != 2 ]; then
-        # failure is not due to parse error or the command succeeded; print
-        # command output and exit
+        # the command succeeded or failed with a non-parse error; emit command
+        # output and exit
         echo "$out"
         exit $ret
     fi
--- a/stable/admin/files/readinessprobe
+++ b/stable/admin/files/readinessprobe
@@ -1,0 +1,25 @@
+#!/bin/sh
+
+# 'nuocmd check server' is introduced in versions >4.1.2; this checks that the
+# --api-server satisfies the following conditions:
+#   1. It is ACTIVE and able to service REST requests.
+#   2. It is able to ping its own advertised address, i.e. altAddr in
+#      nuoadmin.conf, which is derived from the --alt-address argument of
+#      'nuodocker start admin'.
+#   3. It has the same commit index as the current Raft leader, which means
+#      that its Raft state is current.
+out="$(nuocmd check server --check-active --check-connected --check-converged 2>&1)"
+ret=$?
+
+# nuocmd returns exit code 2 if there is a parse error due to unrecognized
+# subcommand or arguments
+if [ $ret != 2 ]; then
+    # failure is not due to the parse error, so print error message and exit
+    echo "$out"
+    exit $ret
+fi
+
+# 'nuocmd check server' is unsupported; use 'nuocmd check servers' (plural)
+# with no arguments, which simply checks that the --api-server is ACTIVE and
+# able to service REST requests
+nuocmd check servers
--- a/stable/admin/files/readinessprobe
+++ b/stable/admin/files/readinessprobe
@@ -1,23 +1,35 @@
 #!/bin/sh
 
-# 'nuocmd check server' is introduced in versions >4.1.2; this checks that the
+nuocmd_fallback() {
+    # run nuocmd subcommand and suppress parse errors due to the command
+    # containing unsupported arguments or the subcommand not being supported at
+    # all by the available version of nuocmd; for all non-parse errors or
+    # successful invocations, emit command output and exit
+
+    # capture stdout and stderr and get exit code
+    out="$(nuocmd "$@" 2>&1)"
+    ret=$?
+
+    # nuocmd returns exit code 2 if there is a parse error due to unrecognized
+    # subcommand or arguments
+    if [ $ret != 2 ]; then
+        # failure is not due to parse error or the command succeeded; print
+        # command output and exit
+        echo "$out"
+        exit $ret
+    fi
+}
+
+# 'nuocmd check server' is introduced in versions >4.1.1; this checks that the
 # --api-server satisfies the following conditions:
 #   1. It is ACTIVE and able to service REST requests.
 #   2. It is able to ping its own advertised address, i.e. altAddr in
 #      nuoadmin.conf, which is derived from the --alt-address argument of
-#      'nuodocker start admin'.
+#      'nuodocker start admin' and the NUODB_ALT_ADDRESS environment variable
+#      for the nuoadmin entrypoint script.
 #   3. It has the same commit index as the current Raft leader, which means
-#      that its Raft state is current.
-out="$(nuocmd check server --check-active --check-connected --check-converged 2>&1)"
-ret=$?
-
-# nuocmd returns exit code 2 if there is a parse error due to unrecognized
-# subcommand or arguments
-if [ $ret != 2 ]; then
-    # failure is not due to the parse error, so print error message and exit
-    echo "$out"
-    exit $ret
-fi
+#      that its Raft state is up-to-date.
+nuocmd_fallback check server --check-active --check-connected --check-converged
 
 # 'nuocmd check server' is unsupported; use 'nuocmd check servers' (plural)
 # with no arguments, which simply checks that the --api-server is ACTIVE and
--- a/stable/admin/files/readinessprobe
+++ b/stable/admin/files/readinessprobe
@@ -31,7 +31,7 @@ nuocmd_fallback() {
 #      that its Raft state is up-to-date.
 nuocmd_fallback check server --check-active --check-connected --check-converged
 
-# 'nuocmd check server' is unsupported; use 'nuocmd check servers' (plural)
-# with no arguments, which simply checks that the --api-server is ACTIVE and
-# able to service REST requests
-nuocmd check servers
+# 'nuocmd check server' is unsupported; use the 'nuocmd check servers' (plural)
+# subcommand, which was used as the readiness probe in releases of
+# nuodb-helm-charts <=3.0.0
+nuocmd check servers --check-active --check-connected --check-leader
--- a/test/minikube/minikube_rolling_upgrade_test.go
+++ b/test/minikube/minikube_rolling_upgrade_test.go
@@ -60,11 +60,10 @@ func TestAdminReadinessProbe(t *testing.T) {
 	output, err = k8s.RunKubectlAndGetOutputE(t, options, "exec", admin1, "--", "readinessprobe")
 	require.NoError(t, err, fmt.Sprintf("readinessprobe failed: %s", output))
 
-	// kill admin-0 and delete its PVC so that it bootstraps a disjoint
-	// domain and refuses messages from admin-1
-	k8s.RunKubectl(t, options, "delete", "pod", admin0)
-	k8s.RunKubectl(t, options, "delete", "pvc", "raftlog-"+admin0)
-	testlib.AwaitNoPods(t, namespaceName, admin0)
+	// delete Raft log on admin-0 and kill admin-0 so that it bootstraps a
+	// disjoint domain when it is restarted and refuses messages from
+	// admin-1
+	k8s.RunKubectl(t, options, "exec", admin0, "--", "bash", "-c", "rm -f /var/opt/nuodb/raftlog && kill -9 1")
 
 	// make sure readinessprobe on admin-1 eventually fails, because there
 	// is no leader for it to converge with
@@ -76,9 +75,10 @@ func TestAdminReadinessProbe(t *testing.T) {
 		require.NoError(t, err)
 		require.Equal(t, 1, code)
 		return true
-	}, 300*time.Second)
+	}, 60*time.Second)
 
-	// make sure 'nuocmd check servers' (plural) is successful
+	// make sure 'nuocmd check servers' (plural) is successful, since it
+	// only checks that the --api-server is ACTIVE
 	output, err = k8s.RunKubectlAndGetOutputE(t, options, "exec", admin1, "--", "nuocmd", "check", "servers")
 	require.NoError(t, err, fmt.Sprintf("'nuocmd check servers' failed: %s", output))
 }
--- a/test/minikube/minikube_rolling_upgrade_test.go
+++ b/test/minikube/minikube_rolling_upgrade_test.go
@@ -33,6 +33,8 @@ func verifyAllProcessesRunning(t *testing.T, namespaceName string, adminPod stri
 }
 
 func TestAdminReadinessProbe(t *testing.T) {
+	// TODO: remove this whenever the image tested in nuodb-helm-charts CI
+	// supports 'nuocmd check server' (singular)
 	if os.Getenv("NUODB_DEV") != "true" {
 		t.Skip("'nuocmd check server' is not supported in released versions")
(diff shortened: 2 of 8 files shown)


sivanov-nuodb (commented):
@adriansuarez I can't see where in `statefulset.yaml` the actual container readiness probe is changed. Am I missing something?

adriansuarez:
> @adriansuarez I can't see where in `statefulset.yaml` the actual container readiness probe is changed. Am I missing something?

Lines 147-149 and 210-214 are adding the volume mount for the new ConfigMap.

test/minikube/minikube_rolling_upgrade_test.go
```diff
@@ -31,6 +32,114 @@ func verifyAllProcessesRunning(t *testing.T, namespaceName string, adminPod stri
...
+			"admin.replicas": "2",
+		},
+	}, 2, "")
+	adminStatefulSet := helmChartReleaseName + "-nuodb-cluster0"
+	admin0 := adminStatefulSet + "-0"
+	admin1 := adminStatefulSet + "-1"
+
+	// make sure readinessprobe succeeds on both Admin processes
+	options := k8s.NewKubectlOptions("", "", namespaceName)
+	output, err := k8s.RunKubectlAndGetOutputE(t, options, "exec", admin0, "--", "readinessprobe")
```
sivanov-nuodb, line 60: It's it better to check that the pod is READY in addition/instead of running the prob directly?

adriansuarez: I think `testlib.StartAdmin()` blocks until the Admin Pods are READY, but I can check that explicitly here.

sivanov-nuodb:
> > @adriansuarez I can't see where in `statefulset.yaml` the actual container readiness probe is changed. Am I missing something?
> 
> Lines 147-149 and 210-214 are adding the volume mount for the new ConfigMap.

Yes, but I was expecting to see this code changed as well:
```
        readinessProbe:
          initialDelaySeconds: 30
          periodSeconds: 15
          exec:
            command: [ "nuodocker", "check", "servers", "--check-connected", "--check-active", "--check-leader" ]
```

adriansuarez:
> > > @adriansuarez I can't see where in `statefulset.yaml` the actual container readiness probe is changed. Am I missing something?
> > 
> > 
> > Lines 147-149 and 210-214 are adding the volume mount for the new ConfigMap.
> 
> Yes, but I was expecting to see this code changed as well:
> 
> ```
>         readinessProbe:
>           initialDelaySeconds: 30
>           periodSeconds: 15
>           exec:
>             command: [ "nuodocker", "check", "servers", "--check-connected", "--check-active", "--check-leader" ]
> ```

Ah, of course. Good catch!

stable/admin/files/readinessprobe
```diff
@@ -0,0 +1,37 @@
...
+#      'nuodocker start admin' and the NUODB_ALT_ADDRESS environment variable
+#      for the nuoadmin entrypoint script.
+#   3. It has the same commit index as the current Raft leader, which means
+#      that its Raft state is up-to-date.
+nuocmd_fallback check server --check-active --check-connected --check-converged
+
+# 'nuocmd check server' is unsupported; use 'nuocmd check servers' (plural)
+# with no arguments, which simply checks that the --api-server is ACTIVE and
+# able to service REST requests
+nuocmd check servers
```
vegichan, line 37: I just wanna point out that in pre-4.1.1 NuoDB this is a change in behavior. This probe is laxer than it used to be.

adriansuarez:
Updated to actually use readiness probe, and also checking that the Pods become Ready.

test/minikube/minikube_rolling_upgrade_test.go
vegichan: you dont need anything between lines 58 and here. All of that is done as part of `StartAdmin`

stable/admin/files/readinessprobe
adriansuarez: Yes, this leaves out `--check-connected` and `--check-leader`, which actually check all processes in the domain, not just the process being probed, as I mentioned in the PR summary:

> Previously the readiness probe for Admin Pods checked for full-server connectivity and that all Admin processes agree on a Raft leader. Since these are not properties of a particular Admin process, but of the entire domain, they were not appropriate checks for a readiness probe (e.g. admin-1 could show as not ready because admin-2 is having connectivity issues).

We'll see if this actually breaks anything.

test/minikube/minikube_rolling_upgrade_test.go
```diff
@@ -31,6 +32,119 @@ func verifyAllProcessesRunning(t *testing.T, namespaceName string, adminPod stri
...
+	// make sure both Admin Pods become Ready
+	testlib.AwaitPodUp(t, namespaceName, admin0, 120*time.Second)
+	testlib.AwaitPodUp(t, namespaceName, admin1, 120*time.Second)
+
+	// make sure direct invocation of readinessprobe script succeeds on both
+	// Admin processes
+	options := k8s.NewKubectlOptions("", "", namespaceName)
+	output, err := k8s.RunKubectlAndGetOutputE(t, options, "exec", admin0, "--", "readinessprobe")
+	require.NoError(t, err, "readinessprobe failed: %s", output)
+	output, err = k8s.RunKubectlAndGetOutputE(t, options, "exec", admin1, "--", "readinessprobe")
```
vegichan, line 67: I dont think that you can reach this code. If the readiness probe fails here, it would have failed during `StartAdmin` and the test would have been aborted.

sivanov-nuodb (approved):
Thanks!

test/minikube/minikube_rolling_upgrade_test.go
```diff
@@ -31,6 +32,119 @@ func verifyAllProcessesRunning(t *testing.T, namespaceName string, adminPod stri
...
+
+	// make sure readinessprobe on admin-1 eventually fails, because there
+	// is no leader for it to converge with even if the backwards-compatible
+	// 'nuocmd check servers' (plural) succeeds; this verifies that the
+	// fallback behavior does not apply to actual failures of the readiness
+	// check
+	testlib.Await(t, func() bool {
+		// make sure 'nuocmd check servers' (plural) is successful, since it
+		// only checks that the --api-server is ACTIVE
+		output, err = k8s.RunKubectlAndGetOutputE(t, options, "exec", admin1, "--", "nuocmd", "check", "servers")
```
vegichan, line 92: do you need the output here? I would use `RunKubectlE`

test/minikube/minikube_rolling_upgrade_test.go
```diff
@@ -31,6 +32,119 @@ func verifyAllProcessesRunning(t *testing.T, namespaceName string, adminPod stri
...
+		if err != nil {
+			return false
+		}
+
+		// make sure readinessprobe fails
+		output, err = k8s.RunKubectlAndGetOutputE(t, options, "exec", admin1, "--", "readinessprobe")
```
vegichan, line 98: same here, are you doing anything with the `output`?

test/testlib/minikube_utilities.go
```diff
@@ -733,7 +733,13 @@ func GetAppLog(t *testing.T, namespace string, podName string, fileNameSuffix st
 
 	writer := io.Writer(f)
 
-	reader := getAppLogStream(t, namespace, podName, podLogOptions)
+	reader, err := getAppLogStreamE(t, namespace, podName, podLogOptions)
```
vegichan, line 736: how did you hit an issue with this? We always ensure that a pod is running before we call GetAppLog on it.

test/testlib/minikube_utilities.go
```diff
@@ -923,17 +952,19 @@ func VerifyAdminKvSetAndGet(t *testing.T, podName string, namespaceName string)
 	)
 	require.NoError(t, err, "Could not set KV value")
 	elapsed := time.Since(start)
-
-	require.True(t, elapsed.Seconds() < 2.0, fmt.Sprintf("KV set took longer than 2s: %s", elapsed))
+	if elapsed.Seconds() > 2.0 {
```
vegichan, line 955: @NikTJ777 wanted to have these as regular tests to verify that RAFT is not unexpectedly slow. Nobody looks at the log output.

test/minikube/minikube_rolling_upgrade_test.go
adriansuarez: It looks like `testlib.StartAdmin()` does explicitly check that the Pods are Ready. But this did not catch the fact that `readinessprobe` was not being used at all as Stanimir noticed, probably because the readiness probe is invoked on a schedule (30s, 15s, 15s, ...), which leaves plenty of time for stronger `nuocmd check` command to also be satisfied.

test/testlib/nuodb_admin_utilities.go
```diff
@@ -124,7 +124,7 @@ func StartAdminTemplate(t *testing.T, options *helm.Options, replicaCount int, n
 		AddTeardown(TEARDOWN_ADMIN, func() {
 			_, err := k8s.GetPodE(t, kubectlOptions, adminName)
 			if err != nil {
-				t.Logf("Admin pod '%s' is not available and logs can not be retrieved")
+				t.Logf("Admin pod '%s' is not available and logs can not be retrieved", adminName)
```
vegichan, line 127: this should return here and not proceed to the other two steps on lines 130-131. I mean, if there is no pod, there is no point in executing those.

test/testlib/minikube_utilities.go
adriansuarez: `TestAdminReadinessProbe` kills the `admin-0` Pod to verify that the `readinessprobe` script eventually fails on `admin-1`. We do not wait for `admin-0` to come back up, because we do not care about its state, we only want its presence in the Raft membership to prevent a leader from being elected.

test/testlib/minikube_utilities.go
adriansuarez: It isn't an error for a Raft command to take longer than 2 seconds, and we don't really have any control over what kind of hardware the tests run on in Travis CI. I think catching performance regressions is outside the scope of a continuous test suite.

test/testlib/nuodb_admin_utilities.go
adriansuarez: All I did was add the Pod name that has a placeholder in the format string.

vegichan: I know, but the lack of a return seems to be the reason why you had to make all those changes in GetAppLog

test/minikube/minikube_rolling_upgrade_test.go
adriansuarez: Will do.

test/minikube/minikube_rolling_upgrade_test.go
adriansuarez: Will change to `RunKubectlE`.

test/testlib/minikube_utilities.go
vegichan: the issue here being that you terminate the test before admin-0 comes back up? That sounds unlikely.

test/minikube/minikube_rolling_upgrade_test.go
vegichan: StartAdmin executes 
```
		// first await could be pulling the image from the repo
		AwaitPodUp(t, namespaceName, adminName, 300*time.Second)
```

on every pod. This waits for the pod to be Ready. Which by definition means that the readiness probe had to succeed.

What am I missing?

test/testlib/nuodb_admin_utilities.go
adriansuarez: The Pod existed but was in PodInitializing state.

test/minikube/minikube_rolling_upgrade_test.go
```diff
@@ -31,6 +32,119 @@ func verifyAllProcessesRunning(t *testing.T, namespaceName string, adminPod stri
...
+	// elected when it is restarted
+	if admin1 == strings.TrimSpace(leader) {
+		k8s.RunKubectl(t, options, "delete", "pod", admin1)
+	}
+
+	// make sure readinessprobe on admin-1 eventually fails, because there
+	// is no leader for it to converge with even if the backwards-compatible
+	// 'nuocmd check servers' (plural) succeeds; this verifies that the
+	// fallback behavior does not apply to actual failures of the readiness
+	// check
```
vegichan, line 88: so with the old probe, the disjoint state would result in a visible error. With the new (laxer) probe this results in a READY but unusable domain?

test/testlib/minikube_utilities.go
adriansuarez: That's what happened. It only takes between 2-5 seconds for `admin-1` to become a CANDIDATE after `admin-0` is killed, and all of the remaining polling loops would complete in that case. From my experience, 2-5 seconds is not enough time for an Admin Pod to start up, but maybe that is only an issue when running on my machine.

test/testlib/nuodb_admin_utilities.go
vegichan: while you are here, can you please also add a return :)?

test/testlib/minikube_utilities.go
vegichan: is this in the case when -1 is not the LEADER and hence you don't kill it?

test/minikube/minikube_rolling_upgrade_test.go
adriansuarez: The test is supposed to verify expected behavior. Why shouldn't we verify that `readinessprobe` actually passes as expected? It adds nothing to the running time.

test/testlib/minikube_utilities.go
adriansuarez: Yes, and that's the most likely scenario, since `admin-0` becomes initial leader and it is unlikely for a new leader election to be triggered while it's running.

test/testlib/nuodb_admin_utilities.go
adriansuarez: Will do.

test/minikube/minikube_rolling_upgrade_test.go
vegichan: I guess...

test/minikube/minikube_rolling_upgrade_test.go
adriansuarez: You mean if the NuoDB version was <=4.1.1, `admin-1` would fall back to the weaker health-check and would show as Ready. You are correct.

Do you consider this an issue because customers rely on the status reported by K8s to determine the health of the domain? In that case, I think the old health-check was just as likely to yield false positives and cause customers to try to "fix" Admin processes that did not have anything wrong with them, but that were able to detect issues with some peer. The old health-check is actually more of a StatefulSet probe (which doesn't actually exist).

We'll have to let Nik weigh in on this, and describe any issues he observed with the old health-check.

test/testlib/minikube_utilities.go
NikTJ777: Actually, the KV timeout test was _not_ to test for performance regressions, but to bail out of the test with a useful explanation in the event that the hardware that Travis has allocated to us is just too slow.

The goal is to avoid chasing our tails trying to diagnose (false) failures in tests that are actually caused by the hardware being so slow that the admin layer and KV store are causing otherwise correct tests to timeout and fail.

test/minikube/minikube_rolling_upgrade_test.go
adriansuarez: Updated the test to check that output, and also to account for the new fallback behavior.

test/minikube/minikube_rolling_upgrade_test.go
adriansuarez: This is no longer relevant now that the test has been updated to account for the new fallback behavior.

test/minikube/minikube_rolling_upgrade_test.go
adriansuarez: This conversations is no longer relevant after the latest round of changes.

adriansuarez:
Updated the branch to fall back to the previous subcommand (`nuocmd check servers --check-active --check-connected --check-leader`) if the available `nuocmd` does not support `nuocmd check server`, and also make changes based on review feedback.

vegichan (approved):
thanks for the updates

PR #168, opened 2020-12-03
Transparent Data Encryption support for NuoDB Helm Charts

sivanov-nuodb:
**Changes**
- User now can provide a pre-created secret name per database containing TDE passwords
- The above secrets will be mounted in AP, HCSM and non-HCSM pods
- New service in NuoAdmin will be used to read the TDE password files and enforce the passwords in admin layer
- `nuosm` has been updated to use `nuodocker restore archive` for the actual restore as TDE passwords are needed when invoking `nuoarchive restore` against encrypted backupset

**Testing**
- Integration tests for admin and database chart changes
- Functional tests exercising restore-in-place with TDE enabled database (tagged with `enterprise`)

**Usage**
These are example steps to deploy NuoDB with TDE enabled using Kubernetes secret:
```bash
kubectl create secret generic demo-tde-secret -n nuodb \
  --from-literal target='topSecret'

helm install -n nuodb admin stable/admin \
  --set admin.tde.secrets.demo=demo-tde-secret

helm install -n nuodb database stable/database \
  --set database.sm.resources.requests.cpu=500m \
  --set database.sm.resources.requests.memory=256Mi \
  --set database.te.resources.requests.cpu=500m \
  --set database.te.resources.requests.memory=256Mi \
  --set admin.tde.secrets.demo=demo-tde-secret

kubectl exec -n nuodb admin-nuodb-cluster0-0 -- bash -c \
  'echo "alter database change encryption type AES128;" | /opt/nuodb/bin/nuosql demo --user dba --password secret'
```

**Password rotation**
During password rotation user will need to provide the new target storage password along with optional historical passwords. The rotation will be performed automatically, however, it will be asynchronous. Users can check if the password rotation is successful by using `nuocmd check data-encryption` command.

```bash
# Update the secret and add the old password as historical; 
# it will be needed to restore any backupset encrypted
# with that password

kubectl create secret generic demo-tde-secret -n nuodb \
  --from-literal target='superSecret' \
  --from-literal historical-20201110='topSecret' \
  --dry-run=client -o yaml | kubectl apply -f -

# Wait for the secret being updated inside the pods and 
# the new password to be propagated to all engines; 
# in practice, this takes ~60sec and depends on the
# kubelet sync period + kubelet cache propagation delay

kubectl exec -n nuodb admin-nuodb-cluster0-0 -- \
  nuocmd check data-encryption --db-name demo --password superSecret
```

For more information, please check "Transparent Data Encryption in cloud" wiki page.


test/minikube/minikube_base_tde_test.go
```diff
@@ -0,0 +1,213 @@
+// +build enterprise
```
adriansuarez, line 1: Does this actually require EE, i.e. do these tests start multiple SMs? Or do we just not want to run this test in GitHub CI because it uses unreleased functionality? I introduced the NUODB_DEV environment variable for the latter case.

test/minikube/minikube_base_tde_test.go
```diff
@@ -0,0 +1,213 @@
...
+	namespaceName := fmt.Sprintf("admin-cold-start-with-tde-%s", randomSuffix)
+	testlib.CreateNamespace(t, namespaceName)
+	kubectlOptions := k8s.NewKubectlOptions("", "", namespaceName)
+	options.KubectlOptions = kubectlOptions
+
+	password := strings.ToLower(random.UniqueId())
+	applyStoragePasswordSecret(t, namespaceName, options.SetValues["admin.tde.secrets.demo"], []string{password})
+
+	defer testlib.Teardown(testlib.TEARDOWN_ADMIN)
+	helmChartReleaseName, _ := testlib.StartAdmin(t, &options, 1, namespaceName)
```
adriansuarez, line 74: What happens if the secret is created after the Admin layer is started?

sivanov-nuodb: The pods will stick in "Init" state with the following warning events:
``` 
Warning  FailedMount       5s (x5 over 13s)  kubelet            MountVolume.SetUp failed for volume "tde-volume-demo" : secret "demo-tde-secrett" not found
Warning  FailedMount       5s (x5 over 13s)  kubelet            MountVolume.SetUp failed for volume "tde-volume-demo2" : secret "demo2-tde-secrett" not found
```
They need to be restarted after needed secrets are created.

test/minikube/minikube_base_tde_test.go
sivanov-nuodb: Good point! The test doesn't require enterprise functionality so we could leverage of `NUODB_DEV` variable.

test/minikube/minikube_base_tde_test.go
```diff
@@ -0,0 +1,225 @@
...
+	testlib.Await(t, func() bool {
+		return testlib.GetStringOccurrenceInLog(t, namespaceName, admin0,
+			"Successfully updated storage passwords for dbName=demo", &corev1.PodLogOptions{}) >= 1
+	}, 30*time.Second)
+
+	defer testlib.Teardown(testlib.TEARDOWN_DATABASE)
+	databaseChartName := testlib.StartDatabase(t, namespaceName, admin0, &options)
+	// Enable TDE on database level
+	output, _ := testlib.RunSQL(t, namespaceName, admin0, "demo", "alter database change encryption type AES128")
+	assert.False(t, strings.Contains(output, "Error"), "Failed to enable TDE: %s", output)
```
vegichan, line 91: this should be `require`

test/minikube/minikube_base_tde_test.go
```diff
@@ -0,0 +1,225 @@
...
+	testlib.Await(t, func() bool {
+		return testlib.GetStringOccurrenceInLog(t, namespaceName, admin0,
+			"Successfully updated storage passwords for dbName=demo", &corev1.PodLogOptions{}) >= 1
+	}, 30*time.Second)
+
+	// verify that restarting an SM pod will succeed to start encrypted archive
+	opt := testlib.GetExtractedOptions(&options)
+	smPodNameTemplate := fmt.Sprintf("sm-%s-nuodb-%s-%s", databaseChartName, opt.ClusterName, opt.DbName)
+	smPodName0 := testlib.GetPodName(t, namespaceName, smPodNameTemplate)
+	testlib.DeletePod(t, namespaceName, "pod/"+smPodName0)
```
vegichan, line 109: I wish we had the ability to start multiple SMs with the CE license. An SM restart would not take down the whole database...

test/minikube/minikube_base_tde_test.go
```diff
@@ -0,0 +1,225 @@
...
+	options := helm.Options{
+		SetValues: map[string]string{
+			"admin.tde.secrets.demo":                "demo-tde-secret",
+			"database.name":                         "demo",
+			"database.sm.resources.requests.cpu":    testlib.MINIMAL_VIABLE_ENGINE_CPU,
+			"database.sm.resources.requests.memory": testlib.MINIMAL_VIABLE_ENGINE_MEMORY,
+			"database.te.resources.requests.cpu":    testlib.MINIMAL_VIABLE_ENGINE_CPU,
+			"database.te.resources.requests.memory": testlib.MINIMAL_VIABLE_ENGINE_MEMORY,
+			"backup.persistence.enabled":            "true",
+			"backup.persistence.size":               "1Gi",
```
vegichan, line 132: is this not the default?

test/minikube/minikube_base_tde_test.go
```diff
@@ -0,0 +1,225 @@
...
+	testlib.Await(t, func() bool {
+		return testlib.GetStringOccurrenceInLog(t, namespaceName, admin0,
+			"Successfully updated storage passwords for dbName=demo", &corev1.PodLogOptions{}) >= 1
+	}, 30*time.Second)
+
+	defer testlib.Teardown(testlib.TEARDOWN_DATABASE)
+	databaseChartName := testlib.StartDatabase(t, namespaceName, admin0, &options)
+
+	// Generate diagnose in case this test fails
+	testlib.AddTeardown(testlib.TEARDOWN_DATABASE, func() {
```
vegichan, line 157: you might wanna check out the addDiagnosticTeardown. I am OK with what is here.

test/minikube/minikube_base_tde_test.go
```diff
@@ -0,0 +1,225 @@
...
+		if t.Failed() {
+			testlib.GetDiagnoseOnTestFailure(t, namespaceName, admin0)
+			testlib.RecoverCoresFromEngine(t, namespaceName, "te", "demo-log-te-volume")
+		}
+	})
+
+	// Enable TDE on database level
+	output, _ := testlib.RunSQL(t, namespaceName, admin0, "demo", "alter database change encryption type AES128")
+	assert.False(t, strings.Contains(output, "Error"), "Failed to enable TDE: %s", output)
```
vegichan, line 166: require

test/minikube/minikube_base_tde_test.go
```diff
@@ -0,0 +1,225 @@
...
+	smPodNameTemplate := fmt.Sprintf("sm-%s-nuodb-%s-%s", databaseChartName, opt.ClusterName, opt.DbName)
+
+	tePodName := testlib.GetPodName(t, namespaceName, tePodNameTemplate)
+	go testlib.GetAppLog(t, namespaceName, tePodName, "_pre-restart", &corev1.PodLogOptions{Follow: true})
+
+	smPodName0 := testlib.GetPodName(t, namespaceName, smPodNameTemplate)
+	go testlib.GetAppLog(t, namespaceName, smPodName0, "_pre-restart", &corev1.PodLogOptions{Follow: true})
+
+	// Take encrypted backup
+	backupset := testlib.BackupDatabase(t, namespaceName, smPodName0, opt.DbName, "full", opt.ClusterName)
```
vegichan, line 193: I am assuming that BackupDatabase is synchronous?

test/minikube/minikube_base_tde_test.go
```diff
@@ -0,0 +1,225 @@
...
+	applyStoragePasswordSecret(t, namespaceName, options.SetValues["admin.tde.secrets.demo"], []string{passwordNew, password})
+	testlib.Await(t, func() bool {
+		err := k8s.RunKubectlE(t, kubectlOptions,
+			"exec", admin0, "--",
+			"nuocmd", "check", "data-encryption",
+			"--db-name", opt.DbName,
+			"--password", passwordNew,
+		)
+		return err == nil
+	}, 90*time.Second)
```
vegichan, line 216: I would refactor this into its own method.

vegichan (approved):
besides minor comments, this is good to go

stable/admin/values.yaml
```diff
@@ -228,6 +228,29 @@ admin:
...
+  # secrets. TDE secrets for all databases should be created and 
+  # provided before deploying NuoDB. For example:
+  #   kubectl create secret generic demo-tde-secret -n nuodb \
+  #     --from-literal target='topSecret'
+  #
+  # Note: TDE should be enabled on database layer using
+  #   SQL> alter database change encryption type AES128;
+  #
+  # Storage password rotation is performed by updating te secret and 
+  # providing current, new target and historical (optional) passwords.
```
adriansuarez, line 242: This needs updating since we don't require the current password anymore.

stable/database/values.yaml
```diff
@@ -91,6 +91,29 @@ admin:
...
+  # secrets. TDE secrets for all databases should be created and 
+  # provided before deploying NuoDB. For example:
+  #   kubectl create secret generic demo-tde-secret -n nuodb \
+  #     --from-literal target='topSecret'
+  #
+  # Note: TDE should be enabled on database layer using
+  #   SQL> alter database change encryption type AES128;
+  #
+  # Storage password rotation is performed by updating te secret and 
+  # providing current, new target and historical (optional) passwords.
```
adriansuarez, line 105: Same comment here about `current` no longer being used.

test/minikube/minikube_base_tde_test.go
```diff
@@ -0,0 +1,225 @@
...
+	applyStoragePasswordSecret(t, namespaceName, options.SetValues["admin.tde.secrets.demo"], []string{password})
+
+	defer testlib.Teardown(testlib.TEARDOWN_ADMIN)
+	helmChartReleaseName, _ := testlib.StartAdmin(t, &options, 1, namespaceName)
+	admin0 := fmt.Sprintf("%s-nuodb-cluster0-0", helmChartReleaseName)
+	testlib.Await(t, func() bool {
+		return testlib.GetStringOccurrenceInLog(t, namespaceName, admin0,
+			"Successfully updated storage passwords for dbName=demo", &corev1.PodLogOptions{}) >= 1
+	}, 30*time.Second)
+
```
adriansuarez, line 86: Can you assert that `nuocmd check data-encryption --password <the password>` is successful. You could also specify `--timeout` and eliminate the need for polling in the test.

test/minikube/minikube_base_tde_test.go
```diff
@@ -0,0 +1,225 @@
...
+	adminPod := testlib.GetPod(t, namespaceName, admin0)
+	testlib.DeletePod(t, namespaceName, "pod/"+admin0)
+	testlib.AwaitPodObjectRecreated(t, namespaceName, adminPod, 30*time.Second)
+	testlib.AwaitPodUp(t, namespaceName, admin0, 300*time.Second)
+	testlib.AwaitAdminFullyConnected(t, namespaceName, admin0, 1)
+	testlib.Await(t, func() bool {
+		return testlib.GetStringOccurrenceInLog(t, namespaceName, admin0,
+			"Successfully updated storage passwords for dbName=demo", &corev1.PodLogOptions{}) >= 1
+	}, 30*time.Second)
+
```
adriansuarez, line 104: Same comment as above about `nuocmd check data-encryption`.

test/minikube/minikube_base_tde_test.go
adriansuarez, line 216: I notice that this is longer than the 30 seconds you use elsewhere. Is because of the time it takes K8s to update the secret?

adriansuarez (approved):
Looks good.

test/minikube/minikube_base_tde_test.go
sivanov-nuodb: Actually, this was a leftover and I don't think we have such options anymore. I will remove them.

test/minikube/minikube_base_tde_test.go
sivanov-nuodb: Good call! All will change it.

test/minikube/minikube_base_tde_test.go
sivanov-nuodb: Yes, it is.

test/minikube/minikube_base_tde_test.go
sivanov-nuodb: Will do.

stable/admin/values.yaml
sivanov-nuodb: Good catch! It looks like I missed this one after removing the current from the service.

test/minikube/minikube_base_tde_test.go
adriansuarez: Why would it take down the database? Only the SM that exceeds the limit on number of SMs would be rejected, but the original one would remain running.

test/minikube/minikube_base_tde_test.go
sivanov-nuodb: Okay, I will add a one-shot `check data-encryption` to assert that the password is set.
Unfortunately, we don't support `--timeout` with this command. Certainly, something that we could add.

test/minikube/minikube_base_tde_test.go
sivanov-nuodb: Exactly --- it takes ~60sec for Kubernetes to update the secret inside the pods. Once the files mounted from the secret are updated, the resync will be triggered almost instantly because of the IO notifications.

test/minikube/minikube_base_tde_test.go
sivanov-nuodb: In this test, we have 1SM and 1TE database and this is why it will take the database down once we restart the SM.
I don't see a point in using multiple SMs in this case because we just want to test SM restart will succeed. If for some reason the password is not set or it's incorrect, the SM will fail to start with an error because it won't be able to decrypt the master password in the archive.

PR #169, opened 2020-12-09
Bump NuoDB Version to 4.0.8

vegichan:


stable/restore/values.yaml
```diff
@@ -32,7 +32,7 @@ nuodb:
   image:
     registry: docker.io
     repository: nuodb/nuodb-ce
```
NikTJ777, line 34: One day - when we have a `client-only` container image, we can use that here - since `nuorestore` only uses `nuocmd`.

test/minikube/minikube_rolling_upgrade_test.go
```diff
@@ -254,7 +254,7 @@ func TestKubernetesUpgradeFullDatabaseMinorVersion(t *testing.T) {
 }
 
 func TestKubernetesRollingUpgradeAdminMinorVersion(t *testing.T) {
-	t.Skip("4.0.7 Admin is not rolling upgradeable from pre-4.0.7")
+	t.Skip("4.0.7+ Admin is not rolling upgradeable from pre-4.0.7")
```
NikTJ777, line 257: Ok, that's subtle.

I had to read it 3 times to find the change... :)

NikTJ777 (approved):
Looks good!

acabrele (approved):

PR #172, opened 2020-12-17
Initial backup job in multicluster deployment fails

sivanov-nuodb:
**Issue**
The initial backup job will fail in multicluster deployments on every cluster other than `cluster0` with:
```
'backup database' failed: No SMs found matching the provided labels backup cluster1
```
This happens because the init container will check if database is in `RUNNING` state which is already the case as there will be engines already deployed in `cluster0`. If the initial backup job fails, then all other incremental and journal backups will fail until the regular full backup is performed on schedule.

**Changes**
- `nuobackup` has been changed so that it can wait for a certain number of SMs with requested labels to become `RUNNING` before performing a backup.
- The init container of the initial job has been replaced by using the new functionality

**Testing**
- Automatic testing is already available in `TestKubernetesRestoreDatabase` test
- Regression testing
- Manual testing using Calico multi-cluster with three k3d clusters

### DIFF
--- a/stable/database/files/nuobackup
+++ b/stable/database/files/nuobackup
@@ -48,59 +48,59 @@ wait_for_num_processes=0
 wait_timeout=300
 
 function await() {
-	local timeout=$1
-	shift
-	local end_time=$(( $(date -u +%s) + $timeout ))
-	local rc=0
-	while [ "$(date -u +%s)" -lt "$end_time" ]; do
-		eval "$*"
-		rc=$?
-		if [ $rc -eq 0 ]; then
-			return $rc
-		fi
-		sleep 1
-	done
-	return 124
+   local timeout=$1
+   shift
+   local end_time=$(( $(date -u +%s) + $timeout ))
+   local rc=0
+   while [ "$(date -u +%s)" -lt "$end_time" ]; do
+      eval "$*"
+      rc=$?
+      if [ $rc -eq 0 ]; then
+         return $rc
+      fi
+      sleep 1
+   done
+   return 124
 }
 
 function check_backup_processes() {
-	declare -a label_patterns=()
+   declare -a label_patterns=()
    local all_process_labels=""
 
-	for ((i=0; i<${#labels_arr[@]}; i+=2)); do
-	  label_patterns[${#label_patterns[@]}]="\"${labels_arr[i]}\"\s*:\s*\"${labels_arr[i+1]}\""
-	done
+   for ((i=0; i<${#labels_arr[@]}; i+=2)); do
+     label_patterns[${#label_patterns[@]}]="\"${labels_arr[i]}\"\s*:\s*\"${labels_arr[i+1]}\""
+   done
 
    all_process_labels="$(nuocmd --api-server $NUOCMD_API_SERVER show database \
       --db-name $db_name \
       --process-format '#{engine_type} {state} {labels}')" || return $?
 
-	num_processes=0
-	while read process_labels; do
+   num_processes=0
+   while read process_labels; do
       [ -z "$process_labels" ] && continue
-		parsed_labels=$(echo "$process_labels" | tr "'" '"' | python -m json.tool 2>&1)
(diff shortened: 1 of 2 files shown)


stable/database/files/nuobackup
```diff
@@ -42,6 +44,64 @@ labels=""
 semaphore=""
 timeout=1800
 backup_root=$BACKUP_DIR
+wait_for_num_processes=0
+wait_timeout=300
```
vegichan, line 48: is there potentially any collision with K8s max job durations?

vegichan (approved):
looks good to me

stable/database/files/nuobackup
sivanov-nuodb: I suppose that you are referring to `.spec.activeDeadlineSeconds` parameter?
We don't set this for backup jobs.

stable/database/files/nuobackup
```diff
@@ -42,6 +44,64 @@ labels=""
...
+
+function check_backup_processes() {
+   declare -a label_patterns=()
+   local all_process_labels=""
+
+   for ((i=0; i<${#labels_arr[@]}; i+=2)); do
+     label_patterns[${#label_patterns[@]}]="\"${labels_arr[i]}\"\s*:\s*\"${labels_arr[i+1]}\""
+   done
+
+   all_process_labels="$(nuocmd --api-server $NUOCMD_API_SERVER show database \
```
adriansuarez, line 74: `NUOCMD_API_SERVER` is used internally by `nuocmd` as the default. Why do we need to specify it explicitly?

stable/database/files/nuobackup
```diff
@@ -42,6 +44,64 @@ labels=""
...
+      --db-name $db_name \
+      --process-format '#{engine_type} {state} {labels}')" || return $?
+
+   num_processes=0
+   while read process_labels; do
```
adriansuarez, line 79: The label patterns are specified via standard input? Also, the labels are printed in the following format:
```
{'label': 'value'[, ...]}
```

Are we expecting users to specify something like `'type': 'backup'`, for example?

stable/database/files/nuobackup
sivanov-nuodb: Good point. I used the same approach as in other `nuocmd` invocations. I see no point in specifying it explicitly unless @NikTJ777 say otherwise.

stable/database/files/nuobackup
sivanov-nuodb: Not sure I understand the question completely. The expected labels are formed from the value of `--group` + values of `--labels` options. Currently `nuobackup --labels ...` is not used anywhere so effectively we are searching for SMs that have the following label `'backup': 'cluster0'`.
There is an explicit label `backup` injected in HC SMs which value is the backup group:
```
        args:
          - "nuosm"
          ...
          - "--labels"
          - "backup {{ include "hotcopy.group" . }} {{- include "opt.key-values" .Values.database.sm.labels }}"
```

adriansuarez: Unless I'm misunderstanding some Bashism, `read process_labels` reads from standard input into the `process_labels` variable. If standard input is not a tty, then this loop is skipped entirely. At least that's what I think the code does from reading it, but I haven't tested it.

Have you done any manual testing that actually causes us to enter the loop? If this is not used anywhere, I would just remove the label-matching functionality.

sivanov-nuodb: Yes, manual testing is done and this is how we await for the processes to become running so that we make the initial backup.

See line 97 --- the `stdin` is replaced with a specific string so that `read`. For example:
```
while read file; do
  echo $file
  # Do something
done <<< "$(ls -1 /tmp)"

adriansuarez: Ah, I didn't even get that far... what a wonderful Bash idiom.

Thanks for the explanation.

adriansuarez (approved):
Looks good.

PR #174, opened 2020-12-30
Backup and restore improvements

sivanov-nuodb:
The primary work was been done by @NikTJ777 and posted in https://github.com/nuodb/nuodb-helm-charts/pull/80.
This PR contains his work and have some extra fixes, improvements, and test automation. The main issues addressed by this change are described below.

**Issue**
Running DB restore to a specific backupset fails if non-HC SM is started first when `autoRestart` is enabled.
As non-HC SMs doesn't have persistent storage for backup, they can only restore/import database from URL. If the restore source is not found or it isn't URL, they should fail so that HC SM that has access to the backupset restore database from it.

**Issue**
Invalid restore source URL or credentials are not handled well and the SM gracefully continues to start even if restore/import has failed. SM ends up SYNCing when `nuosm` should exit with an error.

**Issue**
If `restore.autoRestart` is set to `false`, this doesn't disable database auto restart.

**Changes**
- `nuosm` script now causes the SM startup to exit in the event of fatal errors during restore. Any user errors like invalid restore source or credentials are causing the startup to fail fast signaling to the user that there is an invalid input.
- Logging has been added in `nuosm` and `nuorestore` scripts to better guide the user's expectations.
- Unified code path for `autoRestore`, `autoImport` and `in-place` restore source URL handling.
- Remove any dependence in the restore chart on any values in the database chart.
- Documentation enhancement for restore values in `database` and `restore` charts

**Testing**
- Manual testing for `autoRestore`
- Test automation for `autoImport` and in-place restore using multiple SMs
- Shell script linting integration test
- Restore job integration tests


test/testlib/nuodb_database_utilities.go
```diff
@@ -108,38 +108,46 @@ func StartDatabaseTemplate(t *testing.T, namespaceName string, adminPod string,
 
 	installationStep(t, options, helmChartReleaseName)
 
-	AwaitNrReplicasScheduled(t, namespaceName, tePodNameTemplate, opt.NrTePods)
-	AwaitNrReplicasScheduled(t, namespaceName, smPodName, opt.NrSmPods)
+	if awaitDatabase {
```
sivanov-nuodb, line 111: Sorry for the diff mess. The change done is just to wrap the database await logic in `if awaitDatabase { ... }`.
The change starts here.

test/testlib/nuodb_database_utilities.go
```diff
@@ -108,38 +108,46 @@ func StartDatabaseTemplate(t *testing.T, namespaceName string, adminPod string,
...
-	AwaitPodUp(t, namespaceName, smPodName0, 240*time.Second)
+		smPodName0 := GetPodName(t, namespaceName, smPodName)
+		AddTeardown(TEARDOWN_DATABASE, func() {
+			go GetAppLog(t, namespaceName, GetPodName(t, namespaceName, smPodName), "", &v12.PodLogOptions{Follow: true})
+		})
+		AwaitPodUp(t, namespaceName, smPodName0, 240*time.Second)
 
-	AwaitDatabaseUp(t, namespaceName, adminPod, opt.DbName, opt.NrSmPods+opt.NrTePods)
+		AwaitDatabaseUp(t, namespaceName, adminPod, opt.DbName, opt.NrSmPods+opt.NrTePods)
+	}
```
sivanov-nuodb, line 132: The change ends here.

stable/database/README.md
```diff
@@ -229,6 +229,12 @@ The following tables list the configurable parameters of the `database` chart an
 | `configFilesPath` | Directory path where `configFiles.*` are found | `/etc/nuodb/` |
 | `configFiles.*` | See below. | `{}` |
 | `podAnnotations` | Annotations to pass through to the SM an TE pods | `nil` |
+| `autoImport.*` | Enable and configure the automatic initialising of the initial database state | `disabled` |
```
vegichan, line 232: `initialize` in the US spelling

stable/database/README.md
```diff
@@ -229,6 +229,12 @@ The following tables list the configurable parameters of the `database` chart an
...
+| `autoImport.source` | The source - typically a URL - of the database copy to import | `""` |
+| `autoImport.credentials` | Authentication for the download of `source` in the form `user`:`password` | '""'|
+| `autoImport.stripLevels` | The number of levels to strip off pathnames when unpacking a TAR file of an archive | `1` |
+| `autoImport.type` | Type of content in `source`. One of `stream` -> exact copy of an archive; or `backupset` -> a NuoDB hotcopy backupset | 'backupset' |
+| `autoRestore.*` | Enable and configure the automatic reinitialising of a single archive in a running database - see the options in `autoImport` | `disabled` |
```
vegichan, line 237: s -> z

stable/database/README.md
```diff
@@ -303,7 +309,20 @@ Any file located in `database.configFilesPath` can be replaced; the YAML key cor
...
 | `nuodb.config` | [NuoDB database options][6] | `nil` |
 
-#### database.serviceSuffix
+#### restore.*
+
+The following tables list the configurable parameters of the restore chart and their default values.
+
+| Parameter | Description | Default |
+| ----- | ----------- | ------ |
+| `restore.type` | What type of restore to perform: [ "database" | "archive" ]. A "database" restore restarts the entire database at a previous state. An "archive" restore restores/repairs a SINGLE archive in a RUNNING database. | `"database"` |
```
vegichan, line 318: does the database have to be `RUNNING`? Can you restore a single archive with a DB that is down? I know that it virtually becomes a `database` restore, but will the flag work?

stable/database/values.yaml
```diff
@@ -116,7 +116,7 @@ admin:
 
 
 database:
-  ## Provide a name in place of prometheus-operator for `app:` labels
+  ## Provide a name in place of the chart name for `app:` labels
```
vegichan, line 119: nice!

stable/restore/README.md
```diff
@@ -146,14 +146,15 @@ nuodb:
 
 #### restore.*
 
-The following tables list the configurable parameters of the backup chart and their default values.
+The following tables list the configurable parameters of the restore chart and their default values.
 
 | Parameter | Description | Default |
 | ----- | ----------- | ------ |
+| `restore.type` | What type of restore to perform: [ "database" | "archive" ]. A "database" restore restarts the entire database at a previous state. An "archive" restore restores/repairs a SINGLE archive in a RUNNING database. | `"database"` |
```
vegichan, line 153: same as above

sivanov-nuodb:
> You need to either write your shell scripts using POSIX shell syntax (preferred) or else use `#!/bin/bash` as the first line not `#!/bin/sh` (not preferred).

Agree. Although it's not in the scope of this PR, I've changed the database `readinessprobe` script to be POSIX compliant.
I know that you mentioned it before but please remind me how is the best way to check it. I'm finding this extension useful - https://marketplace.visualstudio.com/items?itemName=timonwong.shellcheck

stable/database/README.md
```diff
@@ -229,6 +229,12 @@ The following tables list the configurable parameters of the `database` chart an
 | `configFilesPath` | Directory path where `configFiles.*` are found | `/etc/nuodb/` |
 | `configFiles.*` | See below. | `{}` |
 | `podAnnotations` | Annotations to pass through to the SM an TE pods | `nil` |
+| `autoImport.*` | Enable and configure the automatic initialising of the initial database state | `disabled` |
+| `autoImport.source` | The source - typically a URL - of the database copy to import | `""` |
+| `autoImport.credentials` | Authentication for the download of `source` in the form `user`:`password` | '""'|
+| `autoImport.stripLevels` | The number of levels to strip off pathnames when unpacking a TAR file of an archive | `1` |
+| `autoImport.type` | Type of content in `source`. One of `stream` -> exact copy of an archive; or `backupset` -> a NuoDB hotcopy backupset | 'backupset' |
```
adriansuarez, line 236: Why the name "stream" instead of "archive"?

stable/database/README.md
```diff
@@ -229,6 +229,12 @@ The following tables list the configurable parameters of the `database` chart an
 | `configFilesPath` | Directory path where `configFiles.*` are found | `/etc/nuodb/` |
 | `configFiles.*` | See below. | `{}` |
 | `podAnnotations` | Annotations to pass through to the SM an TE pods | `nil` |
+| `autoImport.*` | Enable and configure the automatic initialising of the initial database state | `disabled` |
+| `autoImport.source` | The source - typically a URL - of the database copy to import | `""` |
+| `autoImport.credentials` | Authentication for the download of `source` in the form `user`:`password` | '""'|
+| `autoImport.stripLevels` | The number of levels to strip off pathnames when unpacking a TAR file of an archive | `1` |
```
adriansuarez, line 235: This assumes that the entire contents of the TAR file are contained in a path of that length. Otherwise it would be ambiguous which path within the TAR file contains the archive. I think it would be better if we took a pathname within the TAR file that contains the archive. Also, it is possible that the archive and journal are in different directories, but I'm not sure how this parameter would account for that.

madscientist:
> > You need to either write your shell scripts using POSIX shell syntax (preferred) or else use `#!/bin/bash` as the first line not `#!/bin/sh` (not preferred).
> 
> Agree. Although it's not in the scope of this PR, I've changed the database `readinessprobe` script to be POSIX compliant.
> I know that you mentioned it before but please remind me how is the best way to check it.

The simplest way to check is to do development and testing on a Debian or Debian-based system like Ubuntu, which provides dash (a basic POSIX-compliant shell) as /bin/sh.  If you don't want to do that, you can probably install dash on your Red Hat or CentOS system then run the script specifically using that, as `dash myscript`.  Do not change your /bin/sh to point to dash on a Red Hat / CentOS system... that will likely break many many things!!

There may be other ways to check but I'm not sure.

stable/database/files/nuosm
```diff
@@ -15,12 +13,20 @@ NUODB_DEBUG="false"
 
 : ${NUODB_IMPORT_CREDENTIALS:=:}
 
+ERR_ADMIN_UNAVAILABLE=71
+ERR_BACKUPSET_NOT_FOUND=72
+ERR_NO_SPACE=73
+ERR_INCORRECT_BACKUP=74
+ERR_RESTORE_FAILED=75
```
adriansuarez, line 20: Do we use these return values anywhere? I don't want to add anything to the `nuosm` "contract" that isn't really needed to make the migration of this code to `nuodocker` (where it belongs) simpler.

stable/restore/files/nuorestore
```diff
@@ -10,9 +10,47 @@
...
 
+LOGFILE=${NUODB_LOGDIR:=/var/log/nuodb}/nuorestore.log
+
+#=======================================
+# function - log messages
+#
+function log() {
+  if [ -n "$NUODB_DEBUG" ]; then
+    echo "nuorestore $@" | tee -a $LOGFILE
+   #  curl -k -X POST -E /etc/nuodb/keys/nuocmd.pem -H "Content-type: application/json" $ap_protocol://$NUOADMIN_AP0:8888/api/1/diagnostics/log -d "{\"message\": \"nuorestore $@\"}"
```
adriansuarez, line 26: Can we remove the commented out code? Also, we now have `nuocmd log message`, so we would not have to use `curl`.

stable/database/files/nuosm
```diff
@@ -15,12 +13,20 @@ NUODB_DEBUG="false"
...
+
 startup_key="/nuodb/nuosm/startup"
 
-DB_DIR=${NUODB_ARCHIVEDIR}/${NUODB_DOMAIN}/${DB_NAME}
+DB_PARENTDIR=${NUODB_ARCHIVEDIR}/${NUODB_DOMAIN}
+DB_DIR=${DB_PARENTDIR}/${DB_NAME}
```
adriansuarez, line 25: This should check that DB_NAME is defined before we proceed further.

stable/database/files/nuosm
```diff
@@ -33,312 +39,499 @@ if [ $crashcount -ge 1 ]; then
...
   retval=$1
   shift
-  echo "$@"
+  error="$@"
+  [ -n "$wotIdid" ] && error="Error while ${wotIdid}: ${error}"
+  log "$error"
+
+  # cleanup
+  nuocmd set value --key $first_req --value '' --expected-value $HOSTNAME
+  nuocmd set value --key $startup_key/$DB_NAME --value '' --expected-value $HOSTNAME
```
adriansuarez, line 56: `first_req` and `startup_key` are defined later, which is weird but still legal in Bash (what isn't?). We should define those eagerly along with all the other variables above.

stable/database/files/nuosm
```diff
@@ -33,312 +39,499 @@ if [ $crashcount -ge 1 ]; then
...
+
+function isRestoreSourceAvailable() {
+  if isUrl "$restore_source" || [ -d $NUODB_BACKUPDIR/$restore_source ]; then
+    return 0
+  else
+    return 1
+  fi
+}
+
+function releaseLockedArchive() {
```
adriansuarez, line 112: This is just resurrecting the archive object that is still in the domain state as _removed_. I don't understand why this is called `releaseLockedArchive`.

stable/database/files/nuosm
```diff
@@ -33,312 +39,499 @@ if [ $crashcount -ge 1 ]; then
...
+  maxlog=5000000
+  log "logsize=$logsize; maxlog=$maxlog"
+  if [ ${logsize:=0} -gt $maxlog ]; then
+    lines=$(wc -l $LOGFILE)
+    tail -n $(( lines / 2 )) $LOGFILE > ${LOGFILE}-new
+    rm -rf $LOGFILE
+    mv $LOGFILE-new $LOGFILE
+    log "(nuosm) log file wrapped around"
+  fi
+}
```
adriansuarez, line 94: How much logging do we expect this to generate that we are not just emitting to standard output?

stable/database/files/nuosm
```diff
@@ -33,312 +39,499 @@ if [ $crashcount -ge 1 ]; then
...
+
+function isUrl() {
+  if echo $1 | grep -q '^[a-z]\+:/[^ ]\+'; then
```
adriansuarez, line 97: We should quote the argument, i.e. `"$1"`. Also, do we actually need the `if; then; else; fi`? Isn't it the same to just invoke this command? I guess we would get whatever non-0 exit code the command returns instead of 1, but I'm not sure if that matters.

stable/database/files/nuosm
```diff
@@ -33,312 +39,499 @@ if [ $crashcount -ge 1 ]; then
...
+    return 0
+  else
+    return 1
+  fi
+}
+
+function isRestoreSourceAvailable() {
+  if isUrl "$restore_source" || [ -d $NUODB_BACKUPDIR/$restore_source ]; then
```
adriansuarez, line 105: Same comment as above about `if; then; else; fi`. Also, this has to be called in a scope that defines `restore_source`. I think it would be better to make that an explicit argument (i.e. `"$1"`).

stable/database/files/nuosm
```diff
@@ -33,312 +39,499 @@ if [ $crashcount -ge 1 ]; then
...
+    return 0
+  else
+    return 1
+  fi
+}
+
+function releaseLockedArchive() {
+  trace "releasing my locked archive metadata"
+  locked_archive=$( nuocmd show archives --db-name $DB_NAME --removed --removed-archive-format "archive-id: {id}" | sed -En "/^archive-id: / {N; /$HOSTNAME/ s/^archive-id: ([0-9]+).*$/\1/; T; p}" | head -n 1 )
```
adriansuarez, line 114: I'm not familiar with the sed syntax used here, but this seems to assume that there is only one archive on the current host and that it has been started at least once. Can you include a comment describing what this doozy of a one-liner does (maybe @NikTJ777 can explain)?

stable/database/files/nuosm
```diff
@@ -33,312 +39,499 @@ if [ $crashcount -ge 1 ]; then
...
-[ -z "$myArchive" ] && myArchive=-1
-[ -n "$NUODB_DEBUG" ] && echo "myArchive=$myArchive; DB=$DB_NAME; hostname=$HOSTNAME"
-[ -n "$NUODB_DEBUG" -a "$myArchive" = "-1" ] && echo "$($NUOCMD show archives --db-name $DB_NAME)"
+trace "checking archive raft metadata"
+myArchive=$( nuocmd show archives --db-name $DB_NAME --archive-format "archive-id: {id}" | sed -En "/^archive-id: / {N; /$HOSTNAME/ s/^archive-id: ([0-9]+).*$/\1/; T; p}" | head -n 1 )
+[ -z "$myArchive" ] && myArchive=$( nuocmd show archives \
+  --db-name $DB_NAME \
+  --removed --removed-archive-format "archive-id: {id}" | sed -En "/^archive-id: / {N; /$HOSTNAME/ s/^archive-id: ([0-9]+).*$/\1/; T; p}" | head -n 1 )
+[ -z "$myArchive" ] && myArchive="-1"
+log "myArchive=$myArchive; DB=$DB_NAME; hostname=$HOSTNAME"
```
adriansuarez, line 292: Can we log `archiveID` instead of `myArchive`, which reads weirdly.

stable/database/files/nuosm
```diff
@@ -33,312 +39,499 @@ if [ $crashcount -ge 1 ]; then
...
 
-    echo "I am first-in: $first_in == $HOSTNAME"
-
-    # [ -z "$restore_source" ] && die -1 "There is no valid LATEST backup - restore of LATEST failed."
+    log "I am first-in: $first_in == $HOSTNAME"
 
     # take ownership of the SM startup semaphore
-    $NUOCMD set value --key $startup_key/$DB_NAME --value $HOSTNAME --unconditional
+    trace "Take ownership of SM startup semaphore"
+    nuocmd set value --key $startup_key/$DB_NAME --value $HOSTNAME --unconditional
```
adriansuarez, line 414: What is this KV wrangling for? This is probably a question for @NikTJ777.

stable/database/README.md
sivanov-nuodb: The database doesn't have to be running, however, the admin layer should. `nuorestore` just sets a flag in KV store which indicates that restore needs to be performed on the next database restart.

stable/database/README.md
sivanov-nuodb: This was the name originally chosen. I'm not sure about the reason. Effectively `stream` is a `tar.gz` of NuoDB archive which is ready to be downloaded.

stable/database/README.md
sivanov-nuodb: This value of `stripLevels` is transparently passed to `tar` as `--strip-components` option. The typical use case is that the source tarball contains only one archive directory (including the journal).
Currently, an external journal directory is not supported.

stable/database/files/nuosm
sivanov-nuodb: We have no automation to check engine container return codes. Maybe @NikTJ777 had an idea to use them somehow in the future. `nuodocker` doesn't have specific return codes nowadays. Users would mostly rely on the error messages in the container logs to get a sense of the problem.
I'm happy to remove the exit code "contract" from `nuosm` if there are no other comments about this.

stable/database/files/nuosm
sivanov-nuodb: I moved all KV store request keys at the beginning of the script.

stable/database/files/nuosm
sivanov-nuodb: I see the value of having a `nuosm` persistent log which is available between container restarts. All restore/import logic will be executed only once and it's easy to lose the logging from it especially if the user tried to resolve a restore problem on their own. Of course, all this will work if `log.pesistance` is enabled for SMs.

Example restore log:
```
===========================================
logsize=1487; maxlog=5000000
Directory /var/opt/nuodb/archive/nuodb/demo exists
myArchive=0; DB=demo; hostname=sm-database-aqochm-nuodb-cluster0-demo-hotcopy-0
restore_source=20201229T225653; restore_requested=20201229T225653; path=/var/opt/nuodb/archive/nuodb/demo; atoms=78; catalogs=213
First-in = sm-database-aqochm-nuodb-cluster0-demo-hotcopy-0
I am first-in: sm-database-aqochm-nuodb-cluster0-demo-hotcopy-0 == sm-database-aqochm-nuodb-cluster0-demo-hotcopy-0
Deleting archiveId=1
Restoring 20201229T225653; existing archive directores: total 4
drwxr-xr-x 37 nuodb root 4096 Dec 29 22:57 demo
(restore) recreated /var/opt/nuodb/archive/nuodb/demo; atoms=0
Calling nuodocker to restore 20201229T225653 into /var/opt/nuodb/archive/nuodb/demo
Finished restoring /var/opt/nuodb/backup/20201229T225653 to /var/opt/nuodb/archive/nuodb/demo. Created archive with archive ID 2
Deleting my archive metadata: 0
Restored from 20201229T225653
Clearing restore credential request from raft
[2] <NO VALUE> : /var/opt/nuodb/archive/nuodb/demo @ demo [journal_path = ] [snapshot_archive_path = ] NOT_RUNNING
```

Example normal startup:
```
===========================================
logsize=44; maxlog=5000000
Created new dir /var/opt/nuodb/archive/nuodb/demo
myArchive=-1; DB=demo; hostname=sm-database-aqochm-nuodb-cluster0-demo-hotcopy-0
restore_source=; restore_requested=; path=/var/opt/nuodb/archive/nuodb/demo; atoms=0; catalogs=0
```

stable/database/files/nuosm
sivanov-nuodb: I'm fine removing the `if/else` but I prefer to keep the `return $?` so that it's clear what the function returns.

stable/database/files/nuosm
sivanov-nuodb: This is my understanding as well. It expects the `archiveId` on the first line followed by the exited process on the next line. AFAIK this is the case always in docker deployments as the process is started right after the archive is created by `nuodocker start sm`.
I will add a comment.

stable/database/files/nuosm
sivanov-nuodb: As far as I can tell, there are two main KV keys used for synchronization:
- `$NUODB_RESTORE_REQUEST_PREFIX/$DB_NAME/first` - this is held by the SM which will perform restore. Any other SM will wait for the "first" SM to go into `RUNNING` state before it starts. 
- `$startup_key/$DB_NAME` - this is held by every SM when they start so that SMs start in sequential order so that we don't have more than one SYNCing SM at a time. This is probably to address an old issue where we had problems when more than 2 SMs start SYNCing.

stable/database/files/nuosm
sivanov-nuodb: We have a default value for `NUODB_DOMAIN` so I could add one for `DB_NAME`.

adriansuarez (approved):
Looks good and it is good that we have test coverage of this now.

We should definitely move a lot of this functionality into `nuodocker`. The behavior of blocking process startup based on some key in the KV store is useful and something that we could generalize, because it also has other use cases like diagnostics (e.g. someone wants to figure out why an SM keeps crashing and wants access its archive directory, but that is only possible in K8s if we inhibit startup to that it does not crash again).

vegichan (approved):
Thanks for getting this over the line

acabrele:
I only just got round to looking at this since I was on hols. I was mostly interested in whether all the issues found during welab testing were addressed.  I thought I would leave this comment here just to confirm, as far as I can see, the changes do resolve them all.

sivanov-nuodb:
@acabrele Thanks for checking it! The three issues in the PR description should have been fixed by this PR (two of them are found during the initial B&R testing by QA team). If you find any major problems, please open issues so that we can fix them in addition.

acabrele:
> @acabrele Thanks for checking it! The three issues in the PR description should have been fixed by this PR (two of them are found during the initial B&R testing by QA team). If you find any major problems, please open issues so that we can fix them in addition.

It also fixes some other issues not listed :-)

PR #176, opened 2021-01-06
Fix for marking restore as successful

sivanov-nuodb:
**Issue**
There could be a timing hazard during restore when non-first SM will not enter the restore waiting loop because the restore request has already been cleared from Raft state. There is a possibility that such SM starts first which will cause a restore failure.

First-in SM:
```
...
2021-01-05T14:54:21.226+0000 archiveId=0; DB=demo; hostname=sm-database-jc0lbt-nuodb-cluster0-demo-hotcopy-0
2021-01-05T14:54:21.229+0000 trace: retrieving restore request from raft
2021-01-05T14:54:21.906+0000 trace: retrieving restore credentials from raft
2021-01-05T14:54:23.259+0000 restore_source=20210105T145238; restore_requested=20210105T145238; path=/var/opt/nuodb/archive/nuodb/demo; atoms=78; catalogs=213
...
2021-01-05T14:54:24.539+0000 First-in = sm-database-jc0lbt-nuodb-cluster0-demo-hotcopy-0
2021-01-05T14:54:24.542+0000 I am first-in: sm-database-jc0lbt-nuodb-cluster0-demo-hotcopy-0 == sm-database-jc0lbt-nuodb-cluster0-demo-hotcopy-0
2021-01-05T14:54:24.545+0000 trace: Take ownership of SM startup semaphore
2021-01-05T14:54:25.189+0000 trace: resurrecting removed archive metadata
2021-01-05T14:54:25.863+0000 trace: Disable all archive metadata except my own
2021-01-05T14:54:26.522+0000 Deleting archiveId=1
2021-01-05T14:54:27.200+0000 trace: performing restore
...
2021-01-05T14:54:29.215+0000 Restored from 20210105T145238
2021-01-05T14:54:29.226+0000 Clearing restore credential request from raft
2021-01-05T14:54:29.918+0000 trace: clearing restore request
...
2021-01-05T14:54:33.355+0000 trace: executing nuodocker to start SM
```

Other SM:
```
...
2021-01-05T14:54:30.359+0000 archiveId=1; DB=demo; hostname=sm-database-jc0lbt-nuodb-cluster0-demo-0
2021-01-05T14:54:30.361+0000 trace: retrieving restore request from raft
2021-01-05T14:54:31.049+0000 restore_source=; restore_requested=; path=/var/opt/nuodb/archive/nuodb/demo; atoms=78; catalogs=213
2021-01-05T14:54:31.052+0000 trace: resurrecting removed archive metadata
2021-01-05T14:54:31.759+0000 Resurrecting removed archiveId=1
Archive(archive_path=/var/opt/nuodb/archive/nuodb/demo, db_name=demo, id=1, state=NOT_RUNNING)
2021-01-05T14:54:33.137+0000 [1] <NO VALUE> : /var/opt/nuodb/archive/nuodb/demo @ demo [journal_path = ] [snapshot_archive_path = ] NOT_RUNNING
  [SM] sm-database-jc0lbt-nuodb-cluster0-demo-0/172.17.0.6:48006 [start_id = 1] [server_id = admin-jz9byl-nuodb-cluster0-0] [pid = 127] [node_id = 3] EXITED(REQUESTED_SHUTDOWN:SHUTTING_DOWN):(2021-01-05T14:53:26.107+0000) Gracefully shutdown engine (?)
[2] <NO VALUE> : /var/opt/nuodb/archive/nuodb/demo @ demo [journal_path = ] [snapshot_archive_path = ] NOT_RUNNING
2021-01-05T14:54:33.140+0000 trace: executing nuodocker to start SM
```

**Changes**
Fetch the restore request from KV store once the pod is scheduled and clear it right before "first-in" SM is started. 
This will not resolve the timing hazard completely but will make it less likely to happen. The real fix should be to clear the restore request only after the "first-in" SM goes to `RUNNING` state which is hard to be done in `nuosm`.

EDIT: Restore request will be cleared once database goes to `RUNNING` state which means that the first SM has started successfully.

**Testing**
- Regression testing with multiple SMs


stable/database/files/nuosm
```diff
@@ -517,6 +512,12 @@ log "$( nuocmd show archives --db-name $DB_NAME )"
 wotIdid="releasing first-in semaphore"
 [ -n "$first_in" ] && nuocmd set value --key $first_req --value '' --expected-value $HOSTNAME
 
+if [ -n "$restore_requested" ] && [ -n "$restore_successful" ]; then 
+    # clear the restore_request only if restore is successfull
+    trace "clearing restore request"
+    nuocmd set value --key $restore_req --value '' --expected-value "$restore_requested"
+fi
+
```
adriansuarez, line 520: I think this actually makes the timing issue less likely but does not eliminate it. In order to do that we would have to clear that entry in the KV store _after_ the initial SM is RUNNING, and then all other SMs can sync from it. But I don't know how you would do that if `nuodocker start sm` is "taking over".

Another approach would be to leave the entry untouched and add more intelligence to the check that the non-initial SMs perform when rejoining. The non-initial SMs should be able to join if there is already a running SM.

Of course, this would all be much simpler if it was implemented in `nuodocker` or in the Admin itself, but I think this process should be made more manual to eliminate this kind of complexity. Data recovery has to be performed with human intervention anyway, so why is there a bunch of complex logic to automate parts of it?

I think we should write down the steps that a user would follow to perform data recovery, and make sure we can support that. What I would prefer is something like the following,
1. Scale down SM StatefulSet to 1 replica.
2. Set some value in the KV store to pause the (one and only) SM.
3. Perform some explicit restore action on the SM filesystem, e.g. copy files, run `nuoarchive restore`, etc.
4. Unset the value in the KV store to unpause the (one and only) SM.
5. Wait for the SM to become RUNNING and then set the replica count to whatever was desired.

That would make it possible for us to diagnose issues with the data recovery process, which I think would be extremely difficult with all of this automatic logic that we now have.

sivanov-nuodb: Yes, the initial change didn't solve the timing hazard completely as stated in the PR description. I've invested some more time to think about if this is possible using an asynchronous process that will wait for the database to become `RUNNING` before clearing the request in the KV store. Please check the latest changes.
If someone wants to have more control over the restore process, he/she can still scale up one SM at a time (similar to what the `TestKubernetesRestoreMultipleSMs/manualRestart` does).

This is how `in-place-restore` is triggered now:
```
helm install -n nuodb --generate-name stable/restore \
  --set restore.target=demo \
  --set restore.source="20210106T203316"
```

I agree that if we make the process more manual, we might remove the complexity that we currently have, however, this doesn't necessarily mean that the process will be less error-prone. In fact, I think that the most complex part of the `restore-in-place` is understanding and synchronizing the archives state in admin with the archives on disk which can be easily messed up.

I like the idea to also provide a "pause" option along with instructions in case someone wants to do the restore manually, however, I prefer to still support the automatic way. Happy to hear other comments on this.

adriansuarez (approved):
I will defer to you on verifying that this works (which I assume will require a Bamboo run).

vegichan (approved):
looks good!

PR #177, opened 2021-01-06
Make LB job optional

vegichan:
Create new section in admin YAML for `legacy` features. Make LB job legacy, is has been superceded by the KAA module.


acabrele (approved):
I like the idea of a legacy / deprecated section.  It makes it very obvious.

sivanov-nuodb (approved):
Looks good. Thanks @vegichan!

PR #178, opened 2021-01-08
Remove database name check after restore

sivanov-nuodb:
**Changes**
There is a valid use case when a backupset/archive is taken from database with name `A` and is used to bootstrap database with name `B` in the same domain. Currently, it's not possible to use `autoImport` to fulfill the above requirements.
Removed the database name check after the restore is performed.

There shouldn't be a problem with `autoRestore` because if a backup from a different database is used, the SM startup will fail and complain about a mismatch in the database UUIDs.

**Testing**
Regression testing

### DIFF
--- a/stable/database/files/nuosm
+++ b/stable/database/files/nuosm
@@ -237,14 +237,6 @@ function perform_restore() {
     return $retval
   fi
 
-  # make sure the archive i s correct for this database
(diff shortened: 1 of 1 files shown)


adriansuarez (approved):

vegichan (approved):
makes sense

PR #179, opened 2021-01-12
Kubernetes Jobs cause rolling upgrade errors

sivanov-nuodb:
**Changes**
- Remove initial database backup job
- Remove post restore datadase job
- Modified `nuobackup` so that `full` hotcopy is executed as a prerequisite of any `journal` or `incremental` hotcopy
- Remove the number of engines await logic in `nuobackup` as it's not used anymore

After the changes, the initial database backup will be performed along with the first backup on a schedule.
If multiple full backups are requested at the same time, one of them will fail with:
```
Executing full hotcopy as a prerequisite for incremental hotcopy: 'get current-backup' failed: No existing backup path found for archive id 0
'backup database' failed: Failure while performing hot-copy: Error while sending engine request: Only one full hot copy at a time allowed
Error running hotcopy 1
```

**Testing**
- Modifed backup test case
- Regression testing


test/integration/template_resources_test.go
```diff
@@ -204,7 +204,6 @@ func TestPullSecretsRenderAllGlobal(t *testing.T) {
 	helm.RenderTemplate(t, options, "../../stable/database", "release-name", []string{"templates/statefulset.yaml"})
 	helm.RenderTemplate(t, options, "../../stable/database", "release-name", []string{"templates/deployment.yaml"})
 	helm.RenderTemplate(t, options, "../../stable/database", "release-name", []string{"templates/cronjob.yaml"})
-	helm.RenderTemplate(t, options, "../../stable/database", "release-name", []string{"templates/job.yaml"})
```
sivanov-nuodb, line 207: This is the relevant change.

test/integration/template_resources_test.go
```diff
@@ -183,7 +184,6 @@ func TestPullSecretsRenderAllNuoDB(t *testing.T) {
...
```
sivanov-nuodb, line 186: This is the relevant change.

stable/database/files/nuobackup
```diff
@@ -175,82 +172,28 @@ if [ -n "$semaphore" ]; then
...
+      --db-name "${db_name}" \
+      --labels "backup ${backup_group} ${labels}" 2>&1)"
+   if [ $? -ne 0 ] && echo "$error" | grep -q "No existing backup path found"; then
+      # Check if full backup has been started already for all archiveIds
+      # This will most likely execute in the following scenarious:
+      # * initial database backup request
+      # * backup request after the database have been restored
+      # * new HC SM has been started since the last full backup
+      echo "Executing full hotcopy as a prerequisite for ${backup_type} hotcopy: ${error}"
+      nuocmd check database --db-name $db_name --check-running --wait-forever
```
NikTJ777, line 194: This command will hang forever if an SM has been shut down, since the database will be in AWAITING_ARCHIVE_HISTORY_INC state, and _not_ the RUNNING state.

stable/database/files/nuobackup
```diff
@@ -175,82 +172,28 @@ if [ -n "$semaphore" ]; then
...
-      exit -1
+if [ "$backup_type" != "full" ]; then
+   error="$( nuodocker get current-backup \
+      --db-name "${db_name}" \
+      --labels "backup ${backup_group} ${labels}" 2>&1)"
+   if [ $? -ne 0 ] && echo "$error" | grep -q "No existing backup path found"; then
+      # Check if full backup has been started already for all archiveIds
+      # This will most likely execute in the following scenarious:
+      # * initial database backup request
+      # * backup request after the database have been restored
```
NikTJ777, line 191: I see the comment - but I don't yet understand how this works after a database has been restored.
Why does the `nuodocker get current-backup` fail after the database has been restored?

stable/database/files/nuobackup
sivanov-nuodb: Good point! You are correct that `check database` won't succeed if the database state is other than `RUNNING`.

>if any SM has been shut down
KAA will remove the archive for scaled-down replicas, hence, the database will be in `RUNNING` state even if SM has been shutdown.
Even if KAA is disabled, we won't get into the _if block_ as `get current_backup` is checking archives for registered domain processes only.

>or is failing to start
We might get into a situation where the `check database` is stuck if:
* Database is running
* HC SM statefulset is scaled up but the SM fails to start
* The failed SM is left into a crash-looping state.

The only purpose of the check here is to prevent several full backups after initial deployment.
We could just add a timeout of 300sec so that this doesn't block forever.

stable/database/files/nuobackup
sivanov-nuodb: If `get current-backup` is not able to find the backupset name for some `archiveId`, it fails with "No existing backup path found for archive id x".
This can be used is several use cases:
* after restore - The restoration process involves the creation of new archives with new `archiveIds`. 
* after HC SM scale-up - The new `archiveId` will be detected.

NikTJ777: Got it!
THanks!

test/minikube/minikube_rolling_upgrade_test.go
```diff
@@ -196,11 +196,6 @@ func TestKubernetesUpgradeFullDatabaseMinorVersion(t *testing.T) {
 
 	databaseHelmChartReleaseName := testlib.StartDatabase(t, namespaceName, admin0, &databaseOptions)
 
-	// all jobs need to be deleted before an upgrade can be performed
```
NikTJ777, line 199: Woot!
All the other work just to delete these lines :)
Great Job! :)

test/minikube/minikube_upgrade_helm_test.go
```diff
@@ -138,8 +139,12 @@ func upgradeDatabaseTest(t *testing.T, fromHelmVersion string, updateOptions *Up
 	// all jobs need to be deleted before an upgrade can be performed
 	// so far we have not found an automated way to delete them as part of a pre-upgrade hook
 	// if we find it, this line can be removed and the test should still pass
-	testlib.DeletePod(t, namespaceName, "jobs/job-lb-policy-nearest")
```
NikTJ777, line 141: I think removing this reliance on knowing all the jobs will make the tests more robust going forward.

test/minikube/minikube_base_database_test.go
```diff
@@ -207,12 +207,7 @@ func verifyEngineAltAddress(t *testing.T, namespaceName string, admin0 string, e
 	}
 }
 
-func backupDatabase(t *testing.T, namespaceName string, podName string, databaseName string, options *helm.Options) {
-
-	// wait for the backup to both start _and_ complete successfully
-	backupJob := fmt.Sprintf("hotcopy-%s-job-initial", databaseName)
-	testlib.AwaitPodPhase(t, namespaceName, backupJob, corev1.PodSucceeded, 120*time.Second)
-
+func verifyBackup(t *testing.T, namespaceName string, podName string, databaseName string, options *helm.Options) {
```
NikTJ777, line 210: Nice simplification :)

test/testlib/minikube_utilities.go
```diff
@@ -253,7 +254,11 @@ func arePodConditionsMet(pod *corev1.Pod, condition corev1.PodConditionType,
 func findAllPodsInSchema(t *testing.T, namespace string) []corev1.Pod {
 	options := k8s.NewKubectlOptions("", "", namespace)
 	filter := metav1.ListOptions{}
-	return k8s.ListPods(t, options, filter)
+	pods := k8s.ListPods(t, options, filter)
+	sort.SliceStable(pods, func(i, j int) bool {
```
NikTJ777, line 258: Nice :)

NikTJ777 (approved):
Awesome!

I am really impressed with the simplifications you've achieved.

vegichan (approved):
looks good!

vegichan:
@sivanov-nuodb  can you help me out with the RN for this one?

PR #180, opened 2021-01-20
Add evicted-servers to admin statefulset

vegichan:
Simplify recovery from a disaster that leaves the admin domain in a minority.

Usually, this is handled by KAA, but we allow this extension for multi-cluster installs and for environments that do not have KAA enabled.

For more info on eviction, read:
https://doc.nuodb.com/nuodb/latest/database-administration/recovering-from-a-lost-majority/

### DIFF
--- a/stable/admin/README.md
+++ b/stable/admin/README.md
@@ -186,7 +186,7 @@ The following tables list the configurable parameters for the `admin` option of 
 | `podAnnotations` | Annotations to pass through to the Admin pod | `nil` |
 | `tde.secrets` | Transparent Data Encryption secret names used for different databases | `{}` |
 | `tde.storagePasswordsDir` | Transparent Data Encryption storage passwords mount path | `/etc/nuodb/tde` |
-| `evicted.servers` | A list of evicted servers excluded RAFT consensus. Used during disaster recovery. | `[]` |
+| `evicted.servers` | A list of evicted servers excluded from RAFT consensus. Used during disaster recovery. | `[]` |
 
 
 For example, when using GlusterFS storage class, you would supply the following parameter:
--- a/stable/admin/README.md
+++ b/stable/admin/README.md
@@ -186,6 +186,8 @@ The following tables list the configurable parameters for the `admin` option of 
 | `podAnnotations` | Annotations to pass through to the Admin pod | `nil` |
 | `tde.secrets` | Transparent Data Encryption secret names used for different databases | `{}` |
 | `tde.storagePasswordsDir` | Transparent Data Encryption storage passwords mount path | `/etc/nuodb/tde` |
+| `evicted.servers` | A list of evicted servers excluded RAFT consensus. Used during disaster recovery. | `[]` |
+
 
 For example, when using GlusterFS storage class, you would supply the following parameter:
 
--- a/stable/admin/templates/_helpers.tpl
+++ b/stable/admin/templates/_helpers.tpl
@@ -1,4 +1,12 @@
 {{/*
+Generic helper function to turn a list into a comma separated string
+*/}}
+{{- define "helm-toolkit.utils.joinListWithComma" -}}
+{{- $local := dict "first" true -}}
+{{- range $k, $v := . -}}{{- if not $local.first -}},{{- end -}}{{- $v -}}{{- $_ := set $local "first" false -}}{{- end -}}
+{{- end -}}
+
+{{/*
 Expand the name of the chart.
 */}}
 {{- define "admin.name" -}}
--- a/stable/admin/templates/statefulset.yaml
+++ b/stable/admin/templates/statefulset.yaml
@@ -112,9 +112,11 @@ spec:
       {{- end }}
         args:
           - "nuoadmin"
+          {{- if .Values.admin.evicted}}
           {{- if .Values.admin.evicted.servers}}
           - "--evicted-servers"
           - "{{ include "helm-toolkit.utils.joinListWithComma" .Values.admin.evicted.servers }}"
+          {{- end }}
           {{- end }}
           - "--"
           {{- if .Values.admin.tde }}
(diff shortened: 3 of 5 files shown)


test/integration/template_admin_test.go
```diff
@@ -447,3 +447,22 @@ func TestLoadBalancerLegacyRenders(t *testing.T) {
 	}
 }
 
+func TestAdminEvictedServers(t *testing.T) {
+	// Path to the helm chart we will test
+	helmChartPath := testlib.ADMIN_HELM_CHART_PATH
+
+	options := &helm.Options{
+		SetValues: map[string]string{
+			"admin.evicted.servers":   "{nuoadmin-1,nuoadmin-2}",
```
adriansuarez, line 456: I'm guessing the string `"{nuoadmin-1,nuoadmin-2}"` expand to a list. Is it possible to check that `nuoadmin-1` and `nuoadmin-2` are expanded as distinct entries in that list, rather than this expanding as a single string with a comma embedded inside of it?

adriansuarez (approved):
Looks good. I just have the one comment in the test, since I'm not sure if this actually exercises the list to CSV conversion.

test/integration/template_admin_test.go
vegichan: that is the limitation of the test. I don't like committing YAML files to git and using them in tests. The `{...,...} form is how you specify a list via the command line.

This is how a customer would specify it in real life:
```
$cat file.yaml 
admin:
  evicted:
    servers:
      - nuoadmin1
      - nuoadmin2
```

```
$ helm template --values ./file.yaml --show-only  templates/statefulset.yaml release-name stable/admin/
---
# Source: admin/templates/statefulset.yaml
apiVersion: "apps/v1"
kind: StatefulSet
...
        args:
          - "nuoadmin"
          - "--evicted-servers"
          - "nuoadmin1,nuoadmin2"
          - "--"
          - "pendingReconnectTimeout=60000"
          - "processLivenessCheckSec=30"
```

adriansuarez: What happens if you change the string to `"{ 'nuoadmin-1', 'nuoadmin-2' }"`? I'm not sure if that is valid Helm syntanx, but if that works it would be more convincing that we are actually combining distinct elements in a list to obtain `nuoadmin1,nuoadmin-2`.

sivanov-nuodb: I think that the syntax used by Martin is the correct one to specify a list.
There is also an alternative syntax:
```
"admin.evicted.servers[0]": "nuoadmin1",
"admin.evicted.servers[1]": "nuoadmin2",
```

stable/admin/README.md
```diff
@@ -186,6 +186,8 @@ The following tables list the configurable parameters for the `admin` option of
 | `podAnnotations` | Annotations to pass through to the Admin pod | `nil` |
 | `tde.secrets` | Transparent Data Encryption secret names used for different databases | `{}` |
 | `tde.storagePasswordsDir` | Transparent Data Encryption storage passwords mount path | `/etc/nuodb/tde` |
+| `evicted.servers` | A list of evicted servers excluded from RAFT consensus. Used during disaster recovery. | `[]` |
```
sivanov-nuodb, line 189: Why not `admin.evictedServers`? Do we plan to add more keys under `admin.evicted` variable?

stable/admin/templates/statefulset.yaml
```diff
@@ -112,6 +112,10 @@ spec:
       {{- end }}
         args:
           - "nuoadmin"
+          {{- if .Values.admin.evicted.servers}}
```
sivanov-nuodb, line 115: If you decide to keep the variable nesting, then we might want to also check that `.Values.admin.evicted` is not `nil`.

acabrele:
I initially though we might make this more generic like options - what else might we want to specify, for example use something like 
evictedServers: nuoadmin1,nuoadmin2 so that both the arg and the value are supplied.  But, I'm not sure what else we would ever want to supply to nuoadmin like this!

stable/admin/templates/statefulset.yaml
acabrele: I can see a top level evicted _might_ be useful when we come to looking at failover automation, but it's too early to say given there is no solution yet.

acabrele (approved):
Comments only

stable/admin/templates/statefulset.yaml
vegichan: well spotted. Fixed

stable/admin/README.md
adriansuarez: +1

stable/admin/templates/_helpers.tpl
```diff
@@ -1,3 +1,11 @@
+{{/*
+Generic helper function to turn a list into a comma separated string
```
NikTJ777, line 2: I believe this reimplements the helm `join` function:
So you could remove this, and change the statefulset template to simply use:
`{{ .Values.admin.evict.servers | join "," }}`

See http://masterminds.github.io/sprig/ for the list of functions that are included in helm.
(Somewhere there is a list of the Sprig functions that Helm does _not_ include - but I can't find it off-hand...)

stable/admin/templates/statefulset.yaml
```diff
@@ -112,6 +112,12 @@ spec:
       {{- end }}
         args:
           - "nuoadmin"
+          {{- if .Values.admin.evicted}}
+          {{- if .Values.admin.evicted.servers}}
```
NikTJ777, line 116: As a stylistic choice, I prefer using {{- with ... }} in place of {{- if ... }}  - and particularly a nested {{- if }} as the {{with}} avoids the need to dereference the desired value twice.

So something like:
{{- if Values.admin.evicted }}
{{- with Values.admin.evicted.servers }}
--evicted-servers
{{ join . "," }}
{{- end}}
{{- end}}

I think it's more clear that the {{- if}} is a guard against an empty structure, and the {{- with}} encloses some conditional rendering.

But a minor point...

NikTJ777 (approved):
Awesome!

Sometimes it is the small elegant ones that are most satisfying :)

PR #182, opened 2021-02-03
Incremental hotcopy has a prerequisite that full exist in the backupset

sivanov-nuodb:
**Issue**
Full hotcopy is requested as prerequisite of journal and incremental hotcopies. If a full hotcopy fails for some reason, all subsequent incremental hotcopies using this backupset will fail as they need a full backup element in the backupset. This will require the user to check what is the reason for the full hotcopy failure, take corrective actions, and run *manually* full hotcopy again. Some of the reasons could be a temporary problem like HC SM or database restart, others could be non-transient like the backup volume space issue. 

**Changes**
`nuobackup` now marks backupsets for which the full hotcopy has failed. If an incremental hotcopy is requested in backupset with a failed full, another full hotcopy is executed as a prerequisite.

### DIFF
--- a/stable/database/files/nuobackup
+++ b/stable/database/files/nuobackup
@@ -43,6 +43,12 @@ semaphore=""
 timeout=1800
 backup_root=$BACKUP_DIR
 
+function get_current_backup() {
+   output=$( nuodocker get current-backup \
+      --db-name "${db_name}" \
+      --labels "${labels}" 2>/dev/null)
+   echo "$output" | head -1 | awk '{print $2}'
+}
 
 function execute_backup() {
    local type="$1"
@@ -51,6 +57,7 @@ function execute_backup() {
    local next
    local backupset
    local output
+   local failed
 
    # call nuodocker to perform the actual backup
    echo "Starting ${type} backup for database ${db_name} on processes with labels '${labels}' ..."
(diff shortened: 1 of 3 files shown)


stable/database/files/nuobackup
```diff
@@ -64,19 +71,32 @@ function execute_backup() {
...
+      # Store failed backupset in KV store
+      if [ "$type" = "full" ] && [ -n "$backup_group" ]; then
+         backupset=$(get_current_backup)
+         if [ -n "$backupset" ]; then
+            failed=$(nuocmd get value --key "$NUODB_BACKUP_KEY/$db_name/${backup_group}/failed")
+            nuocmd set value \
+               --key "$NUODB_BACKUP_KEY/$db_name/${backup_group}/failed" \
+               --value "${failed} ${backupset}" --expected-value "${failed}"
+         fi
+      fi
```
acabrele, line 83: Could the reason for backup failure also cause the key value command to fail.  I guess then you are no worse a position than without these changes.

stable/database/files/nuobackup
```diff
@@ -184,7 +204,7 @@ fi
 if [ "$backup_type" != "full" ]; then
    error="$( nuodocker get current-backup \
       --db-name "${db_name}" \
-      --labels "backup ${backup_group} ${labels}" 2>&1)"
+      --labels "${labels}" 2>&1)"
```
acabrele, line 207: I think this is correct not to restrict to the backup group, but why was it there before?

acabrele (approved):

stable/database/files/nuobackup
sivanov-nuodb: The backup group (if exist) is already included in the labels on line 164 in the script. It was incorrectly included again here. We should use here the same labels that we are using to issue the hotcopy because they define the `archiveIds` that will be selected for the hotcopy operation.

stable/database/files/nuobackup
sivanov-nuodb: Correct, for this to work, we need a cooperative admin layer.

NikTJ777 (approved):
Nice!
Really elegant solution to failed full backups.

vegichan:
@sivanov-nuodb would you please 
- add labels to this? I think that it is a bug
- update the tile of the PR to a descriptive Release Note. I can help with this if you need my help.

PR #183, opened 2021-02-04
Fix outdated commands in database chart documentation

vegichan:


eresende-nuodb (approved):
Looks good. Didn't spot the case sensitive issue with `hotcopy` vs `HotCopy`. Nice catch.

PR #184, opened 2021-02-10
Database in-place restore support for NuoDB 4.2+

sivanov-nuodb:
**New Features**
- Restore features parity with Helm Charts v3.1.0
- Fine-graned selection of archive ids that need to be restored. Users can now use process labels or specify the archive ids that need to be selected for restore. By default, all HC SMs will be restored when requesting database in-place restore
- Database in-place restore with storage group support.
- Manual restore option for complex database restore requirements

**Changes**
- Restore chart has been updated so that it supports `nuodocker request restore`
- Added support for new `nuodocker` commands in `nuosm`. The implementation is made so that the "legacy" code is not changed but a new code path is followed in case we have a newer enough product version. 

**Testing**
- Updated some of the restore tests
- Added restore using storage groups test

**Upgrade notes**
NuoDB Helm Charts will continue to support the "legacy" way of doing in-place restore for backwards compatibility with older product versions. To use the new restore code, both `database` and `restore` charts should use NuoDB 4.2+.

**Not part of this PR**
- General storage groups support in NuoDB Helm Charts. They still need to be configured manually after database deployment.
- Moving the restore logic from `nuosm` to init container


stable/database/files/nuosm
```diff
@@ -12,8 +12,6 @@
 : ${DB_NAME:="demo"}
 : ${NUODB_SEQUENCE_SYNC:=true}
 
-: ${NUODB_IMPORT_CREDENTIALS:=:}
```
sivanov-nuodb, line 15: GitHub diff viewer is not doing a good job with this file changes. You might want to review this file using "split" more which is side-by-side is a bit easier.
I will also put markers on which part should be reviewed.

stable/database/files/nuosm
```diff
@@ -246,19 +260,134 @@ function perform_restore() {
     return $retval
   fi
 
-  # completely delete my previous archive metadata
-  # this is required because nuodocker always creates a new archive
-  trace "completely deleting the raft metadata for my archive $myArchive"
-  if [ "$myArchive" -ne "-1" ]; then
-    log "Purging my archive metadata archiveId=$myArchive"
-    log "$( nuocmd delete archive --archive-id $myArchive --purge 2>&1 )"
+  if isRestoreRequestSupported \
```
sivanov-nuodb, line 263: The code that should be reviewed starts here.

stable/database/files/nuosm
```diff
@@ -276,235 +405,312 @@ else
...
+      restore_credentials_encoded="$(printf "%s" "${DATABASE_IMPORT_CREDENTIALS:-:}" | base64)"
+      restore_type=${NUODB_AUTO_IMPORT_TYPE}
+      strip_levels=${NUODB_IMPORT_STRIP_LEVELS:-1}
+
+      log "Automatic archive import will be performed for archiveId=${myArchive}, source=${restore_source}, type=${restore_type}, strip=${strip_levels}"
+      trace "restoring empty archive"
+      perform_restore || die $? "$error"
+    fi
+  fi
+else
```
sivanov-nuodb, line 500: The code that should be reviewed ends here.
The rest of the code is just indented with minor changes.

stable/restore/files/nuorestore
```diff
@@ -9,6 +9,9 @@
 #   --semaphore - a key in the raftlog KV store to set to trigger a full backup after the restore
 #   --source    - source to restore from - can be a backupset or a URL
 #   --auto      - automatically start the restore operation ( true | false )
+#   --labels         - additional arbitrary labels to select the SMs for restore
```
sivanov-nuodb, line 12: It's easier to review this file in "split" mode or side-by-side.

stable/restore/README.md
```diff
@@ -152,9 +152,12 @@ The following tables list the configurable parameters of the restore chart and t
 | ----- | ----------- | ------ |
 | `restore.type` | What type of restore to perform: [ "database" | "archive" ]. A "database" restore restarts the entire database at a previous state. An "archive" restore restores/repairs a SINGLE archive in a RUNNING database. | `"database"` |
 | `restore.target` | Where to restore `TO` | `{{ .Values.database.name }}` |
-| `restore.source` | Where to restore `FROM` [ backupset | url | `:latest` ] | `:latest` |
+| `restore.source` | Where to restore `FROM` [ backupset | url | `:latest` | `:group-latest` ] | `:latest` |
```
vegichan, line 155: this does not seem to render correctly

stable/restore/README.md
```diff
@@ -152,9 +152,12 @@ The following tables list the configurable parameters of the restore chart and t
...
 | `restore.credentials` | Credentials to use for a URL source (user:password) | `""` |
 | `restore.autoRestart` | Whether to automatically restart the database and trigger the restore (true/false). Only valid for a "database" restore | `true` |
+| `restore.labels` | Process labels used to filter the complete set of archiveIds that should be restored which then defines the new state of the database upon restore. The setting works with NuoDB 4.2+ | `backup {{ .Values.cloud.cluster.name }}` |
```
vegichan, line 158: I would not use the helm template here. I would use the default rendered value.

stable/restore/files/nuorestore
```diff
@@ -51,6 +54,18 @@ function wrapLogfile() {
   fi
 }
 
+# function - NuoDB 4.2.2+ supports new way of requesting database in-place
```
vegichan, line 57: 4.2.2+?

test/minikube/minikube_base_database_test.go
```diff
@@ -729,7 +795,101 @@ func TestKubernetesRestoreMultipleSMs(t *testing.T) {
 		require.NoError(t, err, "error running SQL: show schema User")
 		require.True(t, strings.Contains(tables, "No tables found in schema "), "Show schema returned: ", tables)
 		testlib.CheckArchives(t, namespaceName, admin0, opt.DbName, 2, 0)
+		checkRestoreRequests(t, namespaceName, admin0, opt.DbName)
+	})
+}
+
+func TestKubernetesRestoreWithStorageGroups(t *testing.T) {
```
vegichan, line 802: how long does the full "upgrade suite" take at this point? Currently CircleCi looks like this:
- short 40min
- long 40min
- diagnostics 20 min
- upgrade 40 min

Once we enable these, will the short/basic suite bloat to 1h+?

In parallel, could I ask you to move the B&R tests from the "base database tests" to their own file, please?

test/minikube/minikube_base_database_test.go
```diff
@@ -729,7 +795,101 @@ func TestKubernetesRestoreMultipleSMs(t *testing.T) {
...
+	databaseOptions := helm.Options{
+		SetValues: map[string]string{
+			"database.name":                         "demo",
+			"database.sm.resources.requests.cpu":    testlib.MINIMAL_VIABLE_ENGINE_CPU,
+			"database.sm.resources.requests.memory": testlib.MINIMAL_VIABLE_ENGINE_MEMORY,
+			"database.te.resources.requests.cpu":    testlib.MINIMAL_VIABLE_ENGINE_CPU,
+			"database.te.resources.requests.memory": testlib.MINIMAL_VIABLE_ENGINE_MEMORY,
+			"database.sm.hotCopy.replicas":          "2",
+			"database.te.logPersistence.enabled":    "true",
+			"database.env[0].name":                  "NUODB_DEBUG",
```
vegichan, line 823: what does this do?

stable/restore/README.md
sivanov-nuodb: Good catch!

stable/restore/README.md
sivanov-nuodb: I will change it.

stable/restore/files/nuorestore
sivanov-nuodb: oops ... I must have missed that.

test/minikube/minikube_base_database_test.go
sivanov-nuodb: `minikube.TestKubernetesRestoreDatabase/restoreDatabaseBackwardsCompatibility` adds 74.93s
`minikube.TestKubernetesRestoreWithStorageGroups ` adds 171.83s (measured in bamboo)

As they are now more test coverage on the B&R, I will move the basic tests in `minikube_base_restore_test.go` marked with +short and the other in `minikube_long_restore_test.go`.

test/minikube/minikube_base_database_test.go
sivanov-nuodb: If set this environment variable enables debug logging in `nuosm` and `nuodocker`.

test/minikube/minikube_base_database_test.go
```diff
@@ -396,419 +340,3 @@ func TestKubernetesAltAddress(t *testing.T) {
 		t.Run("verifyEnginesAltAddress", func(t *testing.T) { verifyEngineAltAddress(t, namespaceName, admin0, expectedNrEngines) })
 	})
 }
-
-func TestKubernetesBackupDatabase(t *testing.T) {
```
sivanov-nuodb, line 400: Moved into database restore specific file.

test/minikube/minikube_base_database_test.go
```diff
@@ -396,419 +340,3 @@ func TestKubernetesAltAddress(t *testing.T) {
...
-
-		populateCreateDBData(t, namespaceName, admin0)
-
-		defer testlib.Teardown(testlib.TEARDOWN_BACKUP)
-		testlib.AwaitJobSucceeded(t, namespaceName, "incremental-hotcopy-demo-cronjob", 120*time.Second)
-		verifyBackup(t, namespaceName, admin0, "demo", &databaseOptions)
-	})
-}
-
-func TestKubernetesRestoreDatabase(t *testing.T) {
```
sivanov-nuodb, line 480: Moved into database restore specific file.

test/minikube/minikube_base_database_test.go
```diff
@@ -396,419 +340,3 @@ func TestKubernetesAltAddress(t *testing.T) {
...
-	defer testlib.Teardown(testlib.TEARDOWN_RESTORE)
-	testlib.RestoreDatabase(t, namespaceName, admin0, &databaseOptions)
-
-	// verify that the database does NOT contain the data from AFTER the backup
-	tables, err := testlib.RunSQL(t, namespaceName, admin0, "demo", "show schema User")
-	require.NoError(t, err, "error running SQL: show schema User")
-	require.True(t, strings.Contains(tables, "No tables found in schema "), "Show schema returned: ", tables)
-}
-
-func TestKubernetesImportDatabase(t *testing.T) {
```
sivanov-nuodb, line 543: Moved into database restore specific file.

test/minikube/minikube_base_database_test.go
```diff
@@ -396,419 +340,3 @@ func TestKubernetesAltAddress(t *testing.T) {
...
-		)
-
-		// verify that the database contains the restored data
-		tables, err := testlib.RunSQL(t, namespaceName, admin0, "demo", "show schema User")
-		require.NoError(t, err, "error running SQL: show schema User")
-		require.True(t, strings.Contains(tables, "HOCKEY"))
-	})
-}
-
-func TestKubernetesRestoreMultipleSMs(t *testing.T) {
```
sivanov-nuodb, line 627: Moved into database restore specific file marked as `long`.

test/minikube/minikube_base_database_test.go
```diff
@@ -396,419 +340,3 @@ func TestKubernetesAltAddress(t *testing.T) {
...
-
-		// verify that the database does NOT contain the data from AFTER the backup
-		tables, err := testlib.RunSQL(t, namespaceName, admin0, opt.DbName, "show schema User")
-		require.NoError(t, err, "error running SQL: show schema User")
-		require.True(t, strings.Contains(tables, "No tables found in schema "), "Show schema returned: ", tables)
-		testlib.CheckArchives(t, namespaceName, admin0, opt.DbName, 2, 0)
-	})
-}
-
-func TestKubernetesAutoRestore(t *testing.T) {
```
sivanov-nuodb, line 735: Moved into database restore specific file.

test/testlib/minikube_utilities.go
```diff
@@ -836,6 +836,12 @@ func GetAdminEventLog(t *testing.T, namespace string, podName string) {
 	)
 }
 
+func AwaitPodLog(t *testing.T, namespaceName string, podName string, fileNameSuffix string) {
```
vegichan, line 839: this function does not block until the log is collected. It's a goroutine. Is that what you intended?

vegichan (approved):

stable/restore/README.md
```diff
@@ -150,11 +150,14 @@ The following tables list the configurable parameters of the restore chart and t
 
 | Parameter | Description | Default |
 | ----- | ----------- | ------ |
-| `restore.type` | What type of restore to perform: [ "database" | "archive" ]. A "database" restore restarts the entire database at a previous state. An "archive" restore restores/repairs a SINGLE archive in a RUNNING database. | `"database"` |
-| `restore.target` | Where to restore `TO` | `{{ .Values.database.name }}` |
-| `restore.source` | Where to restore `FROM` [ backupset | url | `:latest` ] | `:latest` |
+| `restore.type` | What type of restore to perform: [ "database" \| "archive" ]. A "database" restore restarts the entire database at a previous state. An "archive" restore restores/repairs a SINGLE archive in a RUNNING database. | `"database"` |
+| `restore.target` | Where to restore `TO` | `"demo"` |
+| `restore.source` | Where to restore `FROM` [ _backupset_ \| _url_ \| `:latest` \| `:group-latest` ] | `:latest` |
```
adriansuarez, line 155: What do `TO` and `FROM` mean here, and why are they formatting this way (normally fixed width is used to denote code, but think it is just used for emphasis here)? The description doesn't actually say what these values should be, though I can guess that `restore.target` is a database name and `restore.source` is a backup set or URL. I'm not sure how we would differentiate between the two; perhaps the protocol/scheme is used to determine that it is a URL, otherwise it is interpreted as a backup set. We should document that.

stable/restore/README.md
```diff
@@ -150,11 +150,14 @@ The following tables list the configurable parameters of the restore chart and t
...
 | `restore.credentials` | Credentials to use for a URL source (user:password) | `""` |
 | `restore.autoRestart` | Whether to automatically restart the database and trigger the restore (true/false). Only valid for a "database" restore | `true` |
+| `restore.labels` | Process labels used to filter the complete set of archiveIds that should be restored which then defines the new state of the database upon restore. The setting works with NuoDB 4.2+ | `backup cluster0` |
```
adriansuarez, line 158: We should say that this value is interpreted as `<key> <value> [<key> <value> ...]`. I'm not sure what `backup cluster0` means. I would expect labels to look like `type backup`. Also, perhaps we should say how multiple key-value pairs interpreted, i.e. _all_ labels have to match as opposed to _any_.

stable/restore/README.md
```diff
@@ -225,6 +232,24 @@ Verify the restore completed successfully; view the log output from the restarte
 Finished restoring /var/opt/nuodb/backup/20190619T101450 to /var/opt/nuodb/archive/nuodb/demo. Created archive with archive ID 8
 ```
 
+### Manual archive restore
+
+For complex requirements where automatic archive restore is not sufficient, `restore.manual="true"` can be set during restore chart installation.
+The NuoDB SM engine startup will block waiting for the user to perform the archives restore manually.
```
adriansuarez, line 238: TEs will be blocked also.

test/testlib/minikube_utilities.go
sivanov-nuodb: The main purpose of the function is to wait until pod is running and then to request its log. The actual log collection is non-blocking. If you request the log before that, it will report that either the pod or container is not running and will skip the log collection.

stable/restore/README.md
sivanov-nuodb: Good point. I'll edit the description as suggested.

stable/restore/README.md
sivanov-nuodb: The label syntax in Helm is a bit different. It expects a dictionary of labels, like:
```
restore:
  labels:
    label1: value1
    label2: value2
...
```
I will update the default value as it implies that `<key> <value> [<key> <value>]` is used. 
The default value of `{ "backup": "cluster0" }` will select the archive ids of all HC SMs in `cluster0` to be restored. The `backup` label is used historically to define the so-called "backup group" (a group of HC SMs that are backed-up together using `nuodocker backup database`).

stable/restore/README.md
sivanov-nuodb: If `database` restore type is selected, then it will block all engines startup including TEs. If the restore type is `archive`, it's expected that the database is running, and restore for a specific archive id is needed (this is mainly to support the accelerated sync use case).
I will remove the SM reference here.

stable/restore/files/nuorestore
```diff
@@ -86,49 +101,105 @@ do
...
 
-log "Setting flag for database $db_name to restore from $restore_source on next database/SM startup"
+   if [ "$auto" = "true" ]; then
+      if [ "$restore_type" = "database" ]; then
+         log "restore.autoRestart=true - initiating full database restart for database $db_name"
+         nuocmd shutdown database --db-name "$db_name"
+      else
+         for archive_id in $(nuodocker get restore-requests --db-name "$db_name" | grep '[0-9]\+ automatic' | awk '{print $1}'); do
+            start_id=$(nuocmd show domain \
+               --process-format "###{archive_id::-1} {start_id}" | grep "###${archive_id}" | awk '{print $2}')
```
adriansuarez, line 155: Can you grep on `###$start_id ` (note the space)? Otherwise we can match an archive ID that is a prefix of `$start_id`, e.g. we want the running SM on `archive_id=1`, but this matches the one on `archive_id=10`.

stable/restore/files/nuorestore
```diff
@@ -86,49 +101,105 @@ do
...
+            if [ -n "$start_id" ]; then
+               log "restore.autoRestart=true - initiating process startId=${start_id} restart"
+               nuocmd shutdown process --start-id "$start_id"
```
adriansuarez, line 158: `nuocmd shutdown process ... --archive-id ...` would be useful.

stable/restore/files/nuorestore
sivanov-nuodb: Good point! Thanks.

stable/restore/files/nuorestore
sivanov-nuodb: Indeed. It would have saved the two steps process here.

stable/restore/values.yaml
```diff
@@ -81,12 +81,16 @@ restore:
   # name of the existing database being restored 
   target: "demo"
 
-  # source
-  # Can be one of:
+  # The source can be one of:
   # * the name of a backupset on a hotcopy-SM's backup disk;
   # * the metaname ':latest'
   # * the URL of a downloadable TAR file of a hotcopy backupset
   # * stream:URL of a downloadable TAR file of a exact copy of a NuoDB archive
```
acabrele, line 88: This comment about "stream" is here but not in the readme.  Also do we have to mention anything abou the strip levels we used to supply when downloading a tar?

stable/restore/README.md
```diff
@@ -150,11 +150,14 @@ The following tables list the configurable parameters of the restore chart and t
...
+| `restore.target` | The database name to request restore operations for | `"demo"` |
+| `restore.source` | The source which will be restored from. Supported values are [ _backupset_ \| _URL_ \| `:latest` \| `:group-latest` ]. URL restore source is expected to be in form of `protocol://authority/path`. Otherwise the value is interpreted as backupset name | `:latest` |
 | `restore.credentials` | Credentials to use for a URL source (user:password) | `""` |
 | `restore.autoRestart` | Whether to automatically restart the database and trigger the restore (true/false). Only valid for a "database" restore | `true` |
+| `restore.labels` | Process labels used to filter the complete set of archiveIds that should be restored which then defines the new state of the database upon restore. If multiple labels are defined, _any_ of them have to match so that the archive id is selected for restore. The setting works with NuoDB 4.2+ | `{ "backup": "cluster0" }` |
```
acabrele, line 158: Should we add a comment that a multi cluster MUST be modified away from the default in order to effect a full restore (e.g cluster1 must **also** be specified)?

stable/restore/values.yaml
sivanov-nuodb: I will update the README that the URL should point either to backupset or an exact copy of NuoDB archive (_stream_). 
Currently, we don't expose strip levels in the restore chart (default of 1 is used). I will open another issue about this which will fix in separate PR.

adriansuarez (approved):
Looks good.

stable/restore/README.md
sivanov-nuodb: When restoring multi-cluster it's easier to just restore one of the clusters and let the other SMs sync from it. Alternatively, the "selected" backupset should be put on a remote location available to both clusters so that all SMs are restored.

I have some text about multi-cluster in the restore chart README. More should come in B&R documentation.

```
In multi-cluster deployments, different backup cron jobs execute with different schedules.
Specific backupsets are produced for database deployment in each cluster.
To perform a database restore in multi-cluster deployment, select the hotcopy SMs in one of the clusters and provide a backupset available in that cluster or `:group-latest` as `restore.source`.
```

PR #185, opened 2021-02-11
Fix broken alt-address test in bamboo

vegichan:
This test stopped working in Bamboo with NuoDB 4.2.

A simple refactor to use JSON instead of a grep/cut fixes the test.

### DIFF
--- a/test/minikube/minikube_base_database_test.go
+++ b/test/minikube/minikube_base_database_test.go
@@ -183,27 +183,20 @@ func verifyPacketFetch(t *testing.T, namespaceName string, admin0 string) {
 	t.Log("tar contents: ", output)
 }
 
-func verifyEngineAltAddress(t *testing.T, namespaceName string, admin0 string, expectedNrEngines int) {
+func verifyEngineAltAddress(t *testing.T, namespaceName string, adminPod string, expectedNrEngines int) {
 	kubectlOptions := k8s.NewKubectlOptions("", "", namespaceName)
 
-	podNames, err := k8s.RunKubectlAndGetOutputE(t, kubectlOptions, "exec", admin0, "--",
-		"bash", "-c",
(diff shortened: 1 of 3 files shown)


test/minikube/minikube_base_database_test.go
```diff
@@ -183,27 +183,20 @@ func verifyPacketFetch(t *testing.T, namespaceName string, admin0 string) {
...
-		"nuocmd get processes | grep -o \"alt-address: [^}]*\" | cut -f2 -d' '")
-	require.NoError(t, err)
-	actualAltAddressesSlice := strings.Split(strings.TrimSuffix(actualAltAddresses, "\n"), "\n")
+	output, err := k8s.RunKubectlAndGetOutputE(t, kubectlOptions, "exec", adminPod, "--",
+		"nuocmd", "--show-json", "get", "processes", "--db-name", "demo")
+	require.NoError(t, err, output)
 
-	require.True(t, len(podNamesSlice) == expectedNrEngines, "Expected number of process names don't match")
-	require.True(t, len(actualAltAddressesSlice) == expectedNrEngines, "Expected number of process addresses don't match")
+	err, objects := testlib.Unmarshal(output)
```
adriansuarez, line 193: How does the parser handle multiple JSON object in the same string, e.g.
```
{
  "startId": "0",
  ...
}
{
  "startId": "1",
  ...
}
```

vegichan: Go does not care. It continues unmarshalling the string until it runs out of top-level objects. We have tests for that. It can handle this form of JSON.

adriansuarez (approved):

sivanov-nuodb (commented):
Thanks for fixing this @vegichan ! Way better than what was before.

PR #187, opened 2021-02-23
NuoDB Backup and Restore in Kubernetes (doc)

sivanov-nuodb:
Added user guide on how to use Backup and Restore mechanics with NuoDB Helm Charts in Kubernetes.

### DIFF
--- a/docs/BackupAndRestore.md
+++ b/docs/BackupAndRestore.md
@@ -152,7 +152,8 @@ Standard protocols such as _HTTP/s_, _FTP_ and _SFTP_ can be used as a source.
 Any other protocols that have built-in support in the shipped `cURL` binary in the NuoDB image can be used as well.
 Basic authentication or an authentication token that is embedded in the URL are supported.
 The remote URL should point to the `tar.gz` file which will be automatically extracted using the configured `stripLevels` setting (by default `1`). 
-These are the number of leading path elements removed by the `tar` command during archive extraction.
+These are the number of leading path elements removed by the `tar` command during archive extraction so that the root data (e.g for a _stream_, the *.atm files) can be accessed.
+Typically, the `tar.gz` only contains one parent level folder, being the database name, for example _demo_. 
 
 - backupset available in the backup root directory.
 HC SMs have a backup volume attached and are selected for hot copy operation during the scheduled backups. Only these SMs have access to local hot copies and their archives can be restored using a local backup set.
--- a/docs/BackupAndRestore.md
+++ b/docs/BackupAndRestore.md
@@ -4,7 +4,7 @@
 
 NuoDB provides several automated mechanisms for backing up and restoring a database in a Kubernetes environment when deploying NuoDB using the NuoDB Helm Charts.
 
-After database installation, the available backup and restore features supported by these mechanisms are: 
+After database installation, the available backup and restore mechanisms are: 
 
 1. [Automatic archive initial import](#automatic-archive-initial-import)
 2. [Scheduled online database backups](#scheduled-online-database-backups)
@@ -15,7 +15,7 @@ After database installation, the available backup and restore features supported
 7. [Distributed Database restore with storage groups](#distributed-database-restore-with-storage-groups)
 8. [Assisted database restore](#assisted-database-restore)
 
-Several of these features require additional configuration before using.
+Several of these mechanisms require additional configuration before using.
 
 > **NOTE**: For information on NuoDB backup and restore operations in non-Kubernetes deployments of NuoDB, see [Backing Up and Restoring Databases](https://doc.nuodb.com/nuodb/latest/deployment-models/physical-or-vmware-environments-with-nuodb-admin/database-operations/backing-up-and-restoring-databases/). 
 The documentation in this repository expands on the product documentation and is specific to this Helm Chart repository.
@@ -46,7 +46,7 @@ Normally used to speed up the SYNC process for the so-called "accelerated" sync 
 
 ## Compatibility Matrix
 
-The table below shows the compatibility matrix for different backup and restore features in combination with the NuoDB product version and NuoDB Helm Charts.
+The table below shows the compatibility matrix for different backup and restore mechanisms in combination with the NuoDB product version and NuoDB Helm Charts.
 
 |                                       | NuoDB 4.0.x+ Helm Charts 2.x.x | NuoDB 4.0.x+ Helm Charts 3.0.x | NuoDB 4.0.x+ Helm Charts 3.1.x | NuoDB 4.2.x+ Helm Charts 3.2.x+ |
 |---------------------------------------|--------------------------------|--------------------------------|---------------------------------|---------------------------------|
@@ -64,9 +64,9 @@ Additional capability was added into the `nuobackup` script allowing the removal
 
 [2] - For deployments that contain nonHC SMs, it's recommended that database in-place restore is done by manually starting database engines in sequential order. First start the HC SMs, then nonHC SMs and last the TEs.
 
-[3] - Automatic archive import is recommended when restoring a backup set taken from a different environment.
+[3] - _Automatic archive initial import_ is recommended when restoring a backup set taken from a different environment.
 
-[4] - Archive seed restore is not deterministic. Once the restore request is created, the first SM that starts or is restarted will perform the restore of its archive.
+[4] - _Archive seed restore_ is not deterministic. Once the restore request is created, the first SM that starts or is restarted will perform the restore of its archive.
 
 For detailed information about features, enhancements, and fixed problems, please check [NuoDB Release Notes](https://doc.nuodb.com/nuodb/latest/release-notes/) and [NuoDB Helm Charts Release Notes](https://github.com/nuodb/nuodb-helm-charts/releases)
 
@@ -75,7 +75,7 @@ For detailed information about features, enhancements, and fixed problems, pleas
 ### Scheduled online database backups
 
 Multiple backup cron jobs are scheduled as part of the [database](../stable/database) chart deployment by default.
-The backup schedules can be customized to meetdifferent backup and recovery objectives.
+The backup schedules can be customized to meet different backup and recovery objectives.
 They are exposed as Helm chart values in particular:
 - `database.hotCopy.fullSchedule` - configures the schedule of _full_ backup Kubernetes job
 - `database.hotCopy.incrementalSchedule` - configures the schedule of _incremental_ backup Kubernetes job
@@ -249,8 +249,8 @@ All TEs will wait for the database restore to complete before starting, hence we
 ...
 ```
 
-The database restore request will be cleared after a successful database in-place restore and can be checked using `nuodocker get restore requests --db-name <db>` command.
-The database state should be manually verified by `nuocmd show domain` and using SQL queries to ensure that it's the desired one after a successful restore.
+The database restore request will be cleared after a successful database in-place restore and can be viewed using `nuodocker get restore requests --db-name <db>`.
+The database state should be manually verified by `nuocmd show domain` and using SQL queries to ensure that it's in the desired state after a successful restore.
 
 If there is a need to modify the database restore request, a new database shutdown and restart cycle is needed so that the request can be completed.
 The same applies if the reason for the previously failed database restore is corrected.
@@ -258,7 +258,7 @@ When troubleshooting distributed database restore failures, it's important to co
 
 > **NOTE**: In multi-cluster deployments, different backup cron jobs execute with different schedules unless they are configured to be part of the same backup group.
 
-Backup sets produced by different backup groups can't be used together to perform distributed database in-place restore.
+Backup sets produced by different backup groups can't be used together to perform _distributed database in-place restore_.
 To perform a database restore in multi-cluster deployment, proceed with one of the approaches described below.
 
 1. [Select the hot copy SMs in one of the clusters](#fine-grained-archive-selection) and provide a backup set available in that cluster or `:group-latest` as `restore.source`.
@@ -279,13 +279,13 @@ For instance, this can be used to easily select all SMs in a specific backup gro
 
 ### Archive seed restore
 
-The archive seed restore operation restores a corrupted or lost archive while the database is running.
+The _archive seed restore_ operation restores a corrupted or lost archive while the database is running.
 This operation will reset the existing archive state, and once it joins the database, the new SM will sync to the current state of the running database.
 For this reason, you cannot restore the entire database by restoring a single SM in a running database.
-The _seed restore_ reduces the _SYNCing_ time from other running SMs as only the changed atoms will be transferred.
+The _seed restore_ has the potential to reduce the _SYNCing_ time from other running SMs as only the changed atoms will be transferred.
 The database will remain running throughout this operation.
 
-The high-level steps to perform archive seed restore are the following:
+The high-level steps to perform _archive seed restore_ are the following:
 
 1. Identify which archives will be restored.
 2. Ensure that the restore source is available either locally in the backup volume or as a remote URL.
@@ -297,7 +297,7 @@ In this example, there is an SM exited because it experienced archive data corru
 One way to fix the issue is to completely delete the corrupted archive and let the SM sync the whole archive from one of the running processes in the database.
 To reduce the sync time, we can upload one of the recent online database backups to a remote location and restore the corrupted archive.
 
-> **NOTE**: A copy of an archive or backup set from another database can't be used to perform archive seed restore. Otherwise the Storage Manager process will fail to start with error _Archive "/var/opt/nuodb/archive/nuodb/demo" doesn't match database.  Expected UUID \*\*\*, got \*\*\*._
+> **NOTE**: A copy of an archive or backup set from another database can't be used to perform _archive seed restore_. Otherwise the Storage Manager process will fail to start with error _Archive "/var/opt/nuodb/archive/nuodb/demo" doesn't match database.  Expected UUID \*\*\*, got \*\*\*._
 
 Identify its archive id using `nuocmd show archives` commands and select it for restore by installing the _restore_ chart with `restore.type="archive"`.
 
@@ -343,9 +343,9 @@ drwxr-xr-x  2 nuodb root 4096 Feb 19 10:32 demo-save-20210219T103236
 
 ### Automatic archive initial import
 
-The automatic archive initial import is a special case of a database restore.
-It allows specifying an external URL that contains a `tar.gz` archive copy or a backup set which will be automatically downloaded and used to define the database initial state, avoiding the need to do a post-deployment database load manually.
-Initial import can be used for importing a backup taken from a different environment. It is configured in the `database.autoImport` section and will execute for every Storage Manager in the database deployment when all of the below conditions are met:
+The _automatic archive initial import_ is a special case of a database restore operation.
+This operation allows the user to specify an external URL that contains a `tar.gz` archive copy or a backup set which will be automatically downloaded and used to define the database initial state, avoiding the need to do a post-deployment database load manually.
+Initial import is used for importing a backup taken from a different environment. It is configured in the `database.autoImport` section and will execute for every Storage Manager in the database deployment when all of the below conditions are met:
 
 - `database.autoImport.source` and `database.autoImport.type` are configured
 - the archive hasn't been initialized yet - `1.atm` is missing from the archive directory
@@ -359,7 +359,7 @@ To use automatic archive import in a pre-existing environment, it should be dest
 4. List removed archives associated with the database by using `nuocmd show archives --db-name demo --removed`.
 5. Remove all archives associated with the database by using `nuocmd delete archive --archive-id <id> --purge`.
 
-The high-level steps to configure automatic archive import to bootstrap a new database are the following:
+The high-level steps to configure the automatic archive import operation to bootstrap a new database are the following:
 
 1. Prepare the import source and upload it to a remote location available to NuoDB Storage Manager pods.
 2. Set `database.autoImport` variables during NuoDB database chart installation.
@@ -387,7 +387,7 @@ nginx_pod=$(kubectl get pod \
 kubectl cp "${file}" --namespace web "${nginx_pod}:/usr/share/nginx/html/${file}"
 ```
 
-To avoid connectivity issues it's recommended to verify that the remote source is available to pods in the NuoDB namespace.
+To avoid connectivity issues, it's recommended to verify that the remote source is available to pods in the NuoDB namespace.
 We'll start a sample busybox pod to validate the URL.
 
 ```bash
@@ -395,7 +395,7 @@ kubectl run -ti busybox --rm --image=busybox --restart=Never -n nuodb -- \
   wget --spider http://nginx.web.svc.cluster.local/$file
 ```
 
-Automatic archive import is typically used to create new database deployment.
+Automatic archive import is typically used to create a new database deployment.
 After installing the NuoDB admin chart, deploy the NuoDB database chart and enable automatic archive import.
 
 ```bash
@@ -435,25 +435,25 @@ drwxr-xr-x 2 nuodb root 4096 Feb 19 10:32 demo
 
 ### Automatic archive restore
 
-Automatic archive restore is a special case of seed restore.
+_Automatic archive restore_ is a special case of the seed restore operation.
 It allows SM to automatically restore its archive if it is corrupted or missing from a predefined restore source so that the accelerated sync use case is supported automatically.
-Although this feature can be used with a remote URL, it is typically used with `:latest` or `:group-latest` configured as a restore source.
+Although this mechanism can be used with a remote URL, it is typically used with `:latest` or `:group-latest` configured as a restore source.
 
-Automatic archive restore is configured in the `database.autoRestore` section and will execute for every Storage Manager in the database deployment when all of the below conditions are met:
+_Automatic archive restore_ is configured in the `database.autoRestore` section and will execute for every Storage Manager in the database deployment when all of the below conditions are met:
 
 - `database.autoRestore.source` and `database.autoRestore.type` are configured
 - archive metadata is found for the archive in the admin layer or on disk
 - configured restore source is available in the SM pod
 - the archive is detected as corrupted or missing - if the atom files count is less than 20 and catalog atom files count is less than 2
 
-> **NOTE**: Automatic archive restore using local hot copy as restore source can be used only for HC SMs.
+> **NOTE**: _Automatic archive restore_ with a local hot copy is used only for HC SMs.
 Alternatively, a remote _URL_ containing recent backup can be configured to perform a seed restore for all SMs.
 
-To see the feature in action, we can simulate a situation where archive data in a running database for one of the HC SMs is lost.
+To demonstrate, we can simulate a situation where archive data in a running database for one of the HC SMs is lost.
 This will cause the Storage Manager process to assert once it hits the disk and Kubernetes will automatically restart the _engine_ container.
-After the restart automatic archive restore will be performed reducing the archive sync time and making sure that the database is back in the original capacity as soon as possible.
+After, the restart _automatic archive restore_ will be performed, reducing the archive sync time and returning the database back to its original capacity as soon as possible.
 
-Configure automatic archive restore during initial installation of the database chart or later on:
+Configure _automatic archive restore_ during initial installation of the database chart or later on:
 
 ```bash
 helm upgrade --install database nuodb/database \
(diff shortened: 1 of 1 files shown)


docs/BackupAndRestore.md
```diff
@@ -0,0 +1,594 @@
+# NuoDB Backup and Restore in Kubernetes
+
+## Introduction
+
+NuoDB provides several automated mechanisms for backing up and restoring a database in a Kubernetes environment when deploying NuoDB using the NuoDB Helm Charts.
+
+After database installation, the available backup and restore features supported by these mechanisms are: 
+
+1. [Automatic archive initial import](#automatic-archive-initial-import)
```
vegichan, line 9: is the initial import done "after database installation"?

docs/BackupAndRestore.md
```diff
@@ -0,0 +1,594 @@
...
+[4] - Archive seed restore is not deterministic. Once the restore request is created, the first SM that starts or is restarted will perform the restore of its archive.
+
+For detailed information about features, enhancements, and fixed problems, please check [NuoDB Release Notes](https://doc.nuodb.com/nuodb/latest/release-notes/) and [NuoDB Helm Charts Release Notes](https://github.com/nuodb/nuodb-helm-charts/releases)
+
+## Backup and Restore Mechanisms
+
+### Scheduled online database backups
+
+Multiple backup cron jobs are scheduled as part of the [database](../stable/database) chart deployment by default.
+The backup schedules can be customized to meetdifferent backup and recovery objectives.
```
vegichan, line 78: missing space between meet, different

docs/BackupAndRestore.md
```diff
@@ -0,0 +1,594 @@
...
+- selecting archives served by database processes with specific labels - `restore.labels` can be used to configure process labels, which then define the archives which will be selected for a restore.
+Any configured label and value that matches an SM process will add its archive to the list of selected archives for a restore.
+For instance, this can be used to easily select all SMs in a specific backup group - `--set restore.labels.backup="<backup group>"`.
+
+### Archive seed restore
+
+The archive seed restore operation restores a corrupted or lost archive while the database is running.
+This operation will reset the existing archive state, and once it joins the database, the new SM will sync to the current state of the running database.
+For this reason, you cannot restore the entire database by restoring a single SM in a running database.
+The _seed restore_ reduces the _SYNCing_ time from other running SMs as only the changed atoms will be transferred.
```
vegichan, line 285: in general, that's exactly what SYNC does too. In the absence of bugs, a full SM sync should be as fast as seed+diff-sync. I know that in reality, it is not.

But can we still change the wording to something like "The seed restore - has the potential to reduce - the syncing time"?

vegichan (approved):

docs/BackupAndRestore.md
sivanov-nuodb: It is done after `database` chart installation by `nuosm` before SMs database processes are started.

docs/BackupAndRestore.md
sivanov-nuodb: I see the _archive seed restore_ only used when the archive needs to be re-initialized. It compares full archive sync with seed+diff sync here. In absence of bugs, the seed+diff sync should be a lot faster than transferring all atoms.
I agree that in any other use case it's hard to know what the gain will be so I will change the text as suggested.

docs/BackupAndRestore.md
```diff
@@ -0,0 +1,594 @@
...
+- `database.hotCopy.incrementalSchedule` - configures the schedule of _incremental_ backup Kubernetes job
+- `database.hotcopy.journalBackup.intervalMinutes` - configures the schedule of _journal_ backup Kubernetes job. _journal_ hot copy must be explicitly enabled using `database.hotcopy.journalBackup.enabled`.
+
+[Cron expression format](https://en.wikipedia.org/wiki/Cron) documents the format of the first two schedules listed above.
+
+An HC SM has a backup volume attached and is selected for hot copy operation during a scheduled backup.
+Each _full_ hot copy will create a backup set with the same name in the backup root directory (by default `/var/opt/nuodb/backup`) on all HC SMs in a single database and a single backup group.
+By default, all HC SMs in a single Kubernetes cluster are part of one backup group.
+This can be adjusted by changing the `database.hotcopy.backupGroup` value for multiple Kubernetes clusters in multi-cluster deployment.
+_Incremental_ and _journal_ hot copies are always stored in the current backup set if one exists.
```
adriansuarez, line 90: The current backup set is always the latest one?

docs/BackupAndRestore.md
```diff
@@ -0,0 +1,594 @@
...
+If there is no current backup set for some of the archives served by a HC SM when _incremental_ or _journal_ backup is requested, a _full_ hot copy is triggered automatically.
+The `nuobackup` script is made available as a configMap and is used by all backup cron jobs to send a hot copy request of the specified type to the HC SMs.
+The _Full_ hot copy result is recorded into the NuoDB Admin domain key-value (KV) store upon successful finish.
```
adriansuarez, line 93: This seems to answer my questions above, but perhaps this could be reworded to make it clearer. Also, should we include the KV key format, or `nuocmd` invocation that can be used to obtain the current/latest full backup?

docs/BackupAndRestore.md
```diff
@@ -0,0 +1,594 @@
...
+
+It is possible to disable backup jobs creation by setting `hotcopy.enableBackups` to `false`.
+Custom backup jobs that use the `nuobackup` script can be configured to support more complex workflows and backup strategies.
+
+### Restore/Import source and type
+
+The restore/import source location can be one of the following:
+- remote _URL_ in a form of `protocol://authority/path`.
+The restore logic uses [cURL](https://curl.se/) to download a stream of data into an SM pod and restore the archive from it.
+Standard protocols such as _HTTP/s_, _FTP_ and _SFTP_ can be used as a source.
```
adriansuarez, line 140: I guess it can be any protocol that curl supports, since the URL is passed through verbatim. Or do we actually restrict the protocol to these three?

docs/BackupAndRestore.md
```diff
@@ -0,0 +1,594 @@
...
+Any other protocols that have built-in support in the shipped `cURL` binary in the NuoDB image can be used as well.
+Only basic authentication can be specified for remote sources. A remote location that has an authentication token embedded into the _URL_ can also be used.
```
adriansuarez, line 142: The second sentence sort of contradicts the "Only" part of the first. I would reword this as

> Basic authentication or an authentication token that is embedded in the URL are supported.

I guess this rules out certificate authentication or authentication other than basic auth that is passed as a request header (e.g. `Authorization: Bearer <token>`).

docs/BackupAndRestore.md
```diff
@@ -0,0 +1,594 @@
...
+- backupset available in the backup root directory.
+HC SMs have a backup volume attached and are selected for hot copy operation during the scheduled backups. Only these SMs have access to local hot copies and their archives can be restored using a local backup set.
+
+The most recent successful backup set name can be referenced in one of the following ways:
+- `:latest` - The latest backup set name among all backup groups.
+- `:group-latest` - The latest backup set name for the backup group in current database deployment.
+
+Two restore/import source types are available:
+- _stream_ - Used as a synonym for a copy of NuoDB archive which contains both the archive and journal directories.
+A remote _stream_ source is downloaded and extracted directly into the archive directory which doesn't require extra space for holding the restore/import source file.
```
adriansuarez, line 155: Somewhat unrelated, but do we support separate journal directories at all with the StatefulSets?

docs/BackupAndRestore.md
sivanov-nuodb: Thanks! I will change it as suggested.

docs/BackupAndRestore.md
sivanov-nuodb: Currently no.

docs/BackupAndRestore.md
sivanov-nuodb: The _current_ backup set is always the one that was issued (it may or may not have finished).

docs/BackupAndRestore.md
sivanov-nuodb: Good point! I would suggest including the command wrappers rather than the KV prefixes.
The current backup set is available via `nuodocker get current backup --db-name <db>`.
The latest successful backup set is available via `nuobackup --type report-latest --db-name <db> --group <backup group>`.

docs/BackupAndRestore.md
```diff
@@ -0,0 +1,594 @@
...
+The database restart during the restore process is controlled by the `restore.autoRestart` setting value (by default _"true"_).
+Users can retain control to manually stop and restart the database pods by disabling the auto-restart during the _restore_ chart installation.
+All SMs serving restored archives will need to be started together so that the database restore process can complete.
+
+The high-level steps to perform database in-place restore are the following:
+
+1. Identify the restore source which can be either a backup set created by scheduled backup jobs or a remote source.
+2. Identify which archives need to be restored.
+By default, all archives served by HC SMs in a single Kubernetes cluster will be selected for a restore as they can define the database state by restoring from a backup set available in the backup volumes.
+3. Ensure that the restore source is available on all SMs selected for restore.
```
adriansuarez, line 179: I would add a note saying that if the restore source is a backup set, then only HC SMs will have access to the restore source.

docs/BackupAndRestore.md
```diff
@@ -0,0 +1,594 @@
...
+4. Ensure that there is enough free disk space in the archive volume of each SM so that a copy of the archive before the restore is taken.
```
adriansuarez, line 180: Is a copy taken, or are the existing contents of the archive moved? And what would the contents of the archive directory be after restore is complete for a restore and non-restore (i.e. syncing) SM. I would guess the following:
- Restore SM: Archive restored from backup set or copy ("stream"). The backup set and the archive directories would be on different volumes. I think for "stream" we are extracting the backup as we download it (e.g. `curl ... | tar -xf`). So in both cases, the archive volume would need enough space to accommodate the archive we want to restore to and the backup we create of the current contents.
- Non-restore SM: Wouldn't we need the archive to accommodate the backup we are creating and the archive that we would sync after start-up, which in the absence of user-defined SGs, would be the same as the "restore SM".

docs/BackupAndRestore.md
```diff
@@ -0,0 +1,594 @@
...
+Typically all HC SMs deployed in a single cluster are part of one backup group.
+- `Named backup group`= Backup group that has a custom name configured by the user.
+- `Seed restore` = A restore that allows a single archive to be restored using a backup. The archive is then updated with the latest transactions and brought current through the database _SYNCing_ process from a RUNNING SM.
+Normally used to speed up the SYNC process for the so-called "accelerated" sync use case.
+
+## Compatibility Matrix
+
+The table below shows the compatibility matrix for different backup and restore features in combination with the NuoDB product version and NuoDB Helm Charts.
+
+|                                       | NuoDB 4.0.x+ Helm Charts 2.x.x | NuoDB 4.0.x+ Helm Charts 3.0.x | NuoDB 4.0.x+ Helm Charts 3.1.x | NuoDB 4.2.x+ Helm Charts 3.2.x+ |
```
acabrele, line 51: Can we make it clear somewhere that this documents functionality of the last column, and there is no previous equvilent document?

docs/BackupAndRestore.md
```diff
@@ -0,0 +1,594 @@
...
+This can be adjusted by changing the `database.hotcopy.backupGroup` value for multiple Kubernetes clusters in multi-cluster deployment.
+_Incremental_ and _journal_ hot copies are always stored in the current backup set if one exists.
+If there is no current backup set for some of the archives served by a HC SM when _incremental_ or _journal_ backup is requested, a _full_ hot copy is triggered automatically.
+The `nuobackup` script is made available as a configMap and is used by all backup cron jobs to send a hot copy request of the specified type to the HC SMs.
+The _Full_ hot copy result is recorded into the NuoDB Admin domain key-value (KV) store upon successful finish.
+
+The backup jobs execution should be monitored on a regular basis to ensure that they complete successfully.
+By default, each backup job execution will be retried on failure which can be configured by the  `database.hotCopy.restartPolicy` setting.
+There are no backup retention policies currently available. The admin user is responsible to make sure that there is enough space in the backup volume of each SM for new hot copies to be created.
```
acabrele, line 97: This is really important info an feels a little lost between the other two sentences about the job histories. Can we pull it into a section on retention management?  Where we could also mentioned the possibility of using customer provided read write many volumes (EFS, azure file) enabling retention management more easily using cloud tooling for example.

docs/BackupAndRestore.md
```diff
@@ -0,0 +1,594 @@
...
+
+### Restore/Import source and type
+
+The restore/import source location can be one of the following:
+- remote _URL_ in a form of `protocol://authority/path`.
+The restore logic uses [cURL](https://curl.se/) to download a stream of data into an SM pod and restore the archive from it.
+Standard protocols such as _HTTP/s_, _FTP_ and _SFTP_ can be used as a source.
+Any other protocols that have built-in support in the shipped `cURL` binary in the NuoDB image can be used as well.
+Only basic authentication can be specified for remote sources. A remote location that has an authentication token embedded into the _URL_ can also be used.
+The remote URL should point to the `tar.gz` file which will be automatically extracted using the configured `stripLevels` setting (by default `1`).
```
acabrele, line 143: Can we add what the aim is (e.g strip levels needs to remove all parent folders)

docs/BackupAndRestore.md
```diff
@@ -0,0 +1,594 @@
...
+
+- backupset available in the backup root directory.
+HC SMs have a backup volume attached and are selected for hot copy operation during the scheduled backups. Only these SMs have access to local hot copies and their archives can be restored using a local backup set.
+
+The most recent successful backup set name can be referenced in one of the following ways:
+- `:latest` - The latest backup set name among all backup groups.
+- `:group-latest` - The latest backup set name for the backup group in current database deployment.
+
+Two restore/import source types are available:
+- _stream_ - Used as a synonym for a copy of NuoDB archive which contains both the archive and journal directories.
```
acabrele, line 154: Use the terminology you previously defined? - "Offline backup"

docs/BackupAndRestore.md
```diff
@@ -0,0 +1,594 @@
...
+A remote _stream_ source is downloaded and extracted directly into the archive directory which doesn't require extra space for holding the restore/import source file.
+- _backupset_ - Hot copy backup set which is either fetched from a remote source or available in the backup directory.
```
acabrele, line 156: like the previous line, explain this requires additional space for download/extract?

docs/BackupAndRestore.md
```diff
@@ -0,0 +1,594 @@
...
+The database restart during the restore process is controlled by the `restore.autoRestart` setting value (by default _"true"_).
+Users can retain control to manually stop and restart the database pods by disabling the auto-restart during the _restore_ chart installation.
+All SMs serving restored archives will need to be started together so that the database restore process can complete.
+
+The high-level steps to perform database in-place restore are the following:
+
+1. Identify the restore source which can be either a backup set created by scheduled backup jobs or a remote source.
+2. Identify which archives need to be restored.
+By default, all archives served by HC SMs in a single Kubernetes cluster will be selected for a restore as they can define the database state by restoring from a backup set available in the backup volumes.
+3. Ensure that the restore source is available on all SMs selected for restore.
```
acabrele, line 179: I thought we had changes so that if the source didn't exist it would proceed to sync instead?  Maybe we are more explicit now...

acabrele: I guess there is a step/action here to copy the backup set to the SMs, not just check?  Which may not be possible if they are crash looping for example. 

We can't guarantee that a backupset name will be identical on all HCSMs just because they all ran on the same schedule,there could be seconds different

docs/BackupAndRestore.md
```diff
@@ -0,0 +1,594 @@
...
+4. Ensure that there is enough free disk space in the archive volume of each SM so that a copy of the archive before the restore is taken.
+5. Invoke the database restore request by installing NuoDB _restore_ chart.
+By default, the database set as `restore.target` will be **shutdown** and restarted.
+
+Restoring to a previous point in time starts by installing the `restore` chart which will invoke the database restore request.
```
acabrele, line 184: Careful with terminology overlap with true PiT restore

docs/BackupAndRestore.md
```diff
@@ -0,0 +1,594 @@
...
+In this example, a specific backup set that is available on all HC SMs will be used as a restore source.
+
+> **NOTE**: If the restore source is set to `:latest` or `:group-latest`, it's recommended that the incremental and journal backups are temporarily suspended. This is to ensure that another backup is not accidentally taken.
```
adriansuarez, line 187: Would it be possible to check the "restore request" metadata in `nuobackup` and would that solve this race condition?

docs/BackupAndRestore.md
acabrele: Also is this referring to the restore chart or the autoimport, or both (I thought striplevels was removed from the restore chart)

docs/BackupAndRestore.md
```diff
@@ -0,0 +1,594 @@
...
+
+- `database.autoImport.source` and `database.autoImport.type` are configured
+- the archive hasn't been initialized yet - `1.atm` is missing from the archive directory
+- no archive metadata is found for the archive in the admin layer or on disk
+
+To use automatic archive import in a pre-existing environment, it should be destroyed first so that the existing domain is ready to host the new database with the same name.
+
+1. Delete the database chart deployment.
+2. Remove the associated PVC, PV, and cloud storage volume that remains after database deletion.
+3. Remove the database from the domain by using `nuocmd delete database --db-name <db>`.
```
acabrele, line 358: Describe the option of deleting the nuoadmin storage too? (instead of nuocmd delete)

docs/BackupAndRestore.md
```diff
@@ -0,0 +1,594 @@
...
+> **NOTE**: In multi-cluster deployments, different backup cron jobs execute with different schedules unless they are configured to be part of the same backup group.
+
+Backup sets produced by different backup groups can't be used together to perform _distributed database in-place restore_.
+To perform a database restore in multi-cluster deployment, proceed with one of the approaches described below.
+
+1. [Select the hot copy SMs in one of the clusters](#fine-grained-archive-selection) and provide a backup set available in that cluster or `:group-latest` as `restore.source`.
+2. Make the selected backup set available as a remote URL in both clusters. It can be used as a restore source for all the SMs in the database only if no custom storage groups are configured.
+3. Configure HC SMs in all clusters to be part of a single backup group and leave one set of backup jobs enabled in only one of the clusters to control this backup group.
+This will ensure that all HC SMs backups will be coordinated during hot copy requests.
+
```
adriansuarez, line 268: This seems kind of awkward and kind of defeating the purpose of the HC SM StatefulSet. The user has to manually choose a specific backup set to make available remotely, presumably from among the HC SMs in both clusters. Can we achieve the same thing by issuing a restore request that excludes the HC SMs in one of the two clusters, so that it waits for the HC SMs in the other cluster to restore the database before joining?

docs/BackupAndRestore.md
```diff
@@ -0,0 +1,594 @@
...
+4. Invoke the database restore request by installing the NuoDB _restore_ chart by specifying the archive IDs for the selected archives.
+5. Start or restart the selected SM
+
+In this example, there is an SM exited because it experienced archive data corruption.
+One way to fix the issue is to completely delete the corrupted archive and let the SM sync the whole archive from one of the running processes in the database.
+To reduce the sync time, we can upload one of the recent online database backups to a remote location and restore the corrupted archive.
+
+> **NOTE**: A copy of an archive or backup set from another database can't be used to perform _archive seed restore_. Otherwise the Storage Manager process will fail to start with error _Archive "/var/opt/nuodb/archive/nuodb/demo" doesn't match database.  Expected UUID \*\*\*, got \*\*\*._
+
+Identify its archive id using `nuocmd show archives` commands and select it for restore by installing the _restore_ chart with `restore.type="archive"`.
```
adriansuarez, line 302: Do labels work for seed restore? `nuodocker start sm` injects the label `pod-name`, so specifying `--set restore.labels.pod-name <pod name>` might be a more convenient way to issue the seed restore command (assuming that it works).

docs/BackupAndRestore.md
```diff
@@ -0,0 +1,594 @@
...
+3. Remove the database from the domain by using `nuocmd delete database --db-name <db>`.
+4. List removed archives associated with the database by using `nuocmd show archives --db-name demo --removed`.
+5. Remove all archives associated with the database by using `nuocmd delete archive --archive-id <id> --purge`.
+
+The high-level steps to configure the automatic archive import operation to bootstrap a new database are the following:
+
+1. Prepare the import source and upload it to a remote location available to NuoDB Storage Manager pods.
+2. Set `database.autoImport` variables during NuoDB database chart installation.
+
+> **NOTE**: For simplicity _nginx_ deployment will be used in this example so that our remote source is available via HTTP in the Kubernetes cluster.
```
adriansuarez, line 367: _nginx_ is just some arbitrary name for a deployment? Should we choose something more generic sounding like _prod_, _qa_, or _dev_? _qa_ or _dev_ might make sense in this subsection, since that seems like a use-case for importing a database into a fresh domain.

docs/BackupAndRestore.md
```diff
@@ -0,0 +1,594 @@
...
+_Automatic archive restore_ is a special case of the seed restore operation.
+It allows SM to automatically restore its archive if it is corrupted or missing from a predefined restore source so that the accelerated sync use case is supported automatically.
+Although this mechanism can be used with a remote URL, it is typically used with `:latest` or `:group-latest` configured as a restore source.
+
+_Automatic archive restore_ is configured in the `database.autoRestore` section and will execute for every Storage Manager in the database deployment when all of the below conditions are met:
+
+- `database.autoRestore.source` and `database.autoRestore.type` are configured
+- archive metadata is found for the archive in the admin layer or on disk
+- configured restore source is available in the SM pod
+- the archive is detected as corrupted or missing - if the atom files count is less than 20 and catalog atom files count is less than 2
```
adriansuarez, line 447: Isn't it also a condition that the database has running SMs, otherwise this would be a full database restore? Also, how automated is this, i.e. does it require any user intervention whatsoever?

docs/BackupAndRestore.md
```diff
@@ -0,0 +1,594 @@
...
+
+- a complete set of archives serving all storage groups must be restored when performing database in-place restore
+- to ensure that backup coverage is complete, each storage group must be served by at least one HC SM
+
+A complete set of archives can be selected using several of the methods described in [Fine-grained archive selection](#fine-grained-archive-selection) section. NuoDB won't perform any special checks during database restore to ensure that the archives selected for a restore are a complete set of archives.
+If some of the storage groups are missing from the selection, their state won't be restored.
+
+> **NOTE**: For more information about storage groups, check [Using Table Partitions and Storage Groups](https://doc.nuodb.com/nuodb/latest/database-administration/using-table-partitions-and-storage-groups/)
+
+### Assisted database restore
```
adriansuarez, line 511: "Assisted" means "assisted by the user". I would just call this "Manual database restore" since that is clearer.

docs/BackupAndRestore.md
```diff
@@ -0,0 +1,594 @@
...
+
+_Assisted database restore_ is a special case of database restore operation and allows complex restore operations to be executed easier in Kubernetes deployments.
+The _assisted database restore_ operation is initiated by installing the _restore_ chart with `restore.manual="true"` which creates a manual restore request.
+Once the database is restarted all database processes will block waiting for the user to perform the selected archives restore.
```
adriansuarez, line 515: I wouldn't say "Once the database is restarted" because none of the database processes are actually started. I would make clear what the point is of this mode by including a sentence like

> This mode blocks all database pods before allowing them to form a database, in order to give the user access to the persistent volumes used by an SM.

adriansuarez (approved):
Excellent document! I left a few comments, but overall, looks very good.

docs/BackupAndRestore.md
sivanov-nuodb: Fixed

docs/BackupAndRestore.md
sivanov-nuodb: Archive backup is taken only on SMs selected for restore.
The existing archive content is moved to `<db>-save-<timestamp>` and `nuodocker restore archive` restore into the original archive directory which is empty.
In case remote URL pointing to backup set is used as restore source, it will be downloaded into archive directory temporarily.
I will add these in the document.

docs/BackupAndRestore.md
sivanov-nuodb: I would say that some parts of the document are specific to the last column but not all. For example, the _Scheduled online database backups_, _Automatic archive initial import_, and _Automatic archive restore_ haven't been changed a lot since the original implementation.
This compatibility matrix was my attempt to show what is supported where and to point major changes in the B&R logic.
I agree that the examples and the sample output is specific to the latest column so I will call that out.

docs/BackupAndRestore.md
sivanov-nuodb: Good call! I don't know a lot about the "cloud tooling" mentioned here.
I will include the statement as is in a new section. If you would like me to include more info, please let me know.

docs/BackupAndRestore.md
sivanov-nuodb: > Can we add what the aim is (e.g strip levels needs to remove all parent folders)

The very next sentence is explaining what `stripLevels` are used for. Do you want me to rephrase it?

> I thought striplevels was removed from the restore chart

They have never been exposed in the `restore` chart which is a bug that we should fix.

docs/BackupAndRestore.md
sivanov-nuodb: Good point. I will change it.

docs/BackupAndRestore.md
acabrele: Yes I see "These are the number of leading path elements removed by the tar command during archive extraction." But that doesn't say what the aim is or help someone know what it should be set to if it were non-default (should all leading elements be removed... should one parent be left etc)

docs/BackupAndRestore.md
sivanov-nuodb: With the latest changes, all SMs selected for a restore should be restored so that database restore is performed.
By default, only HC SMs are selected to be restored. The rest of the SMs will sync as before.

I saw aiming to consolidate all the restore source rangling in step 3. As you pointed out, the desired backup set could be made available as a remote URL or copied in the backup volume manually, etc.

> We can't guarantee that a backupset name will be identical on all HCSMs just because they all ran on the same schedule

`nuodocker backup database` guarantees that the backup set names are identical for all selected archives in each _full_ hot copy operation.

docs/BackupAndRestore.md
sivanov-nuodb: Oops ... you are right.

docs/BackupAndRestore.md
sivanov-nuodb: Yes, the restore source is available in the restore request. 
It's more of a question, should `nuobackup` abort the requested backup if a database restore using `:latest` or `:group-latest` is requested?

docs/BackupAndRestore.md
sivanov-nuodb: Option 1 will be used by default --- e.g specific backup set or `:group-latest` will restore only HC SMs in one of the clusters. The rest of the SMs will sync.

I decided to include other options in case, the user wants to leverage an accelerated sync use case for all SMs in the database.

docs/BackupAndRestore.md
sivanov-nuodb: Good idea! That should work. I will test and use that instead.

docs/BackupAndRestore.md
sivanov-nuodb: There is no check on the database state. I assume that in the general case when no user storage groups are defined --- if we automatically restore the archive from an old backup set, then it won't be selected as leader candidate or the next database cold start.

The `autoRestore` is fully automated and is triggered by SM pod start/restart.

docs/BackupAndRestore.md
sivanov-nuodb: Okay

docs/BackupAndRestore.md
sivanov-nuodb: Ah, I see. Okay, I added some extra explanation.

sivanov-nuodb:
Thanks, guys for the valuable comments and suggestions!
@acabrele , @adriansuarez I think I addressed most of the comments in case you want to have another look.

sivanov-nuodb:
Added user guide on how to use Backup and Restore mechanics with NuoDB Helm Charts in Kubernetes.

PR #188, opened 2021-02-24
Pod name should contain cluster name.

paulc4:


vegichan (approved):
quick and easy! Thanks paul

stable/admin/templates/NOTES.txt
```diff
@@ -6,4 +6,4 @@ NuoDB can be accessed via port 48004 on the following DNS name from within your
 
 To display your NuoDB cluster state run:
 
-  $ kubectl exec admin-{{ .Values.admin.domain }}-0 -- nuocmd show domain
+  $ kubectl exec admin-{{ .Values.admin.domain }}-{{ .Values.cloud.cluster.name }}-0 -- nuocmd show domain
```
sivanov-nuodb, line 9: To be more accurate I suggest that we use the following template:
```
$ kubectl exec {{ include "admin.fullname" . }}-0 -- nuocmd show domain
```

sivanov-nuodb (commented):
Thanks so much, Paul for fixing this! I added one suggestion.
I think that we have the same problem in the `database` chart notes.

stable/admin/templates/NOTES.txt
adriansuarez: +1

sivanov-nuodb (approved):

adriansuarez (approved):

PR #189, opened 2021-02-26
Multi-cluster support in test framework

sivanov-nuodb:
**Changes**
- Added CircleCI integration with Azure AKS to support multi-cluster tests
- Refactored CircleCI config file so that it supports new CircleCI 2.1 workflows and re-usable jobs. As part of this, I've named the workflow jobs and added multi-cluster tests job execution, and added go modules caching so that they are not downloaded every time.
- Added NuoDB license environment variable to support testing with multiple SMs. It should be masked in all build jobs logs.
- Enabled some of the tests that require a license to run. This included lowering some of the specs for the engines so that they fit in the local minikube installations.
- Added multi-cluster machinery in the `testlib` which allows to easily install multi-cluster environments using already available test infrastructure.
- Bumped up some of the timeouts as AKS seems a bit slower in scheduling pods (maybe due to the dynamic PV provisioning)
- Added more diagnostics in case pods are not schedules in the desired timeouts

**Testing**
- Implemented one basic multi-cluster test in a separate test suite
- Regression testing


test/multicluster/minikube_base_multicluster_test.go
```diff
@@ -0,0 +1,109 @@
...
+	require.NoError(t, err, output)
+	require.True(t, strings.Contains(output, "Storage"))
+	require.True(t, strings.Contains(output, "Transaction"))
+}
+
+func TestKubernetesBasicMultiCluster(t *testing.T) {
+	if os.Getenv("NUODB_LICENSE") != "ENTERPRISE" && os.Getenv("NUODB_LICENSE_CONTENT") == "" {
+		t.Skip("Cannot test multiple SMs without the Enterprise Edition")
+	}
+	testlib.AwaitTillerUp(t)
```
vegichan, line 29: I just noticed that we still have this TillerUp function. But we have switched to Helm3. Does it still do something or is it dead code?

test/multicluster/minikube_base_multicluster_test.go
```diff
@@ -0,0 +1,109 @@
...
+	adminReleaseName := fmt.Sprintf("admin-%s", randomSuffix)
+	namespaceName := fmt.Sprintf("kubernetesbasicmulticluster-%s", randomSuffix)
+
+	options := helm.Options{
+		SetValues: map[string]string{
+			"database.sm.resources.requests.cpu":    testlib.MINIMAL_VIABLE_ENGINE_CPU,
+			"database.sm.resources.requests.memory": testlib.MINIMAL_VIABLE_ENGINE_MEMORY,
+			"database.te.resources.requests.cpu":    testlib.MINIMAL_VIABLE_ENGINE_CPU,
+			"database.te.resources.requests.memory": testlib.MINIMAL_VIABLE_ENGINE_MEMORY,
+			"admin.readinessTimeoutSeconds":         "5",
```
vegichan, line 45: can you provide a comment why this timeout needs to be bumped and what type of issues do we see otherwise? Should we add this knowledge to some runbook?

test/multicluster/minikube_base_multicluster_test.go
```diff
@@ -0,0 +1,109 @@
...
+	})
+
+	t.Run("verifyDomain", func(t *testing.T) {
+		testlib.ExecuteInAllClusters(t, func(context *testlib.ClusterDeploymentContext) {
+			admin0 := fmt.Sprintf("%s-nuodb-%s-0", context.AdminReleaseName, context.ThisCluster.Name)
+			testlib.AwaitAdminFullyConnected(t, context.Namespace, admin0, 2)
+			testlib.AwaitDatabaseUp(t, context.Namespace, admin0, "demo", 4)
+		})
+	})
+	t.Run("verifyNuoSQL", func(t *testing.T) {
```
vegichan, line 103: I dont see any database started. What does NuoSQL do if there are no TEs?

test/multicluster/minikube_base_multicluster_test.go
sivanov-nuodb: You are right, this is not needed as we are running with Helm v3. Actually, the function does nothing if Helm 3 is used. 
It seems that we decided not to remove this code in other tests and that is why I kept it here. I'm happy to remove it if you feel strongly about it.

test/multicluster/minikube_base_multicluster_test.go
sivanov-nuodb: Good point!
This is a leftover from my local testing with K3d multi-cluster. I will remove it.

test/multicluster/minikube_base_multicluster_test.go
sivanov-nuodb: > I dont see any database started
Database deployment is installed in each cluster having 1 TE and 1 SM. All processes should go to running otherwise the _"verifyDomain"_ test will fail.

test/testlib/nuodb_admin_utilities.go
```diff
@@ -177,3 +200,26 @@ func AwaitNrLoadBalancerPolicies(t *testing.T, namespace string, podName string,
 		return err == nil && len(policies) == expectedNumber
 	}, 30*time.Second)
 }
+
+func ApplyNuoDBLicense(t *testing.T, namespace string, adminPod string) {
+	options := k8s.NewKubectlOptions("", "", namespace)
+	licenseContent := os.Getenv("NUODB_LICENSE_CONTENT")
+	if licenseContent != "" {
+		licenseContentBytes, err := base64.StdEncoding.DecodeString(licenseContent)
```
adriansuarez, line 208: Is there any risk of someone creating a fork of this repo and making a change to print out the license, so that they can use it they can copy the license text and use it in their own system? I think they would have to publish a PR so that tests are run on our CircleCI, so we would at least notice it.

sivanov-nuodb: This is a good question!
According to [Circle CI doc](https://circleci.com/docs/2.0/oss/#build-pull-requests-from-forked-repositories):

> Running an unrestricted build in a parent repository can be dangerous. Projects often contain sensitive information, and this information is freely available to anyone who can push code that triggers a build.
By default, CircleCI does not pass secrets to builds from forked PRs for open source projects and hides four types of configuration data: ...

I decided to test this explicitly, however, it seems that Circle CI builds are disabled currently on Forked Pull requests.
See https://github.com/nuodb/nuodb-helm-charts/pull/191.
@vegichan If you decide to enable this feature, we could double-check that configured secrets are not leaked in the above PR.

More on this topic:
* https://circleci.com/blog/managing-secrets-when-you-have-pull-requests-from-outside-contributors
* https://circleci.com/blog/triggering-trusted-ci-jobs-on-untrusted-forks/

adriansuarez (approved):
This is very cool. Approving, assuming that there are no security concerns.

test/multicluster/minikube_base_multicluster_test.go
vegichan: I dont. We should remove it in a separate push from everywhere

vegichan (approved):
+1 from me

PR #193, opened 2021-03-17
Upgrade kiwigrid/sidecar NuoDB Collector dependency to 1.10.8

butson:
… current version has problems on Azure that leads to REQ_URL being called too frequently which reloads telegraf and drops 1-3 collection cycles.


acabrele (approved):
Also good to move to a 1.x release!

sivanov-nuodb (approved):
Thanks!

PR #194, opened 2021-03-24
Allow docker image overrides of upgrade tests

mkysel:


test/testlib/minikube_utilities.go
```diff
@@ -244,6 +244,26 @@ func InjectTestVersion(t *testing.T, options *helm.Options) {
 	options.SetValues["nuodb.image.tag"] = image.Nuodb.Image.Tag
 }
 
+func OverrideUpgradeRepository(t *testing.T, options *helm.Options) {
+	dat, err := ioutil.ReadFile(UPGRADE_INJECT_FILE)
```
adriansuarez, line 248: Do we have an internal repository that has the same naming scheme as DockerHub? I know the internal registry that we use has some naming scheme like `registry.internal.nuodb.com/release/rel41-centos7:latest`.

adriansuarez: Looks like this is not an issue.

adriansuarez (approved):
Looks good.

test/testlib/minikube_utilities.go
```diff
@@ -244,6 +244,26 @@ func InjectTestVersion(t *testing.T, options *helm.Options) {
 	options.SetValues["nuodb.image.tag"] = image.Nuodb.Image.Tag
 }
 
+func OverrideUpgradeRepository(t *testing.T, options *helm.Options) {
```
sivanov-nuodb, line 247: The method says that it overwrites the repository but it actually overwrites also the tag unconditionally.
What if a specific test needs to test upgrade from NuoDB `4.0.8` instead of `4.0.4`?

Another option would be to re-map the tag, if set by the test, in case we are running internally. I.e `4.0.4` -> `rel-4.0.4-ce-centos7`.

mkysel: the scheme is different. The tag is 2. The repo is rel/4.0.4... It does not match the docker naming scheme

sivanov-nuodb (approved):

PR #195, opened 2021-04-06
Added stripLevels options in restore chart

sivanov-nuodb:
**Changes**
Previously it was not possible to set `stripLevels` when using the `restore` chart with remote URL and the default value of `1` was used. Exposed this setting as `restore.stripLevels` which can be set when installing the `restore` chart. The value is stored either in the user-data of the database restore request or using `$NUODB_RESTORE_REQUEST_PREFIX/$DB_NAME/restore/strip-levels` key in legacy mode.

**Testing**
- new template test
- new functional test which executes in compatibility mode in CurcleCI and using the new way internally in Bamboo
- regression testing


stable/restore/files/nuorestore
```diff
@@ -133,6 +134,9 @@ if isRestoreRequestSupported; then
       restore_credentials_encoded="$(printf "%s" "${restore_credentials}" | base64 -w 0)"
       user_data="${user_data};restore_credentials_encoded='${restore_credentials_encoded}'"
    fi
+   if [ -n "$strip_levels" ]; then
+      user_data="${user_data};strip_levels='${strip_levels}'"
```
adriansuarez, line 138: Can you break this up into two lines for readability?

test/minikube/minikube_long_restore_test.go
```diff
@@ -237,3 +237,69 @@ func TestKubernetesRestoreWithStorageGroups(t *testing.T) {
...
+
+	tePodName := testlib.GetPodName(t, namespaceName, tePodNameTemplate)
+
+	go testlib.GetAppLog(t, namespaceName, smPodName0, "_post-restart", &corev1.PodLogOptions{Follow: true})
+	go testlib.GetAppLog(t, namespaceName, tePodName, "_post-restart", &corev1.PodLogOptions{Follow: true})
+
+	// verify that the database does NOT contain the data from AFTER the backup
+	tables, err := testlib.RunSQL(t, namespaceName, admin0, "demo", "show schema User")
+	require.NoError(t, err, "error running SQL: show schema User")
+	require.True(t, strings.Contains(tables, "No tables found in schema "), "Show schema returned: ", tables)
```
mkysel, line 303: `require.Contains`?

test/testlib/nuodb_database_utilities.go
```diff
@@ -311,3 +315,37 @@ func CheckRestoreRequests(t *testing.T, namespaceName string, podName string, da
...
 }
+
+func CreateNginxDeployment(t *testing.T, namespaceName string) {
+	kubectlOptions := k8s.NewKubectlOptions("", "", namespaceName)
+	k8s.RunKubectl(t, kubectlOptions, "create", "deployment", NGINX_DEPLOYMENT, "--image=nginx")
+	k8s.RunKubectl(t, kubectlOptions, "create", "service", "clusterip", NGINX_DEPLOYMENT, "--tcp=80:80")
+	podName := GetPodName(t, namespaceName, NGINX_DEPLOYMENT)
+	AwaitPodUp(t, namespaceName, podName, 60*time.Second)
+
+	AddDiagnosticTeardown(TEARDOWN_DATABASE, t, func() {
```
mkysel, line 326: this should not be reusing the DATABASE teardown.
It should also provide its own teardown which deletes the deployment and the clusterIP service. Otherwise, it leaves garbage in the namespace.

sivanov-nuodb: Good point!

acabrele:
Is there any updated minimum compatible nuodb (nuodocker) version to use this?

sivanov-nuodb:
> Is there any updated minimum compatible nuodb (nuodocker) version to use this?

This doesn't require any change in the product (nuodocker) as the `stripLevels` are stored as user data in the restore request which is an opaque value (arbitrary string).

acabrele (approved):
Thanks!

adriansuarez (approved):

PR #197, opened 2021-04-09
Change the default timeout for all backup jobs

sivanov-nuodb:
**Issue**
Hotcopy timeout value for backup cron-jobs can lead to false positives as the underlying hot-copy operation continues to execute in the background on the SM but the job can fail due to timeout leaving a full backup set not recorded in the KV store.

**Issue**
Journal hotcopy timeout setting is not used for journal backup jobs.

**Changes**
- Default timeout for `full`, `incremental` and `journal` backup jobs has been changed to `0` which means that they should wait forever for the underlying hot-copy operation to complete.
- Journal backup timeout is not used when invoking `nuobackup`
- Make `nuobackup` non-blocking. Similar to our other commands, the user needs to specify `--timeout` if he wants it to block synchronously.

**Product changes**
Starting with NuoDB 4.2.1+ `nuodocker backup database` will use client-side polling to wait for the hot-copy operation to finish.

**Testing**
- integration tests
- regression testing


test/integration/template_backup_test.go
```diff
@@ -38,7 +38,7 @@ func TestDatabaseBackupCronJobDisabled(t *testing.T) {
 	// Run RenderTemplate to render the template and capture the output.
 	_, err := helm.RenderTemplateE(t, options, helmChartPath, "release-name", []string{"templates/cronjob.yaml"})
 
-	assert.Contains(t, err, "could not find template")
+	assert.Contains(t, err.Error(), "could not find template")
```
mkysel, line 41: should this just be `assert.ErrorContains`?

mkysel (approved):

test/integration/template_backup_test.go
sivanov-nuodb: There is no `ErrorContains` in github.com/stretchr/testify/assert. But I needed it for `Subset` function.

mkysel: I see that you changed gotest to stretcher. Got it!

I wonder why it was left behind...

stable/database/templates/cronjob.yaml
```diff
@@ -54,7 +54,6 @@ spec:
             - { name: DB_NAME,             value: "{{ .Values.database.name }}"     }
             - { name: NUOCMD_API_SERVER,   value: "{{ template "admin.address" . }}:8888" }
             - { name: PEER_ADDRESS,        value: "{{ template "admin.address" . }}" }
-            - { name: TIMEOUT,             value: "{{ .Values.database.sm.hotCopy.timeout }}" }
```
sivanov-nuodb, line 57: @NikTJ777 Do you remember why these `TIMEOUT` env vars are set?

test/integration/template_backup_test.go
mkysel: could you remove it from `go.mod` and `go.sum` so that people don't use it accidentally?

sivanov-nuodb: Good point! I will remove `gotest.tools` ... it looks like it isn't used anywhere else.

adriansuarez (approved):

PR #202, opened 2021-05-12
Validate restore source

sivanov-nuodb:
**Changes**

The restore source can be one of the following:
- a backup set name that s available in the backup root directory
- remote URL
- `:latest`
- `:group-latest`

This change validates when a "special" reserved keyword starting with `:` is provided it is one of the valid keywords. A message is logged when automatic archive repair is skipped because the backup set is not available (this is the case in nonHC SMs).

`nuote` has been changed so that `--debug` flag is added if `NUODB_DEBUG` environment variable is set.

**Testing**
- new integration tests
- regression testing

### DIFF
--- a/stable/restore/templates/_helpers.tpl
+++ b/stable/restore/templates/_helpers.tpl
@@ -152,6 +152,12 @@ Return the restore target.
 Return the restore source.
 */}}
 {{- define "restore.source" -}}
+{{- if hasPrefix ":" .Values.restore.source }}
+{{- $valid := list ":latest" ":group-latest" "" }}
+{{- if not (has .Values.restore.source $valid) }}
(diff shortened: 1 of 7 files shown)


stable/restore/templates/_helpers.tpl
```diff
@@ -152,6 +152,12 @@ Return the restore target.
 Return the restore source.
 */}}
 {{- define "restore.source" -}}
+{{- if hasPrefix ":" .Values.restore.source }}
```
NikTJ777, line 155: I like this! :)

NikTJ777 (approved):
Nice!

mkysel (approved):
simple and elegant!

PR #204, opened 2021-05-13
Validate database.sm.hotCopy.enableBackups

sivanov-nuodb:
Example pasring error:

```bash
$ helm template stable/database --set database.sm.hotCopy.enableBackups="truee" | grep full-hotcopy-demo-cronjob
Error: template: database/templates/statefulset.yaml:414:84: executing "database/templates/statefulset.yaml" at <include "defaulttrue" .Values.database.sm.hotCopy.enableBackups>: error calling include: template: database/templates/_helpers.tpl:206:4: executing "defaulttrue" at <fail (printf "Invalid boolean value: %s" .)>: error calling fail: Invalid boolean value: truee

Use --debug flag to render out invalid YAML
```


mkysel (approved):
LGTM

bkelly-ndb (approved):

PR #205, opened 2021-05-13
Fix hotCopy doc case

mkysel:


bkelly-ndb (approved):

adriansuarez (approved):
Looks good.

sivanov-nuodb (approved):
Thanks for fixing this! There are a couple of these in Backup & Restore guide in case you want to update them as well.

mkysel:
@adriansuarez I did not notice those. Will fix

PR #208, opened 2021-05-18
Skip test on NuoDB version

sivanov-nuodb:
**Changes**

- Database restore using user-defined storage groups is supported in NuoDB 4.2+. Disable `TestKubernetesRestoreWithStorageGroups` test for unsupported NuoDB versions
- Get logs for all SM pods if the restore fails for some reason

### DIFF
--- a/test/minikube/minikube_long_restore_test.go
+++ b/test/minikube/minikube_long_restore_test.go
@@ -226,6 +226,12 @@ func TestKubernetesRestoreWithStorageGroups(t *testing.T) {
 
 	defer testlib.Teardown(testlib.TEARDOWN_RESTORE)
 
+	// Get SM pod logs if the test fails
+	testlib.AddDiagnosticTeardown(testlib.TEARDOWN_DATABASE, t, func() {
+		testlib.GetAppLog(t, namespaceName, hcSmPodName0, "_post-restore", &corev1.PodLogOptions{})
+		testlib.GetAppLog(t, namespaceName, hcSmPodName1, "_post-restore", &corev1.PodLogOptions{})
+	})
+
 	// restore database
 	databaseOptions.SetValues["restore.source"] = backupset
 	testlib.RestoreDatabase(t, namespaceName, admin0, &databaseOptions)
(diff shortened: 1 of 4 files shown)


test/minikube/minikube_long_restore_test.go
```diff
@@ -146,6 +146,7 @@ func TestKubernetesRestoreWithStorageGroups(t *testing.T) {
 	if os.Getenv("NUODB_LICENSE") != "ENTERPRISE" && os.Getenv("NUODB_LICENSE_CONTENT") == "" {
 		t.Skip("Cannot test multiple SMs without the Enterprise Edition")
 	}
+	testlib.SkipTestOnNuoDBVersionCondition(t, "< 4.2")
```
mkysel, line 149: love it!

test/testlib/nuodb_database_utilities.go
```diff
@@ -355,3 +357,54 @@ func ServePodFileViaHTTP(t *testing.T, namespaceName string, srcPodName string,
 	k8s.RunKubectl(t, kubectlOptions, "cp", localFilePath, nginxPod+":/usr/share/nginx/html/"+fileName)
 	return fmt.Sprintf("http://nginx.%s.svc.cluster.local/%s", namespaceName, fileName)
 }
+
+func GetNuoDBVersion(t *testing.T, options *helm.Options) string {
+	kubectlOptions := k8s.NewKubectlOptions("", "", "default")
```
mkysel, line 362: the idea behind randomly generated namespaces was to make sure that multiple tests can be run at the same time. We don't do that much, since we dont have large available multi-tenant clusters but I would like to keep that option open. If multiple tests execute in the "default" namespace, they might collide and fail randomly.

sivanov-nuodb: good point. I will generate random namespace name for this.

test/testlib/nuodb_database_utilities.go
```diff
@@ -383,7 +383,9 @@ func GetNuoDBVersion(t *testing.T, options *helm.Options) string {
 }
 
 func SkipTestOnNuoDBVersion(t *testing.T, versionCheckFunc func(*semver.Version) bool) {
-	versionString := GetNuoDBVersion(t, &helm.Options{})
+	randomSuffix := strings.ToLower(random.UniqueId())
+	namespaceName := fmt.Sprintf("nuodbversioncheck-%s", randomSuffix)
+	versionString := GetNuoDBVersion(t, namespaceName, &helm.Options{})
```
mkysel, line 388: do you have to delete the namespace after?

sivanov-nuodb: This should be done by teardown placed in `CreateNamespace` function.

NikTJ777 (approved):
Really nice!

mkysel (approved):

PR #209, opened 2021-05-18
Add 3.2.0 -> master upgrade test

mkysel:
- Add new tests.
- Remove the 40X string from test name. It does not actually test 40X. It tests whatever the default version of NuoDB is.


NikTJ777 (approved):
Looks nice.

PR #213, opened 2021-05-26
Override thrift.message.max to enable catching up admins

adriansuarez:
Currently we cannot catch up admins with a snapshot if the Raft state is
larger than 10MB, because that is the current default value of
thrift.message.max. Override this to allow Raft state to grow up to 1GB.


mkysel (approved):
lgtm

bkelly-ndb (approved):

PR #214, opened 2021-05-27
[HELM-281] Recursively change permissions in init container

adriansuarez:
Recursively change permissions in init container so that any files that
happen to be in the persistent volume are made accessible to the nuodb
user (the nuodb user is in the root group and is given access to files
by making them group-readable and -writable).


mkysel (approved):
LGTM :)

bkelly-ndb (approved):

PR #215, opened 2021-06-01
Driveby test fixes to SplitAndRender

mkysel:
- when using SplitAndRenderStatefulSet with the DB SS, the test should make sure that both backup and non-backup SMs get tested. This is done by setting the expectedNrObjects to 2.
-remove code duplication


adriansuarez (approved):

sivanov-nuodb (approved):
Thanks

bkelly-ndb (approved):

PR #217, opened 2021-06-02
Restarting SM fails if an AP is Disconnected

sivanov-nuodb:
**Issue**
If one of the APs is partitioned from the rest of the domain or shutdown, any SM startup will fail with:

```
2021-04-21T14:57:32.810+0000 ===========================================
2021-04-21T14:57:32.815+0000 logsize=1329; maxlog=5000000
2021-04-21T14:57:32.818+0000 Directory /var/opt/nuodb/archive/nuodb/demo exists
2021-04-21T14:57:32.820+0000 trace: checking Admin layer
2021-04-21T14:58:05.358+0000 Error while checking Admin layer: Admin layer is inoperative - exiting: 'check servers' failed: Health-check failed:
Non-unique LEADER:
  (leader=admin-nuodb-cluster0-2,term=35)->[admin-nuodb-cluster0-2]
  (leader=admin-nuodb-cluster0-1,term=36)->[admin-nuodb-cluster0-0,admin-nuodb-cluster0-1]
```

**Changes**
- remove the admin layer health check as it is too strong for this purpose
- fail the SM fast if the first `nuocmd get value` call fails to indicate a problem with the admin layer

**Testing**
- add E2E test for the scenario above
- regression testing

### DIFF
--- a/stable/database/files/nuosm
+++ b/stable/database/files/nuosm
@@ -351,19 +351,6 @@ function isRestoreRequestSupported() {
 }
 
 #=======================================
-# function - ensures that admin layer has an elected leader
-#
-function checkAdminLayer() {
-  local timeout="$1"
-  local status
-  trace "checking Admin layer"
-  status="$(nuocmd check servers --check-leader --timeout "$timeout" 2>&1)"
-  if [ $? -ne 0 ]; then
-    die 1 "Admin layer is inoperative - exiting: ${status}"
-  fi
-}
-
-#=======================================
 # function - completes archive restore request
 #
 function completeRestoreRequest(){
@@ -422,11 +409,11 @@ else
   log "Directory $DB_DIR exists"
 fi
 
-# ensure the admin layer is intact...
-checkAdminLayer 30
-
 trace "retrieving restore request from raft"
-legacy_restore_requested="$( nuocmd get value --key "$restore_req" )"
+legacy_restore_requested="$( nuocmd get value --key "$restore_req" 2>&1 )"
+if [ $? -ne 0 ]; then
+  die 1 "Unable to retrieve legacy restore request: ${legacy_restore_requested}"
+fi
 
 [ -f "$DB_DIR/info.json" ] && myArchive=$(sed -e 's|.*"id":\s*\([0-9]\+\).*|\1|' "$DB_DIR/info.json")
 trace "checking archive raft metadata"
--- a/stable/database/files/nuosm
+++ b/stable/database/files/nuosm
@@ -393,6 +393,15 @@ function resolveLatestSource() {
   return 0
 }
 
+#=======================================
+# function - fetch the legacy restore request from KV store
+#
+function get_legacy_restore_request() {
+  trace "retrieving legacy restore request from KV store"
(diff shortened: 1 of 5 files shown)


stable/database/files/nuosm
```diff
@@ -422,11 +409,11 @@ else
...
 
-# ensure the admin layer is intact...
-checkAdminLayer 30
-
 trace "retrieving restore request from raft"
-legacy_restore_requested="$( nuocmd get value --key "$restore_req" )"
+legacy_restore_requested="$( nuocmd get value --key "$restore_req" 2>&1 )"
+if [ $? -ne 0 ]; then
+  die 1 "Unable to retrieve legacy restore request: ${legacy_restore_requested}"
+fi
```
adriansuarez, line 416: Do we have any convenience function in here to retry a command? If we really want to have the equivalent behavior to the previous health check, we could retry the `nuocmd get value` command a few times until 30 seconds have elapsed (I think there is an 8-second timeout, so there would be about 3 re-tries). But I'm not sure if we actually ever needed the original health check in the first place, so this might be fine.

adriansuarez (approved):
Looks good to me.

test/testlib/nuodb_admin_utilities.go
```diff
@@ -223,3 +224,53 @@ func ApplyNuoDBLicense(t *testing.T, namespace string, adminPod string) {
...
+	}
+	for _, server := range servers {
+		if server.ConnectedState.State == "Connected" && server.LocalRoleInfo.LeaderServerId != "" {
+			return server.LocalRoleInfo.LeaderServerId, nil
+		}
+	}
+	return "", errors.New("no leader found")
+}
+
+func AwaitDomainLeader(t *testing.T, namespace string, adminPod string, actionFn func(string),
```
mkysel, line 252: its a bit weird that you are calling the lambda inside of the Await function but it has no impact on its logic. It does not define when the await succeeds, it just executes the lambda exactly once before returning. Do you have to pass the lambda down, or can you just call it in the caller?

mkysel:
```func AwaitDomainLeader(t *testing.T, namespace string, adminPod string, timeout time.Duration) (leader string){
	Await(t, func() bool {
		var err error
		leader, err = GetDomainLeaderE(t, namespace, adminPod)
		if err != nil {
			return false
		}
		
		return true
	}, timeout)
	
	return
}```

sivanov-nuodb: The `testlib.Await` function doesn't support returning a value, hence we need to pass another lambda to be executed once a leader is found. The `actionFn` can be any callback including an empty one. If you prefer I can rename the routine to something like `ExecuteWithDomainLeader`.

mkysel: `Await` doesnt have to return a value. You can bind to a local variable.

sivanov-nuodb: I looked again at the example you provided and got your point. Thanks!

sivanov-nuodb:
I pushed another commit which simplifies the test case and reduces the test time by around 2 min.

mkysel (approved):

PR #218, opened 2021-06-08
Add the option to mount the SM journal on a different volume

mkysel:
Storage Manager journals should be located on faster disks than Storage Manager archives. Allow the option of separating the archive from the journal and mount them on different volumes.


mkysel:
I missed the doc change. Will fix

stable/database/templates/statefulset.yaml
```diff
@@ -63,10 +63,22 @@ spec:
       - name: init-disk
         image: {{ template "init.image" . }}
         imagePullPolicy: {{ default "" .Values.busybox.image.pullPolicy | quote }}
-        command: ['chmod' , '770', '/var/opt/nuodb/archive', '/var/log/nuodb']
+        command:
+          - 'chmod'
+          - '770'
+          - '/var/opt/nuodb/archive'
+          - 'var/opt/nuodb/backup'
```
mkysel, line 70: this does not belong here!

stable/database/files/nuosm
```diff
@@ -11,6 +11,7 @@
 
 : ${NUODB_ARCHIVEDIR:=/var/opt/nuodb/archive}
 : ${NUODB_BACKUPDIR:=/var/opt/nuodb/backup}
+: ${NUODB_JOURNALDIR:=/var/opt/nuodb/journal}
```
adriansuarez, line 14: Do we need to define our own default, rather than relying on the internal product default?

stable/database/files/nuosm
```diff
@@ -748,9 +761,16 @@ trace "executing nuodocker to start SM"
...
+  log "journal-dir=${JOURNAL_DIR}"
+fi
+
 # expects NUOCMD_API_SERVER to be set.
 if [ -n "${NUODB_OPTIONS}" ] ; then
-    exec nuodocker "${nuodocker_flags[@]}" start sm --archive-dir "${DB_DIR}" --dba-user "${DB_USER}" --dba-password "${DB_PASSWORD}" --db-name "${DB_NAME}" --options "${NUODB_OPTIONS}" "$@"
+    exec nuodocker "${nuodocker_flags[@]}" start sm --archive-dir "${DB_DIR}" "${journal_flags[@]}" --dba-user "${DB_USER}" --dba-password "${DB_PASSWORD}" --db-name "${DB_NAME}" --options "${NUODB_OPTIONS}" "$@"
 else
-    exec nuodocker "${nuodocker_flags[@]}" start sm --archive-dir "${DB_DIR}" --dba-user "${DB_USER}" --dba-password "${DB_PASSWORD}" --db-name "${DB_NAME}" "$@"
+    exec nuodocker "${nuodocker_flags[@]}" start sm --archive-dir "${DB_DIR}" "${journal_flags[@]}" --dba-user "${DB_USER}" --dba-password "${DB_PASSWORD}" --db-name "${DB_NAME}" "$@"
```
adriansuarez, line 775: I really wish `nuosm` would just pass all arguments verbatim to `nuodocker start sm` rather than duplicating the command-line processing and injecting its own defaults. Is it possible to just make `--journal-dir <the journal directory>` one of the opaque arguments that is passed verbatim? That would probably require converting some the Bash logic to Helm logic, but it would preserve the behavior of not specifying any explicit journal directory, allowing the SM to choose `${DB_DIR}/journal`.

adriansuarez: Actually, it looks like `journal_flags` would be empty, so we are relying on the internal default.

stable/database/templates/statefulset.yaml
NikTJ777: Wherever it lands, there are a few points on the change:
1. the new code seems to no longer chmod on /var/log/nuodb
    1a and this chmod needs to be 'chmod -R'
2. the path 'var/opt/nuodb/backup' should be '/var/opt/nuodb/backup'

stable/database/files/nuosm
mkysel: this is what @acabrele set for the default. Opinions?

stable/database/templates/statefulset.yaml
```diff
@@ -393,10 +440,22 @@ spec:
       - name: init-disk
         image: {{ template "init.image" . }}
         imagePullPolicy: {{ default "" .Values.busybox.image.pullPolicy | quote }}
-        command: ['chmod' , '770', '/var/opt/nuodb/archive', 'var/opt/nuodb/backup', '/var/log/nuodb']
+        command:
+          - 'chmod'
+          - '770'
+          - '/var/opt/nuodb/archive'
+          - 'var/opt/nuodb/backup'
+          - '/var/log/nuodb'
```
mkysel, line 448: @NikTJ777 says:
Wherever it lands, there are a few points on the change:

the new code seems to no longer chmod on /var/log/nuodb
1a and this chmod needs to be 'chmod -R'
the path 'var/opt/nuodb/backup' should be '/var/opt/nuodb/backup'

stable/database/files/nuosm
adriansuarez: Actually, I don't think this comment is relevant, since `$NUODB_JOURNALDIR` is only used if `SEPARATE_JOURNAL=true`.

stable/database/templates/statefulset.yaml
mkysel: OK. I added -R and fixed the missing `/`. Your first point is incorrect. It still does `/var/log/nuodb`

stable/database/templates/statefulset.yaml
```diff
@@ -63,10 +63,21 @@ spec:
       - name: init-disk
         image: {{ template "init.image" . }}
         imagePullPolicy: {{ default "" .Values.busybox.image.pullPolicy | quote }}
-        command: ['chmod' , '770', '/var/opt/nuodb/archive', '/var/log/nuodb']
+        command:
+          - 'chmod'
+          - '770'
+          - '/var/opt/nuodb/archive'
+          - '/var/log/nuodb'
```
mkysel, line 70: @NikTJ777 says:
Wherever it lands, there are a few points on the change:

the new code seems to no longer chmod on /var/log/nuodb
1a and this chmod needs to be 'chmod -R'
the path 'var/opt/nuodb/backup' should be '/var/opt/nuodb/backup'

mkysel: the section I commented above relates to the non-HC SM. I will fix the -R here too.

stable/database/files/nuosm
mkysel: yeah. That is how I do the wiring. I do not allow the user to define the location of the journal. It always goes to `/var/opt/nuodb/journal/<DOMAIN>/<DBNAME>`. But of course only if `Values.database.sm.noHotCopy.journalPath.enable` is set to true

docs/HowToArchiveJournal.md
```diff
@@ -0,0 +1,31 @@
...
+In addition to the archive, NuoDB storage managers also maintain a write-ahead-log called the `journal`.
+The journal has the onerous task of ensuring durability in the face of an unexpected process termination.
+E.g., a power loss, machine meltdown, or an unexpected cloud outage.
+The journal will make sure that the archive is reconstructed in a consistent state when coming back online.
+
+For more info on NuoDB database journaling, please consult the official [NuoDB docs](https://doc.nuodb.com/nuodb/latest/database-administration/about-database-journaling/).
+
+## Separating the Journal from the Archive
+
+Since the `journal` has to write all commits durably to disk, the speed of the disk directly influences to the commit latency.
```
acabrele, line 18: nit: remove "to"

docs/HowToArchiveJournal.md
```diff
@@ -0,0 +1,31 @@
...
+To achieve the best performance, NuoDB recommends placing the `journal` on the fastest disk available.
+
+On the other hand, the `archive` can be located on slightly slower disks since writes to the archive are asynchronous.
+
+By default, the `journal` is located in a subdirectory of the `archive`.
+To achieve the best cost vs speed tradeoff, you can separate the journal from the archive.
```
acabrele, line 24: This implies we would _only_ put the journal on fast disk.  In reality we would probably always want (in production) both archive _and_ journal on their own fast _and_ separate disks.

If the archive is on slower disk then we improve "burst" capacity but at some point will throttle.  I think its something like this from worst to best:
1. archive + journal on same slow disk
2. archive + journal on separate slow disk
3. archive + journal on on same fast disk
4. archive on slow disk, journal on fast disk (good burst capacity)
5. archive + journal on separate fast disk

docs/HowToArchiveJournal.md
```diff
@@ -0,0 +1,31 @@
...
+To achieve that set the database helm template values `database.sm.noHotCopy.journalPath.enabled` and `database.sm.hotCopy.journalPath.enabled` to `true` and configure it with the desired persistence settings.
+
+Kubernetes stateful sets volume mounts are immutable and as such, the setting can not not be changed easily on an existing database.
+More on upgrades below [Upgrading existing domains](#upgrading-existing-domains).
```
acabrele, line 28: Without down time...

With down time it shouldn't be too hard.

acabrele:
When we do a restore or auto import, we can use as the source an archive or a backupset.  In the case of a backupset, the journal is not an issue - a new one is created in the journal location when the DB is started.  But in the case of an archive, the DB will expect the previous journal directory to be present (and it is assumed that the archive was properly provided with the journal inside the archive directory).

Do we need to handle moving the journal from inside the archive to the separate journal directory specifically when the autoImport _stream_ type is used?   Or does the call to` nuodocker restore archive` handle the journal cleanly like it would a backupset?

test/minikube/minikube_base_database_test.go
```diff
@@ -365,3 +365,34 @@ func TestKubernetesStartDatabaseShrinkedAdmin(t *testing.T) {
...
+				"database.sm.resources.requests.cpu":    testlib.MINIMAL_VIABLE_ENGINE_CPU,
+				"database.sm.resources.requests.memory": testlib.MINIMAL_VIABLE_ENGINE_MEMORY,
+				"database.te.resources.requests.cpu":    testlib.MINIMAL_VIABLE_ENGINE_CPU,
+				"database.te.resources.requests.memory": testlib.MINIMAL_VIABLE_ENGINE_MEMORY,
+				"database.sm.hotCopy.journalPath.enabled":       "true",
+			},
+		}
+
+		databaseReleaseName := testlib.StartDatabase(t, namespaceName, admin0, &options)
+		t.Logf(databaseReleaseName)
```
sivanov-nuodb, line 395: Probably a leftover from local runs?

test/minikube/minikube_long_journal_migration_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+		statefulSets := testlib.FindAllStatefulSets(t, namespaceName)
+
+		// Stage 2: Delete non-HC SM, upgrade to journalPath and restart
+
+		testlib.ScaleStatefulSet(t, namespaceName, statefulSets.SmNonHCSet.Name, 0)
+		testlib.AwaitDatabaseUp(t, namespaceName, admin0, "demo", 2)
+
+		nonHCPVC := fmt.Sprintf("archive-volume-sm-%s-nuodb-cluster0-demo-0", databaseReleaseName)
+		testlib.DeletePVC(t, namespaceName, nonHCPVC)
```
sivanov-nuodb, line 96: This will force another PV to be created for the archive and the non-HC SM should do full archive sync from the running SM. Is my understanding correct? If so can we have a comment about this as it's not obvious?

test/minikube/minikube_long_journal_migration_test.go
```diff
@@ -0,0 +1,123 @@
...
+				"database.sm.hotCopy.journalPath.enabled": "false",
+			},
+		}
+
+		databaseReleaseName := testlib.StartDatabase(t, namespaceName, admin0, &options)
+
+		options.SetValues["database.sm.hotCopy.journalPath.enabled"] = "true"
+
+		err := helm.UpgradeE(t, &options, testlib.DATABASE_HELM_CHART_PATH, databaseReleaseName)
+		require.Error(t, err)
```
sivanov-nuodb, line 45: What the error will look like? I suppose that Helm will complain that the PVC are immutable. Is there a reason to assert the message?

test/testlib/nuodb_admin_utilities.go
```diff
@@ -41,7 +41,7 @@ func CreateNamespace(t *testing.T, namespaceName string) {
 	}
 
 	// this method is async
-	go GetK8sEventLog(t, namespaceName)
+	//go GetK8sEventLog(t, namespaceName)
```
sivanov-nuodb, line 44: Is this a leftover?

sivanov-nuodb (commented):
@mkysel Should we introduce the journal separation feature for SM daemon sets as well?
I left some minor comments and questions.

sivanov-nuodb:
> Or does the call to nuodocker restore archive handle the journal cleanly like it would a backupset?
`nuodocker` will invoke `nuoarchive restore`, which will do journal recovery, only if the origin directory is a `backupset`. In the case of an archive, it will just clean the `info.json` file.

The automated _autoImport_ case that we have downloads a model archive which actually has something in the journal dir but it doesn't use the new separated journal.
```
journal
-- 3e2d0593-e884-d245-3ab3-dd19a70f1100
---- njf_0
------ 5.njf
```

docs/HowToArchiveJournal.md
mkysel: I will open a separate PR to wrap up the docs here!

stable/database/templates/statefulset.yaml
```diff
@@ -63,10 +63,22 @@ spec:
...
         image: {{ template "init.image" . }}
         imagePullPolicy: {{ default "" .Values.busybox.image.pullPolicy | quote }}
-        command: ['chmod' , '770', '/var/opt/nuodb/archive', '/var/log/nuodb']
+        command:
+          - 'chmod'
+          - '-R'
+          - '770'
+          - '/var/opt/nuodb/archive'
+          - '/var/log/nuodb'
+  {{- if eq (include "defaultfalse" .Values.database.sm.noHotCopy.journalPath.enabled) "true"}}
```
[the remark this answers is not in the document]
mkysel: It's a template function that compares booleans and strings. true is true.

test/minikube/minikube_long_journal_migration_test.go
mkysel: its here to force KAA to remove that SM with a purge

test/minikube/minikube_long_journal_migration_test.go
mkysel: yeah, I did not check for the exact message.

mkysel:
I will change the `daemonsets` too.

docs/HowToArchiveJournal.md
adriansuarez: I don't know if cloud providers actually make a distinction between latency and throughput, but the system would actually be bound by the write latency of the journal disk and the write throughput of the archive disk.

It would be possible to schedule writes to the archive in a way that maximizes write throughput so that a slower archive disk can keep up with a faster journal disk. Even though there is a lot of room for improvement, I think we probably already benefit from write batching done by the filesystem since writes to the archive are asynchronous and we are not fence-posted with lots of `fsync` calls, the way writes to the journal are. And this would probably be more effective if the journal has its own disk, since journal `fsync`s would not interfere with batching of archive writes.

test/minikube/minikube_long_journal_migration_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+		testlib.ScaleStatefulSet(t, namespaceName, statefulSets.SmNonHCSet.Name, 0)
+		testlib.AwaitDatabaseUp(t, namespaceName, admin0, "demo", 2)
+
+		nonHCPVC := fmt.Sprintf("archive-volume-sm-%s-nuodb-cluster0-demo-0", databaseReleaseName)
+
+		// trigger Kubernetes-Aware-Admin to purge this archive from the database
+		testlib.DeletePVC(t, namespaceName, nonHCPVC)
+
+		testlib.DeleteStatefulSet(t, namespaceName, statefulSets.SmNonHCSet.Name)
```
adriansuarez, line 100: I think if you reversed the order of deleting the statefulset and PVC (deleted the statefulset first), KAA would _not_ delete the associated archive ID. Just something we might want to call out.

mkysel: That is correct. I wanted to file a task for this, but it escaped my mind. I will do so on Monday

docs/HowToArchiveJournal.md
acabrele: I don't remember seeing the ability in any clouds to _choose_ your latency.  I suspect it will naturally improve with higher / provisioned IOPS because of the backing disk being SSD based.  AWS I believe is supposed to deliver "single digit latency" on SSD based volumes.  At least having separate  PV's would also allow the use of local SSD volumes to get sub ms performance of the journal.  There is probably a task here (now we have the ability to separate) to find the best combination - for example there are also throughput optimised disk types.

acabrele:
> > Or does the call to nuodocker restore archive handle the journal cleanly like it would a backupset?
> 
> `nuodocker` will invoke `nuoarchive restore`, which will do journal recovery, only if the origin directory is a `backupset`. In the case of an archive, it will just clean the `info.json` file.
> 
> The automated _autoImport_ case that we have downloads a model archive which actually has something in the journal dir but it doesn't use the new separated journal.
> 
> ```
> journal
> -- 3e2d0593-e884-d245-3ab3-dd19a70f1100
> ---- njf_0
> ------ 5.njf
> ```

I think you are agreeing that there is a gap here that needs to be addressed?  If journal recovery does not occur when the source is a copy-of-archive (and we don't prepare by moving the journal from inside the copy-of-archive to the seperate journal location) then it seems two things could occur:

- DB Starts OK with a new journal, ignores the existing (I guess no data loss consequence if DB starts OK)
- DB won't start with the error like  `Node startup failed: Previous Storage Manager for this archive had journaling enabled and did not stop cleanly: journal 46171df9-af6b-424c-72b2-3de4d47416b6 not found in directory....`

sivanov-nuodb:
> I think you are agreeing that there is a gap here that needs to be addressed?

Yes, I think that the latter will happen, i.e:

```Node startup failed: Previous Storage Manager for this archive had journaling enabled and did not stop cleanly: journal 46171df9-af6b-424c-72b2-3de4d47416b6 not found in directory....```

File: stable/database/templates/_helpers.tpl [hunk 15]
```diff
@@ -209,6 +209,23 @@ it by calling typeIs "bool" https://github.com/Masterminds/sprig/issues/111
 {{- end -}}
 {{- end -}}
 
+{{/*
+Takes a boolean as argument return it's value if it was defined or return true otherwise
```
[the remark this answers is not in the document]
mkysel: well spotted

mkysel:
#222 is a proposal to remove daemonsets

PR #220, opened 2021-06-10
Rewrite upgrade database tests

mkysel:
These tests were failing silently. Rewrite them so that we can identify problems in the future.

Also, pick up a change to admin options that fixes upgrade from 3.1.0.


sivanov-nuodb (commented):
Looks good. Thanks for the extra checks!

PR #222, opened 2021-06-21
Remove option to start Storage Managers as DaemonSets

mkysel:


adriansuarez (approved):

sivanov-nuodb (approved):
It looks like we had good overall coverage for daemonsets.
Thanks for doing this Martin!

PR #223, opened 2021-06-25
Validate boolean helm variables

sivanov-nuodb:
**Issue**

Journal CronJobs are enabled if `database.sm.hotCopy.journalBackup.enabled` is set to `false` (string) or `garbage` in values file.

```bash
$ grep -A2 "journalBackup" values.yaml
      journalBackup:
        enabled: "false"
        intervalMinutes: 15
$ helm template stable/database -f values.yaml | grep journal-
          - "mem 256Mi journal-hot-copy enable alt-address $(NUODB_ALT_ADDRESS)"
```

**Changes**
- Add validation for bool values in `defaulttrue` and `defaultfalse` templates
- Change the conditions which involve bool variables to use the new templates
- Remove PVC for daemonset which is not needed anymore

**Testing**
- Added more integration tests
- Regresion testing


mkysel (approved):
LGTM

bkelly-ndb (approved):

PR #227, opened 2021-07-26
Collect logging from all TEs and SMs

mkysel:
- TestChangingJournalLocationWithMultipleSMs was not collecting pre-restart data.
- DB teardown would only collect 1 SM log. If the test uses the ENTERPRISE license there can be many SMs running. Fix the code so that it collects all currently running SMs.
- DB teardown would only collect 1 TE log. There can be many TEs running. Fix the code so that it collects all currently running TEs.


adriansuarez (approved):

bkelly-ndb (approved):

sivanov-nuodb (approved):
Thanks for doing this!

PR #232, opened 2021-08-02
Rotating NuoDB TLS Certificates doc

sivanov-nuodb:


docs/HowToCertificatesRotation.md
```diff
@@ -0,0 +1,473 @@
...
+kubectl exec -ti generate-nuodb-certs -- \
+  nuocmd create keypair \
+    --keystore /tmp/keys/nuoadmin.p12 --store-password "$PASSWD" --ca \
+    --dname "CN=*.nuodb.svc.cluster.local, OU=Eng, O=NuoDB, L=Boston, ST=MA, C=US, SERIALNUMBER=67890"
+```
+
+In the example above, we specify the `SERIALNUMBER` field so that the new certificate has a different distinguished name from the current certificate.
+
+The new server certificates should be signed by the CA certificate.
+If the TLS keys have been generated using the `setup-keys.sh` script, the private/public key pair of the CA are available in the `ca.p12` file.
```
acabrele, line 371: Should we comment about storing the ca.p12 file securely here, since it can be used to sign new certificates, it shouldn't be left around in the _generate-nuodb-certs_ pod, locally, or copied to all NuoDB pods?

docs/HowToCertificatesRotation.md
```diff
@@ -0,0 +1,473 @@
...
+```bash
+kubectl cp generate-nuodb-certs:/tmp/keys/. /tmp/keys
+
+kubectl cp /tmp/keys/ca.cert nuodb/admin-nuodb-cluster0-0:/tmp/ca.cert
+```
+
+Add the new CA certificate to the truststore of all admin and database processes:
+
+```bash
+kubectl exec -ti admin-nuodb-cluster0-0 -- \
```
acabrele, line 344: I think there is an overlap here that needs to be explained between copying _ca.cert_ to the pod and running _nuocmd add trusted-certificate_, and the earlier steps of _kubectl create secret generic nuodb-ca-cert-renewed_.

A long time ago, I remember having to create a combined ca.cert with both old and new in the k8s secret / vault before running _add trusted-certificate_ but that was for the shared model.

docs/HowToCertificatesRotation.md
```diff
@@ -0,0 +1,473 @@
...
+
+```bash
+kubectl cp generate-nuodb-certs:/tmp/keys/. /tmp/keys
+```
+
+If the server certificates are signed by a public CA, then follow the official steps to renew your certificates.
+
+> **NOTE**: For information on how to create the server keystore when using certificates signed by a Public CA, see [here](https://doc.nuodb.com/nuodb/latest/deployment-models/physical-or-vmware-environments-with-nuodb-admin/domain-operations/enabling-tls-encryption/using-certificates-signed-by-a-public-certificate-authority/).
+
+Update domain keystore with the renewed certificates as described in section [Update Key Pair Certificates](#update-key-pair-certificates).
```
acabrele, line 400: This line about referring to _Update Key Pair Certificates_ is omitted from the _Rotate CA Certificate_ section

docs/HowToCertificatesRotation.md
```diff
@@ -0,0 +1,473 @@
...
+
+The `--timeout` argument specifies the amount of time to wait for the new certificate to be propagated to the truststore of every process in the domain.
+
+Esure that the new client certificate with alias `nuocmd_prime` is trusted by all APs and database processes as described in [Verify Domain Certificates](#verify-domain-certificates).
+
+Update the client key as described in section [Update Key Pair Certificates](#update-key-pair-certificates).
+
+### Cleanup
+
+Backup the `ca.p12` keystore file as it will be needed during the next keys rotation.
```
acabrele, line 449: maybe the security warning about ca.p12 should go here

docs/HowToCertificatesRotation.md
```diff
@@ -0,0 +1,473 @@
...
+        - [Rotate Server Certificate](#rotate-server-certificate)
+        - [Rotate Client Certificate](#rotate-client-certificate)
+        - [Cleanup](#cleanup)
+
+<!-- /TOC -->
+</details>
+
+## Introduction
+
+When using TLS encryption, it is necessary to rotate key pair certificates before the certificates expire.
```
acabrele, line 26: It is a customers decision, but should we recommend some base expiry times & review points?  All the examples use validity 36500

acabrele (commented):
Some questions / suggestions added.  Great job trying to lay this out in a simplified way.  It requires some jumping back and forth to understand the whole, I wonder if the intro should encourage a read-through before blindly following the steps.

docs/HowToCertificatesRotation.md
```diff
@@ -0,0 +1,473 @@
...
+6. The new CA certificate is added to the truststore of every SQL client.
+7. The new CA certificate is added to the truststore of every NuoDB Command client.
+8. For each AP, the keystore is replaced. Make sure that the new key pair certificate is in effect.
+9. For each database process, the process is restarted, so that its certificate chain is based on the new certificate.
+10. _(Only if step 3 is performed)_ For each NuoDB Command client, the client PEM file is replaced. Make sure that the new key pair certificate is in effect.
+11. _(Optional)_ The old CA certificate is removed from the truststore of every AP and every database process.
+12. _(Optional)_ The old CA certificate is removed from the truststore of every SQL client and NuoDB Command client.
+13. _(Optional)_ The old client certificate is removed from the truststore of every AP.
+
+Usually, only the server key pair certificate is renewed, which means that only the keystore file has to be updated for all APs and database processes.
```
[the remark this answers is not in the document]
sivanov-nuodb: I will add "server key" in the terminology section.

docs/HowToCertificatesRotation.md
```diff
@@ -0,0 +1,473 @@
...
+This reduces the steps needed during key rotation as the new key pair certificate can be verified using the old truststore certificate.
+
+An overview of how server key pair certificate rotation is performed is as follows:
+
+1. A new shared admin new key pair certificate is generated.
+2. _(Optional)_ A new client key pair certificate and PEM file are generated.
+3. For each AP, the keystore is replaced. Make sure that the new key pair certificate is in effect.
+4. For each database process, the process is restarted, so that its certificate chain is based on the new certificate.
+5. _(Only if step 3 is performed)_ For each NuoDB Command client, the client PEM file is replaced. Make sure that the new key pair certificate is in effect.
```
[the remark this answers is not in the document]
sivanov-nuodb: Good catch! 
Yes, it should be `step 2`. This is a leftover as before we had only one step-by-step guide.

docs/HowToCertificatesRotation.md
adriansuarez: Good point. We should probably use 365 in the examples to be more realistic.

docs/HowToCertificatesRotation.md
```diff
@@ -0,0 +1,473 @@
...
+6. _(Only if step 3 is performed)_ The old client certificate is removed from the truststore of every AP.
+
+Complete step by step examples can be found in [NuoDB TLS Key Rotation](#nuodb-tls-key-rotation) section.
```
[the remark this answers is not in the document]
sivanov-nuodb: Okay.

docs/HowToCertificatesRotation.md
```diff
@@ -0,0 +1,473 @@
...
+Since the new key pair certificates must be verified using a new trusted certificate, existing processes (admin and database) must have the new trusted certificate propagated to their truststores before the new key pair certificates can be introduced.
+
+To simplify updating and propagating new trusted certificates, run `nuocmd add trusted-certificate`.
+This command adds a trusted certificate to the NuoDB Admin server and causes the trusted certificate to be added and propagated to all APs and all database processes.
+
+An overview of how CA certificate rotation is performed is as follows:
+
+1. A new CA certificate is generated.
+2. A new shared admin new key pair certificate is generated.
+3. _(Optional)_ A new client key pair certificate and PEM file are generated.
```
adriansuarez, line 62: I think we should just remove this step. The rotation policy on client credentials should be totally separate from CA rotation.

docs/HowToCertificatesRotation.md
sivanov-nuodb: Good point!

docs/HowToCertificatesRotation.md
```diff
@@ -0,0 +1,473 @@
...
+4. _(Only if step 3 is performed)_ The new client certificate is added to the truststore of every AP.
+5. The new CA certificate is added to the truststore of every AP and every database process.
+6. The new CA certificate is added to the truststore of every SQL client.
+7. The new CA certificate is added to the truststore of every NuoDB Command client.
+8. For each AP, the keystore is replaced. Make sure that the new key pair certificate is in effect.
+9. For each database process, the process is restarted, so that its certificate chain is based on the new certificate.
```
adriansuarez, line 68: I would rephrase this to "... so that it obtains a certificate signed by the new CA".

docs/HowToCertificatesRotation.md
```diff
@@ -0,0 +1,473 @@
...
+12. _(Optional)_ The old CA certificate is removed from the truststore of every SQL client and NuoDB Command client.
+13. _(Optional)_ The old client certificate is removed from the truststore of every AP.
+
+Usually, only the server key pair certificate is renewed, which means that only the keystore file has to be updated for all APs and database processes.
+This reduces the steps needed during key rotation as the new key pair certificate can be verified using the old truststore certificate.
+
+An overview of how server key pair certificate rotation is performed is as follows:
+
+1. A new shared admin new key pair certificate is generated.
+2. _(Optional)_ A new client key pair certificate and PEM file are generated.
```
adriansuarez, line 80: Same comment as above about decoupling client credentials from server credentials.

docs/HowToCertificatesRotation.md
sivanov-nuodb: I'm not sure that I understand the first part. The `nuodb-ca-cert-renewed` is the actual certificate that `nuocmd` uses to verify the admin. In the steps we use the exact same `ca.cert` file for both invoking `nuocmd add trusted-certificate` and creating the secret.

> A long time ago, I remember having to create a combined ca.cert with both old and new in the k8s secret / vault before running add trusted-certificate but that was for the shared model.

I believe that you are referring to updating the certificate that the SQL clients use instead of the one that is added to admin truststore? Step 2 from https://doc.nuodb.com/nuodb/latest/deployment-models/physical-or-vmware-environments-with-nuodb-admin/domain-operations/enabling-tls-encryption/rotating-key-pair-certificates/#_rotating_keys_for_the_shared_admin_key_trust_model. The new CA certificate should be added to the SQL clients truststore, which is documented below. (I will also include an example).

I think that there is no need to combine the old and the new CA certificate for `nuocmd add trusted-certificate` to work. @adriansuarez Can you confirm?

Another question is should the `nuodb-ca-cert-renewed` include the old CA certificate. I think _no_ because `nuodocker start admin` doesn't use the REST. The database chart is upgraded only after the admin certificates are fully rotated so there will be no admin pod running with the old key pair.
I've tested the steps with just 2 admins so I will double-check them with 3 to be sure.

docs/HowToCertificatesRotation.md
sivanov-nuodb: At the end of the _Rotate CA Certificate_ there is a line indicating that all steps in the `Rotate Server Certificate` section should be executed which includes _Update Key Pair Certificates_.

docs/HowToCertificatesRotation.md
sivanov-nuodb: I totally agree. Maybe I copied this from the `setup-keys.sh` script.  ;)

docs/HowToCertificatesRotation.md
sivanov-nuodb: I will add another overview just for rotating the client key then. This will help to remove completely all these conditional steps.

docs/HowToCertificatesRotation.md
acabrele: > I'm not sure that I understand the first part. The nuodb-ca-cert-renewed is the actual certificate that nuocmd uses to verify the admin. In the steps we use the exact same ca.cert file for both invoking nuocmd add trusted-certificate and creating the secret.

Sorry my statement was not clear.  I agree they are the same thing - hence the overlap I meant is that a reader might ask _Why am I supplying the same file via a secret as well as having to copy into the pod manually (to a different location) and use it to run nuocmd add trusted-certificate_.  We know why but from a readers perspective it seems like duplication.

> I think that there is no need to combine the old and the new CA certificate for nuocmd add trusted-certificate to work.

Yes step 2 there looks familiar.  Before, I combined it only for the secret/vault value, not for the nuocmd command (that only used the new one).  Step 2 combines it for the file in /etc/nuodb/keys/ (which is the file in the secret / vault) before running _nuocmd add trusted-certificate_ for the new certificate only.  I only raise it as a checkpoint, it has been a while now since I played with vault rotation.

sivanov-nuodb:
I've pushed another commit that should address most of the comments so far.
The major change is that the overview steps are separated into 3 guides - CA, server, and client certificate rotation. It helps to remove the conditional steps and hopefully it looks more clear now.

docs/HowToCertificatesRotation.md
```diff
@@ -0,0 +1,486 @@
...
+1. A new CA certificate is generated.
+2. A new shared admin new key pair certificate is generated.
+3. The new CA certificate is added to the truststore of every AP and every database process.
+4. The new CA certificate is added to the truststore of every SQL client.
+5. The new CA certificate is added to the truststore of every NuoDB Command client.
+6. For each AP, the keystore is replaced. Make sure that the new key pair certificate is in effect.
+7. For each database process, the process is restarted, so that its certificate chain is based on the new certificate.
+8. _(Optional)_ The old CA certificate is removed from the truststore of every AP and every database process.
+9. _(Optional)_ The old CA certificate is removed from the truststore of every SQL client and NuoDB Command client.
+10. _(Optional)_ The old client certificate is removed from the truststore of every AP.
```
adriansuarez, line 74: This is leftover from the removed steps.

docs/HowToCertificatesRotation.md
```diff
@@ -0,0 +1,486 @@
...
+It is recommended to read through the entire page to get a common understanding of the key rotation process before following the steps.
+
+### Terminology
+
+- `Key` = a combination of a private key with its corresponding X509 certificate chain.
+These keys are usually saved in a PKCS12 file such as `nuoadmin.p12`.
+- `NuoDB Admin` = The administrative interface for domain and database management.
+For information on how to start the NuoDB Admin tier, see [Admin Chart](../stable/admin/README.md).
+- `CA` = Certificate Authority
+- `Sever Key` = The key pair certificate used by NuoDB Admin.
```
mkysel, line 45: *server key

docs/HowToCertificatesRotation.md
```diff
@@ -0,0 +1,486 @@
...
+> **NOTE**: For information on how to configure different TLS key models, see [Security Model of NuoDB in Kubernetes](./HowtoTLS.md).
+
+The [Shared Admin Key + Intermediate CA](./docs/HowtoTLS.md#intermediate-ca) model is used by default in NuoDB Kubernetes deployments.
+This means that the same key pair certificate is used by all APs, a Certificate Authority (CA) is used to sign them and the CA certificate is trusted by all processes in the domain.
+
+If a CA certificate needs to be renewed, then both the keystore and the truststore must be updated for all APs, and by extension, all database processes.
+As stated, processes with the new certificates should be introduced in a rolling fashion to avoid downtime, which means that processes with the old key pair certificate can verify processes with the new key pair certificate.
+Since the new key pair certificates must be verified using a new trusted certificate, existing processes (admin and database) must have the new trusted certificate propagated to their truststores before the new key pair certificates can be introduced.
+
+To simplify updating and propagating new trusted certificates, run `nuocmd add trusted-certificate`.
```
mkysel, line 60: link to https://doc.nuodb.com/nuodb/latest/reference-information/command-line-tools/nuodb-command/nuocmd-reference/#add-trusted-certificate?

docs/HowToCertificatesRotation.md
```diff
@@ -0,0 +1,486 @@
...
+  --namespace nuodb \
+  statefulset sm-database-nuodb-cluster0-demo
+
+kubectl rollout restart \
+  --namespace nuodb \
+  deployment te-database-nuodb-cluster0-demo
+```
+
+Alternatively, the database processes can be restarted manually by shutting down each process using `nuocmd shutdown process --start-id ...` command.
+Kubernetes will automatically restart the engine container which will effectively restart the database process.
```
mkysel, line 188: given that this doc is k8s specific, it might be a better idea to suggest deleting pods, rather than using `nuocmd`.

docs/HowToCertificatesRotation.md
```diff
@@ -0,0 +1,486 @@
...
+```bash
+kubectl exec -ti generate-nuodb-certs -- \
+  nuocmd create keypair \
+    --keystore /tmp/keys/nuoadmin.p12 --store-password "$PASSWD" --ca \
+    --dname "CN=*.nuodb.svc.cluster.local, OU=Eng, O=NuoDB, L=Boston, ST=MA, C=US, SERIALNUMBER=67890"
+```
+
+In the example above, we specify the `SERIALNUMBER` field so that the new certificate has a different distinguished name from the current certificate.
+
+The new server certificates should be signed by the CA certificate.
```
mkysel, line 382: `should` or `has to`? We should also say `by the same CA`, not any.

docs/HowToCertificatesRotation.md
```diff
@@ -0,0 +1,486 @@
...
+
+```bash
+kubectl exec -ti admin-nuodb-cluster0-0 -- \
+  nuocmd add trusted-certificate \
+    --alias nuocmd_prime --cert /tmp/nuocmd.cert --timeout 30
+```
+
+The `--timeout` argument specifies the amount of time to wait for the new certificate to be propagated to the truststore of every process in the domain.
+
+Esure that the new client certificate with alias `nuocmd_prime` is trusted by all APs and database processes as described in [Verify Domain Certificates](#verify-domain-certificates).
```
mkysel, line 455: *ensure

docs/HowToCertificatesRotation.md
```diff
@@ -0,0 +1,486 @@
...
+
+```bash
+kubectl exec -ti admin-nuodb-cluster0-0 \
+  --namespace nuodb -- \
+  nuocmd --show-json get certificate-info
+```
+
+The command displays the certificates information in the below sections:
+
+- `processCertificates` - information about certificates used by each database process identified by `startId`.
```
adriansuarez, line 204: This should say something like `the certificate chain used by each database process` instead of `certificates`, for clarity. Each process has only one certificate for its own identity. Same comment below for `serverCertificates`.

docs/HowToCertificatesRotation.md
```diff
@@ -0,0 +1,486 @@
...
+      "issuerName": "CN=ca.nuodb.com, OU=Eng, O=NuoDB, L=Boston, ST=MA, C=US, SERIALNUMBER=67890",
+      "subjectName": "CN=*.nuodb.svc.cluster.local, OU=Eng, O=NuoDB, L=Boston, ST=MA, C=US, SERIALNUMBER=67890"
+    },
+    "admin-nuodb-cluster0-1": {
+      "caPathLength": 2147483647,
+      "certificatePem": "-----BEGIN CERTIFICATE-----...-----END CERTIFICATE-----",
+      "expires": 4781150383000,
+      "expiresTimestamp": "2121-07-05T09:19:43.000+0000",
+      "issuerName": "CN=ca.nuodb.com, OU=Eng, O=NuoDB, L=Boston, ST=MA, C=US, SERIALNUMBER=67890",
+      "subjectName": "CN=*.nuodb.svc.cluster.local, OU=Eng, O=NuoDB, L=Boston, ST=MA, C=US, SERIALNUMBER=67890"
```
adriansuarez, line 269: The serial number is actually the same for all certificates (DB process, AP, CA)? I would expect this to be unique for each certificate generated.

docs/HowToCertificatesRotation.md
adriansuarez: @acabrele which file are you referring to, `ca.cert`, `nuoadmin.p12`, or `nuoadmin-truststore.p12`? Of those, only the K8s Secrets for `ca.cert` and `nuoadmin.p12` should be updated. `nuoadmin-truststore.p12` should not be updated manually if we are relying on `nuocmd add trusted-certificates` to propagate the new trusted certificate. It will contain the old CA certificate, and the admin will make updates to in-memory trust-store based on updates generated by the Raft configuration via `nuocmd add/remove trusted-certificates`.

docs/HowToCertificatesRotation.md
sivanov-nuodb: Okay

docs/HowToCertificatesRotation.md
sivanov-nuodb: I will update accordingly.

docs/HowToCertificatesRotation.md
sivanov-nuodb: Agree.

docs/HowToCertificatesRotation.md
sivanov-nuodb: This is an output from real deployment. It looks like I used the same `SERIALNUMBER` for the AP, CA, and client keys. The DB process doesn't have it as it is generated by the admin which probably put the engine address as CN.

I can randomize the `SERIALNUMBER` values for different keys so that there is no confusion if the same one should be used or not.

docs/HowToCertificatesRotation.md
sivanov-nuodb: Yes, sorry about that.

docs/HowToCertificatesRotation.md
sivanov-nuodb: > I've tested the steps with just 2 admins so I will double-check them with 3 to be sure.

CA certificate rotation with three APs is working fine as well.

mkysel (approved):
good work!

adriansuarez (approved):
Looks good.

PR #233, opened 2021-08-02
Fix B&R issues with separate journal directory

sivanov-nuodb:
**Issue**
This PR fixes issues discussed in https://github.com/nuodb/nuodb-helm-charts/pull/218#issuecomment-859604537.

**Changes**
- Use external journal path when creating the archive metadata using `nuodocker restore archive`. Otherwise, the archive will be created with an empty `journalPath` and the SM won't be provided with `--journal-dir` option. This problem was masking the original issue.
- Move journal content from `stream` backup to external journal directory as the journal is not consumed during engine shutdown.

**Testing**
- add a new database import test with a separate journal directory. The model import archive already has journal content (see https://github.com/nuodb/nuodb-helm-charts/pull/218#issuecomment-859678181).
- insert additional verifications in the existing test for the archive metadata when a separate journal is used
- regression testing


mkysel (approved):
looks great!

test/minikube/minikube_base_restore_test.go
```diff
@@ -330,3 +330,47 @@ func TestKubernetesAutoRestore(t *testing.T) {
...
+	admin0 := fmt.Sprintf("%s-nuodb-cluster0-0", helmChartReleaseName)
+
+	verifyPacketFetch(t, namespaceName, admin0)
+
+	t.Run("startDatabaseStatefulSet", func(t *testing.T) {
+		defer testlib.Teardown(testlib.TEARDOWN_DATABASE)
+
+		databaseOptions := &helm.Options{
+			SetValues: map[string]string{
+				"database.autoImport.source":              testlib.IMPORT_ARCHIVE_URL,
```
adriansuarez, line 353: Does this backup have a non-empty journal?

sivanov-nuodb: Yes it does. Please see https://github.com/nuodb/nuodb-helm-charts/pull/218#issuecomment-859678181.

stable/database/files/nuosm
```diff
@@ -278,11 +280,22 @@ function perform_restore() {
       rm -rf $download_dir
     fi
 
+    if [ "$SEPARATE_JOURNAL" == "true" ] && [ "$restore_type" == "stream" ]; then
+      # `nuodocker restore archive` doesn't invoke `nuoarchive` on cold archive
+      # (stream) restore type. Move the content of the internal journal
+      # directory to the external journal path manually as the journal is not
+      # consumed during SM shutdown
+      log "Moving restored journal content to $JOURNAL_DIR"
+      mv "$download_dir/journal/"* "$JOURNAL_DIR"
```
acabrele, line 289: There is a risk that, given someone using an environment with a separate archive chooses to prep a _stream_ type of restore, that they are manually moving the journal back into the archive and may not put it inside /journal/.  We have example of this with some customers.  It's probably too much to allow the journal source location within the tar to be configurable, but maybe there is a good place to comment in the restore docs.

acabrele (approved):

stable/database/files/nuosm
sivanov-nuodb: Good point. I will add a note in the B&R document about this rather than making it configurable.

PR #236, opened 2021-08-09
Remove leftover resources from multi-cluster tests

sivanov-nuodb:
**Issue**
If you push multiple commits without waiting for the multi-cluster tests to finish, CIrcleCI will stop the previous job and Gotest won't be able to do the resources teardown.

**Changes**
- added another test case which will remove long-living namespaces (older than 5 hours by default) from all clusters in the multi-cluster setup

### DIFF
--- a/test/testlib/minikube_utilities.go
+++ b/test/testlib/minikube_utilities.go
@@ -30,6 +30,9 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const NAMESPACE_NAME_PREFIX = "test"
+const NAMESPACE_RETENTION_PERIOD = 5 * time.Hour
+
 /** Lists of the teardown and diagnostic teardown funcs */
 var teardownLists = make(map[string][]func())
 var diagnosticTeardownLists = make(map[string][]func())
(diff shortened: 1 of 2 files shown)


mkysel (approved):

test/testlib/minikube_utilities.go
```diff
@@ -1195,3 +1207,15 @@ func FindAllStatefulSets(t *testing.T, namespaceName string) NuoDBStatefulSets {
 
 	return sets
 }
+
+func RemoveOrphanNamespaces(t *testing.T) {
+	kubectlOptions := k8s.NewKubectlOptions("", "", "")
+	for _, namespace := range GetNamespaces(t) {
+		if strings.HasPrefix(namespace.Name, NAMESPACE_NAME_PREFIX) &&
+			namespace.CreationTimestamp.DeepCopy().Add(NAMESPACE_RETENTION_PERIOD).Before(time.Now()) {
```
adriansuarez, line 1215: Do you actually need to make a copy? `Time.Add` seems to return a new `Time` struct rather than mutating the one it is invoked on.

adriansuarez (approved):

test/testlib/minikube_utilities.go
sivanov-nuodb: I'm trying to remember but I think that I checked the implementation and saw:
```
t.wall = t.wall&^nsecMask | uint64(nsec) // update nsec
t.addSec(dsec)
```

but after looking more closely in the documented example it looks like you are right.

PR #237, opened 2021-08-09
Add HashiCorp Vault vs Kubernetes 1.21 bug workaround and documentation of disable_iss_validation=true

mkysel:
- move HC tests into their own larger class so that we can test the engines too
- workaround HC bug
- add doc note

- root cause https://github.com/hashicorp/vault-helm/issues/562 and https://github.com/hashicorp/vault/issues/11953


test/minikube/minikube_large_instance_test.go
```diff
@@ -0,0 +1,73 @@
+// +build large
```
mkysel, line 1: everything here is just moved

adriansuarez (approved):

sivanov-nuodb (approved):
Thanks for finding the root cause!

Issue #238, opened 2021-08-12
Engine does not reconnect to a restarted admin container


mkysel:
**Describe the bug**
After a few seconds the admin gives up on waiting for the engine processes to reconnect. Logging looks like this
```
2021-08-12T18:06:23.390+0000 INFO  [admin-31wu94-nuodb-cluster0-0:processManagerScheduled29-1] ProcessManager Removing process with connectKey=1647498425067523578, startId=0, reason=Timed out (60000ms) awaiting connection from process: connectedState=PENDING_RECONNECT, removeAction=NONE, exitCode=null
2021-08-12T18:06:23.394+0000 INFO  [admin-31wu94-nuodb-cluster0-0:processManagerScheduled29-1] ProcessManager Removing process with connectKey=1026677207646886664, startId=1, reason=Timed out (60000ms) awaiting connection from process: connectedState=PENDING_RECONNECT, removeAction=NONE, exitCode=null
```

Or
```
2021-08-12T18:06:23.484+0000 WARN  [:RetryingRaftClientExec4-1] DomainProcessStateMachine admin-31wu94-nuodb-cluster0-0: No entry found for process with startId=1
```

Also
```
2021-08-12T18:07:15.393+0000 WARN  [admin-31wu94-nuodb-cluster0-0:tagServerExecutor31-2] ProcessManager Evicting unknown reconnecting process EngineNode{databaseName=demo, address=te-database-klkfbv-nuodb-cluster0-demo-69d99ff6bd-fkn7j, port=48006, type=TE, pid=44, state=RUNNING, nodeId=2, version=4.2.1.vee-3, ipAddress=10.1.5.157, hostname=te-database-klkfbv-nuodb-cluster0-demo-69d99ff6bd-fkn7j}
java.lang.NullPointerException: Expected to find durable process with startId=1
	at com.google.common.base.Preconditions.checkNotNull(Preconditions.java:987)
	at com.nuodb.host.ProcessManager$AdminConnectionImpl.setAltAddress(ProcessManager.java:952)
	at com.nuodb.host.ProcessManager$AdminConnectionImpl.reconnect(ProcessManager.java:910)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[...]

PR #239, opened 2021-08-12
Add admin helm upgrade workaround to upgrade tests

mkysel:
When the admin pod gets recreated, the engines do not always reconnect. This is worse in newer versions of K8s.

This is a known issue: https://github.com/nuodb/nuodb-helm-charts/issues/238

Re-enable the tests and add a workaround:
If the engine does not reconnect in time, kill the engine process. This is better than deleting the deployment pod since it won't give the pod a new SHA. This causes at least 2 restarts. 1 restart after the process exits and another restart when the new process gets evicted. There can be as many as 4 restarts before the system stabilizes.

Also refactor the UpgradeOptions as proposed in an earlier PR.

### DIFF
--- a/test/minikube/minikube_upgrade_helm_test.go
+++ b/test/minikube/minikube_upgrade_helm_test.go
@@ -15,11 +15,7 @@ import (
 	v12 "k8s.io/api/core/v1"
 )
 
-type UpdateOptions struct {
-	adminPodShouldGetRecreated            bool
-}
-
-func upgradeAdminTest(t *testing.T, fromHelmVersion string, updateOptions *UpdateOptions) {
+func upgradeAdminTest(t *testing.T, fromHelmVersion string, upgradeOptions *testlib.UpgradeOptions) {
 	testlib.AwaitTillerUp(t)
 	defer testlib.VerifyTeardown(t)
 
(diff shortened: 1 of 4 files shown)


test/minikube/minikube_upgrade_helm_test.go
```diff
@@ -123,38 +117,68 @@ func upgradeDatabaseTest(t *testing.T, fromHelmVersion string, updateOptions *Up
 	opt := testlib.GetExtractedOptions(options)
 
 	// make sure the DB is properly reconnected before restarting
-	testlib.Await(t, func() bool {
-		return testlib.GetStringOccurrenceInLog(t, namespaceName, admin0, "Reconnected with process with connectKey", &v12.PodLogOptions{}) == 2
+	err := testlib.AwaitWithError(t, func() bool {
```
sivanov-nuodb, line 120: Nitpick: What about `AwaitE`?

sivanov-nuodb (approved):
Thanks for enabling the tests again!

test/minikube/minikube_upgrade_helm_test.go
mkysel: will do

PR #242, opened 2021-08-25
Increase leaderAssignmentTimeout to account for CrashLoopBackOff

adriansuarez:
If SMs fail to start due to leaderAssignmentTimeout because some SM is not being scheduled (e.g. because the volume the archive is on is unavailable), then eventually the SMs will enter CrashLoopBackOff state and Kubernetes will begin scheduling them with a 5-minute delay. This would cause the leaderAssignmentTimeout problems to persist even after all SMs become schedulable, because now the SMs are more staggered in when they get scheduled, and we will be lucky for all of them to be scheduled within the same 60-second window as required by the current leaderAssignmentTimeout setting of 60000 milliseconds.

This change bumps the leaderAssignmentTimeout to 300000 milliseconds (5 minutes) to account for the maximum scheduling delay in Kubernetes.

### DIFF
--- a/stable/admin/values.yaml
+++ b/stable/admin/values.yaml
@@ -174,6 +174,9 @@ admin:
     # force reconnecting processes that are not in the domain state to shut
     # themselves down
     evictUnknownProcesses: true
+    # increase leader assignment timeout to account for maximum scheduling
+    # delay of 5 minutes in K8s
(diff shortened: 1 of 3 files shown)


sivanov-nuodb (approved):
Looks good to me.
Thanks for changing this!

bkelly-ndb (approved):

PR #243, opened 2021-09-27
Upgrade database protocol automatically

sivanov-nuodb:
**Changes**

Starting from NuoDB 4.2.3, the Kubernetes Aware Admin (KAA) can upgrade the database protocol version automatically (turned off by default). 
Expose this functionality in NuoDB Helm Charts so that NuoDB product upgrade process can be easily automated.

The database protocol version is eligible for an upgrade to the max possible version if all of the bellows are true:

- the database is in `RUNNING` state. This ensures that all SMs/archives are `RUNNING`.
- all database processes are `MONITORED`
- the automatic upgrade is enabled for this database (disabled by default)
- all database processes are running with the same release binary version
- there are available protocol versions for upgrading to

A TE to restart selection preference can be configured if needed.

**Testing**

- Regresion testing
- New integration tests
- Extend `TestKubernetesUpgradeFullDatabase` test and include conditional sub-test. It will run in the internal CI only until the feature is released.

### DIFF
--- a/stable/database/templates/_helpers.tpl
+++ b/stable/database/templates/_helpers.tpl
@@ -247,6 +247,23 @@ The configuration is imported only in the entrypoint cluster.
 {{- end -}}
 {{- end -}}
 
+{{/*
+Enables automatic database protocol upgrade via annotations.
+The configuration is imported only in the entrypoint cluster.
+*/}}
+{{- define "database.automaticProtocolUpgrade" -}}
+{{- if .Values.database.automaticProtocolUpgrade }}
+{{- if eq (include "defaultfalse" .Values.database.automaticProtocolUpgrade.enabled) "true" -}}
+{{- if (eq (default "cluster0" .Values.cloud.cluster.name) (default "cluster0" .Values.cloud.cluster.entrypointName)) }}
+"nuodb.com/automatic-database-protocol-upgrade": "true"
+{{- with .Values.database.automaticProtocolUpgrade.tePreferenceQuery }}
+"nuodb.com/automatic-database-protocol-upgrade.te-preference-policy": {{ . | quote }}
+{{- end -}}
+{{- end -}}
+{{- end -}}
+{{- end -}}
+{{- end -}}
+
 {{- define "autoRestore.type" -}}
 {{- $valid := list "stream" "backupset" "" }}
 {{- if not (has .Values.database.autoRestore.type $valid) }}
--- a/stable/database/templates/_helpers.tpl
+++ b/stable/database/templates/_helpers.tpl
@@ -252,16 +252,16 @@ Enables automatic database protocol upgrade via annotations.
 The configuration is imported only in the entrypoint cluster.
 */}}
 {{- define "database.automaticProtocolUpgrade" -}}
-{{- if .Values.database.automaticProtocolUpgrade }}
-{{- if eq (include "defaultfalse" .Values.database.automaticProtocolUpgrade.enabled) "true" -}}
-{{- if (eq (default "cluster0" .Values.cloud.cluster.name) (default "cluster0" .Values.cloud.cluster.entrypointName)) }}
+  {{- if .Values.database.automaticProtocolUpgrade }}
+    {{- if eq (include "defaultfalse" .Values.database.automaticProtocolUpgrade.enabled) "true" -}}
+      {{- if (eq (default "cluster0" .Values.cloud.cluster.name) (default "cluster0" .Values.cloud.cluster.entrypointName)) }}
 "nuodb.com/automatic-database-protocol-upgrade": "true"
-{{- with .Values.database.automaticProtocolUpgrade.tePreferenceQuery }}
+        {{- with .Values.database.automaticProtocolUpgrade.tePreferenceQuery }}
 "nuodb.com/automatic-database-protocol-upgrade.te-preference-policy": {{ . | quote }}
-{{- end -}}
-{{- end -}}
-{{- end -}}
-{{- end -}}
+        {{- end -}}
+      {{- end -}}
+    {{- end -}}
+  {{- end -}}
 {{- end -}}
 
 {{- define "autoRestore.type" -}}
--- a/stable/database/values.yaml
+++ b/stable/database/values.yaml
@@ -402,22 +402,14 @@ database:
   isManualVolumeProvisioning: false
   isRestore: false
 
-  # NuoDB supports explicit database protocol version upgrade which is performed
-  # after upgrading the NuoDB image for all database processes. To simplify
-  # NuoDB version rollout, Kubernetes Aware Admin (KAA) is used to automatically
-  # upgrade database protocol and restart a Transaction Engine (TE) as an
-  # upgrade finalization step. For more information on how to perform the steps
-  # manually, please check Upgrading the Database Protocol
+  # A new version of NuoDB database software may also introduce a new version of
+  # the database protocol. NuoDB supports explicit database protocol version
+  # upgrade which is performed after upgrading the NuoDB image for all database
+  # processes. To simplify NuoDB version rollout, Kubernetes Aware Admin (KAA)
+  # is used to automatically upgrade database protocol and restart a Transaction
+  # Engine (TE) as an upgrade finalization step. For more information on how to
+  # perform the steps manually, please check Upgrading the Database Protocol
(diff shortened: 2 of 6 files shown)


stable/database/values.yaml
```diff
@@ -402,6 +402,37 @@ database:
...
+  # NuoDB supports explicit database protocol version upgrade which is performed
+  # after upgrading the NuoDB image for all database processes. To simplify
+  # NuoDB version rollout, Kubernetes Aware Admin (KAA) is used to automatically
+  # upgrade database protocol and restart a Transaction Engine (TE) as an
+  # upgrade finalization step. For more information on how to perform the steps
+  # manually, please check Upgrading the Database Protocol
+  # (https://doc.nuodb.com/nuodb/latest/deployment-models/physical-or-vmware-environments-with-nuodb-admin/installing-nuodb/upgrading-to-a-new-release/upgrading-the-database-protocol/).
+  #
+  # The database protocol version is eligible for an upgrade to the max possible
+  # version if all of the bellows are true:
```
acabrele, line 414: I don't think we mention anywhere in the repo about the upgrade process itself (as in helm upgrade...)? Should we at least point out the minimum engine numbers required for a rolling upgrade somewhere?  Perhaps it is a seperate task to create a HowTo.

stable/database/values.yaml
```diff
@@ -402,6 +402,37 @@ database:
...
+  # - the database is in RUNNING state. This ensures that all SMs/archives are
+  #   RUNNING.
+  # - all database processes are MONITORED
+  # - the automatic upgrade is enabled for this database (disabled by default)
+  # - all database processes are running with the same release binary version
+  # - there are available protocol versions for upgrading to
```
mkysel, line 420: is this an implementation detail of KAA? Should we stick this text in the NuoDB docs instead? It can change independently of the Helm repo and the commend can become misleading/wrong.

test/minikube/minikube_rolling_upgrade_test.go
```diff
@@ -232,6 +240,62 @@ func TestKubernetesUpgradeFullDatabaseMinorVersion(t *testing.T) {
 
 		verifyAllProcessesRunning(t, namespaceName, admin0, 2)
 	})
+
```
mkysel, line 243: the code on lines 239-241 assumes that time is linearly moving forward. With KAA upgrade enabled, the TE might get killed while AwaitDatabaseUp is still running. So it was running, but then isn't again. This means that a) that function might observe 2 restarts and hence take a longer time (which should be fine since we wait for 300s) and b) once it finally returns, the DB version is already upgraded.

This could cause subtle timing hazards in this test.

stable/database/templates/_helpers.tpl
```diff
@@ -247,6 +247,23 @@ The configuration is imported only in the entrypoint cluster.
...
+{{- if eq (include "defaultfalse" .Values.database.automaticProtocolUpgrade.enabled) "true" -}}
+{{- if (eq (default "cluster0" .Values.cloud.cluster.name) (default "cluster0" .Values.cloud.cluster.entrypointName)) }}
+"nuodb.com/automatic-database-protocol-upgrade": "true"
+{{- with .Values.database.automaticProtocolUpgrade.tePreferenceQuery }}
+"nuodb.com/automatic-database-protocol-upgrade.te-preference-policy": {{ . | quote }}
+{{- end -}}
+{{- end -}}
+{{- end -}}
+{{- end -}}
+{{- end -}}
```
adriansuarez, line 265: I don't know if Helm allows it, but if it does I think it would be better to use indentation on nested these control structures to make them readable, e.g.

```
{{- define "database.automaticProtocolUpgrade" -}}
  {{- if .Values.database.automaticProtocolUpgrade }}
    {{- if eq (include "defaultfalse" .Values.database.automaticProtocolUpgrade.enabled) "true" -}}
      {{- if (eq (default "cluster0" .Values.cloud.cluster.name) (default "cluster0" .Values.cloud.cluster.entrypointName)) }}
      "nuodb.com/automatic-database-protocol-upgrade": "true"
        {{- with .Values.database.automaticProtocolUpgrade.tePreferenceQuery }}
        "nuodb.com/automatic-database-protocol-upgrade.te-preference-policy": {{ . | quote }}
        {{- end -}}
      {{- end -}}
    {{- end -}}
  {{- end -}}
{{- end -}}
```

test/minikube/minikube_rolling_upgrade_test.go
```diff
@@ -232,6 +240,62 @@ func TestKubernetesUpgradeFullDatabaseMinorVersion(t *testing.T) {
...
+			// upgrade is performed
+			testlib.Await(t, func() bool {
+				return testlib.GetRegexOccurrenceInLog(t, namespaceName, admin0,
+					"Shutting down startId=[0-9]+ to finalize the database protocol upgrade",
+					&corev1.PodLogOptions{}) >= 1
+			}, 60*time.Second)
+			testlib.AwaitDatabaseUp(t, namespaceName, admin0, opt.DbName, 2)
+			output, err = testlib.RunSQL(t, namespaceName, admin0, opt.DbName,
+				"select case when version = GETEFFECTIVEPLATFORMVERSION() then 'YES' else 'NO' end as TE_RESTARTED_AFTER_MAJOR_VERSION_UPGRADE"+
+					" from system.versions where property = 'SYSTEM_TABLES_VERSION'")
```
adriansuarez, line 283: This SQL statement is enormous. Can we just use the expression `version = GETEFFECTIVEPLATFORMVERSION()` without the `case` statement and check for `TRUE`?

adriansuarez (approved):
Looks good.

stable/database/templates/_helpers.tpl
sivanov-nuodb: Yes, we can nest the control structures.
We should just make sure that the indentation for the rendered text is correct.  For example, the actual annotation should be with 0 indentation because the whole template is indented with 4 spaces later when called.
I will make the change.

stable/database/values.yaml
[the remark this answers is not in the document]
sivanov-nuodb: Thanks for the suggestion! I will remove this block and add a link to the product documentation (once created.)

test/minikube/minikube_rolling_upgrade_test.go
sivanov-nuodb: Okay, I'll simplify it.

test/minikube/minikube_rolling_upgrade_test.go
```diff
@@ -232,6 +240,62 @@ func TestKubernetesUpgradeFullDatabaseMinorVersion(t *testing.T) {
...
+			require.NoError(t, err, "error cheking if SQL layer version is upgraded")
+			require.Contains(t, output, "YES")
+
+			// verify that the container image ID is stored in KV store so that
+			// database version is not checked again for this container image ID
+			testlib.Await(t, func() bool {
```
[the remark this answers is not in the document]
sivanov-nuodb: It will be updated in the next resync interval after checking that there are no other available versions in the database version response and all processes are using the same release.

stable/database/values.yaml
sivanov-nuodb: Maybe now is a good time to document the upgrade process using Helm. I've created a separate DOC task for the _HowTo_ document.

stable/database/values.yaml
```diff
@@ -402,6 +402,37 @@ database:
...
+  #
+  # IMPORTANT: After the database protocol version has been upgraded, the NuoDB
```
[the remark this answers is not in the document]
sivanov-nuodb: Good point! I will add one sentence here as well.

test/minikube/minikube_rolling_upgrade_test.go
sivanov-nuodb: Good point!
Actually, there will be a problem if {{AwaitDatabaseUp}} returns before the TE is restarted and then {{verifyAllProcessesRunning}} may fail if the new TE doesn't go `MONITORED` in 30 sec.

I will enable the automatic upgrade in the sub-test and add await for the TE restart count to harden it.

mkysel: thats a really neat solution!

mkysel (approved):

PR #244, opened 2021-09-27
Disable Helm Upgrade Tests without license

mkysel:
NuoDB infrastructure (bamboo) does not use the license and license content. Disable the full DB helm upgrade tests in those cases.

### DIFF
--- a/test/minikube/minikube_upgrade_helm_test.go
+++ b/test/minikube/minikube_upgrade_helm_test.go
@@ -4,6 +4,7 @@ package minikube
 
 import (
 	"fmt"
+	"os"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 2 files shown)


test/minikube/minikube_upgrade_helm_test.go
```diff
@@ -62,6 +63,10 @@ func upgradeAdminTest(t *testing.T, fromHelmVersion string, upgradeOptions *test
 }
 
 func upgradeDatabaseTest(t *testing.T, fromHelmVersion string, upgradeOptions *testlib.UpgradeOptions) {
+	if os.Getenv("NUODB_LICENSE") == "ENTERPRISE" {
+		t.Skip("Can not test helm upgrade in this environment")
```
sivanov-nuodb, line 67: Can we include a comment or JIRA indicating what is the problem?
It's hard to track these down after some time.

sivanov-nuodb (approved):
Looks good.

test/minikube/minikube_upgrade_helm_test.go
mkysel: done

bkelly-ndb (approved):

PR #254, opened 2021-11-10
Support for multiple TE groups

sivanov-nuodb:
**Motivation**

To enable NuoDB database external access via NuoAdmin load balancer services, users should be able to configure Transaction Engine(TE) database processes separately so that a subset of TEs can be selected for a specific SQL workload. An example configuration per TE is but is not limited to:

- different process labels
- different `mem` setting
- different affinity rules

To support this use case, users can now provision multiple sub-set of TEs called `TE groups`. Each TE group is represented by separate Kubernetes Deployment and has the same configuration. 

Instead of supporting multiple TE groups in one `database` Helm release,  we allow installing several `database` Helm releases for the same database name.  One primary and multiple secondary `database` Helm releases for the same database name can be deployed into the same Kubernetes namespace. When installing secondary `database` Helm release to add additional TE group, the `database.primaryRelease` option should be set to `false`.

To enable external access for each TE group separately, service of type `LoadBalancer` or `NodePort` is provisioned per TE group. NuoDB 4.2.4+ can automatically fetch the ingress address of the balancer service and configure external access for the TE database process. To enable this feature the `database.te.otherOptions.enable-external-access` options should be set to `true`.

The flexible configurations of TE groups allow SQL applications to leverage NuoDB Admin load balancer and select a unique TE group for different SQL workflows. By default, the NuoDB Admin will advertize the external TE address (cloud network load balancer address). To allow SQL applications deployed inside the cluster to use the internal TE address `PreferInternalAddress=true` connection property should be defined.

An example diagram showing a database deployment with two TE groups can be seen below:

database-groups

**Changes**

- added support for primary and secondary releases for the same database
- fixed process `otherOptions` so that they also support flags i.e `--resolve-hostname` or `--enable-external-access`
- render KAA annotations on the primary TE group only
- change the TE balancer service to select TEs from the local TE group only
- add support for `NodePort` TE service

**Testing**

- regression testing
- fixed the names of the TE log PVCs in the tests
- fixed the namespaces names so that they start with the common prefix
- added template testing for `otherOptions`
- added function test for multiple TE groups with external access (requires NuoDB 4.2.4)
- added the needed plumbing in CircleCI for testing external access with minikube

**Demo**

- External Access in GKE (happy path) - https://asciinema.org/a/ZIf7xBv9P48Iku09UNbu9nR3D.
- Extra (logging + internal access) - https://asciinema.org/a/LeW20sCbqntxBEufzgu3dMVAx


stable/database/templates/service-clusterip.yaml
```diff
@@ -13,6 +14,33 @@ metadata:
     chart: {{ template "database.chart" . }}
     release: {{ .Release.Name | quote }}
   name: {{ .Values.database.name }}-{{ default .Values.admin.serviceSuffix.clusterip .Values.database.serviceSuffix.clusterip }}
+spec:
+  ports:
+  - { name: 48006-tcp,  port: 48006,  protocol: TCP,  targetPort: 48006 }
+  selector:
+    domain: {{ .Values.admin.domain }}
+    database: {{ .Values.database.name }}
+    component: te
```
adriansuarez, line 23: How do we prevent this Service from selecting the secondary releases? Is the release name present in the `database` label?

stable/database/templates/service-headless.yaml
```diff
@@ -1,4 +1,5 @@
 {{- if or (not (hasKey .Values.database.te.dbServices "enabled")) (eq (include "defaultfalse" .Values.database.te.dbServices.enabled) "true") }}
```
adriansuarez, line 1: Does this Service do anything? It is supposed to select TEs, but isn't the point of a Headless Service to allocate DNS entries, which is impossible with a Deployment?

stable/database/templates/service.yaml
```diff
@@ -4,13 +4,16 @@ kind: Service
...
-    {{- if and .Values.database.te.externalAccess.internalIP .Values.cloud.provider}}
+    {{- if and .Values.database.te.externalAccess.internalIP .Values.cloud.provider }}
+    {{- if eq (default "LoadBalancer" .Values.database.te.externalAccess.type) "LoadBalancer" }}
     {{- if eq .Values.cloud.provider "amazon" }}
     service.beta.kubernetes.io/aws-load-balancer-internal: 0.0.0.0/0
     {{- else if eq .Values.cloud.provider "azure" }}
     service.beta.kubernetes.io/azure-load-balancer-internal: "true"
     {{- else if eq .Values.cloud.provider "google" }}
     cloud.google.com/load-balancer-type: "Internal"
+    networking.gke.io/load-balancer-type: "Internal"
```
adriansuarez, line 15: Should these cloud-provider-specific annotations be suppressed for NodePort?

stable/database/templates/service-clusterip.yaml
sivanov-nuodb: This service (`demo-clusterip`) is selecting all TEs in a single cluster for database `demo`. This is for backwards compatibility.
I added another service (`group1-nuodb-cluster0-demo-database-clusterip`) which selects only TEs from a specific Helm release (TE group).

adriansuarez: It looks like the Service filters on `app=<database.fullname>` in the secondary release. But don't we also want to use that filter in the primary release, otherwise this Service will select the secondary TEs as well?

stable/database/templates/service-headless.yaml
sivanov-nuodb: Correct! There is no point in keeping this service. Maybe WeLab still use it to connect to the database using `direct=true`. @acabrele Do you know if they changed to using the clusterip service?

I wanted to remove it but it will be in a separate PR but I'm happy to hear other thoughts if it is used in any way.

stable/database/templates/service-clusterip.yaml
adriansuarez: So this is used by `direct=true` currently? Would it be breaking backwards compatibility to make this Service only select the TEs in the primary release? Maybe it would break what @acabrele is doing, but until we properly introduce this feature, we don't expect TEs to be spread across multiple Deployments (ignoring multi-cluster).

Which Service would select only the primary TEs?

sivanov-nuodb: Your comment is on the cluster service installed only by the primary release so I'm confused a bit.

By design the following services should be all services deployed for TE group `group1` (primary) and `group2` (secondary):
* `demo-clusterip` selects all TEs from all TE groups (this is kept for backward compatibility)
* `group1-nuodb-cluster0-demo-database-clusterip` selects only TEs for `group1`
* `group2-nuodb-cluster0-demo-database-clusterip` selects only TEs for `group2`

Please let me know if the implementation doesn't match this somewhere.

adriansuarez: Okay, sorry. The second Service in this file is unconditionally created, so there would be a service specifically for the primary.

stable/database/templates/service.yaml
adriansuarez: Sorry, missed that.

stable/database/README.md
```diff
@@ -251,8 +252,9 @@ The following tables list the configurable parameters of the `database` chart an
 | `sm.nodeSelector` | Node selector rules for NuoDB SM | `{}` |
 | `sm.tolerations` | Tolerations for NuoDB SM | `[]` |
 | `sm.readinessTimeoutSeconds` | SM readiness probe timeout, sometimes needs adjusting depending on environment and pod resources | `5` |
-| `te.externalAccess.enabled` | Whether to deploy a Layer 4 cloud load balancer service for the admin layer | `false` |
-| `te.externalAccess.internalIP` | Whether to use an internal (to the cloud) or external (public) IP address for the load balancer | `nil` |
+| `te.externalAccess.enabled` | Whether to deploy a Layer 4 service for the database | `false` |
+| `te.externalAccess.internalIP` | Whether to use an internal (to the cloud) or external (public) IP address for the load balancer. Only applies to external access of type `LoadBalancer` | `nil` |
+| `te.externalAccess.type` | The service type used to enable external database access. The supported types are `NodePort` and `LoadBalancer` (defaults to `LoadBalancer`) | `nil` |
```
adriansuarez, line 257: I don't see the corresponding change to `stable/admin/README.md` for `admin.externalAccess.type`.

test/minikube/minikube_external_access_test.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+	t.Logf("Running command: <%s>", cmd)
+	out, err := exec.Command("sh", "-c", cmd).Output()
+	if err != nil {
+		if exiterr, ok := err.(*exec.ExitError); ok {
+			// nuosql has exited with an non zero exit code; log command stderr
+			require.NoError(t, err, string(exiterr.Stderr))
+		}
+		require.NoError(t, err)
+	}
```
adriansuarez, line 77: I'm confused about this block. `err` is not `nil` but still not an error?

adriansuarez (approved):
Looks good to me. See my comment about adding the `admin.externalAccess.type` value to the admin README.

stable/database/README.md
sivanov-nuodb: This is a good point. I forgot to add it to the `admin` chart. I will do that.

test/minikube/minikube_external_access_test.go
sivanov-nuodb: If we get into the block then there is an error. I was just using `require.NoError()` to generate the usual message and fail the test. Is there a better way to do that? 
I see the `require.NoError()` internally calls `t.FailNow()` so an alternative would be something like:

```
t.Logf("nuosql command failed: %s", string(exiterr.Stderr))
t.FailNow()
```

stable/database/templates/service-headless.yaml
acabrele: This was in use from before we had a cluster service available.  The headless service can have issues with load balancing too as it is dependant on DNS config.  I support removing in another PR once confirming it is not used by anyone.

stable/database/templates/service.yaml
```diff
@@ -20,15 +23,15 @@ metadata:
     domain: {{ .Values.admin.domain }}
     chart: {{ template "database.chart" . }}
     release: {{ .Release.Name | quote }}
-  name: {{ .Values.database.name }}-{{ default .Values.admin.serviceSuffix.balancer .Values.database.serviceSuffix.balancer }}
+  name: {{ template "database.fullname" . }}-{{ default .Values.admin.serviceSuffix.balancer .Values.database.serviceSuffix.balancer }}
```
acabrele, line 26: Same comment about default service name when it is a nodeport

acabrele (approved):
Thanks Stanimir.  Look forward to storage group support :-)

test/minikube/minikube_external_access_test.go
adriansuarez: Okay, I was just wondering if the `err` check is ternary, because it could either be `nil` or some non-`nil` value that does not indicate an error, but that does not appear to be the case.

You are basically checking `err != nil` explicitly to guard the typecast, right? Maybe just add a comment to make it clear that this is structured like this to generate a better message.

NikTJ777: From my reading of this block:
* if there is an error (err != nil)
* then check to see if the err is a specific type of err (exec.ExitError)
* and if it _is_ an ExitError, then cause the stderr from that ExitErr to be output
* otherwise, just output the standard details from the err.

stable/admin/templates/service.yaml
```diff
@@ -28,7 +30,7 @@ spec:
   selector:
     app: {{ template "admin.fullname" . }}
   sessionAffinity: None
-  type: LoadBalancer
+  type: {{ default "LoadBalancer" .Values.admin.externalAccess.type }}
```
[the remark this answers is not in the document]
sivanov-nuodb: Good point! I will add another key in `admin.serviceSuffix` and `database.serviceSuffix` values to represent `NodePort`.

stable/database/templates/deployment.yaml
```diff
@@ -8,6 +8,9 @@ metadata:
     kubectl.kubernetes.io/default-logs-container: engine
 {{- include "database.loadBalancerConfig" . | indent 4 }}
 {{- include "database.automaticProtocolUpgrade" . | indent 4 }}
+    {{- if eq (include "defaulttrue" .Values.database.primaryRelease) "false" }}
+    nuodb.com/database-secondary-release: "true"
```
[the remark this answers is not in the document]
sivanov-nuodb: Adding `nuodb.com/database-secondary-release: false` instead of `nuodb.com/database-primary-release: true` is done mainly for backward compatibility for cases where you have a newer version of NuoDB with an older version of the Helm Charts. This way any _Deployment_ that doesn't have the `nuodb.com/database-secondary-release` is considered installed from the primary Helm release.

stable/database/templates/service.yaml
sivanov-nuodb: Will fix.

test/minikube/minikube_external_access_test.go
sivanov-nuodb: > You are basically checking err != nil explicitly to guard the typecast, right

Yes, this was my idea, however, when I tested in a bit, it looks like that the cast doesn't need this guard. It will just fail if the `err` is `nil`.  You can see the small example that I used in https://play.golang.org/p/siBFV37YkCw.
I will reformat the code so that it's more readable and simple:

```golang
	out, err := exec.Command("sh", "-c", cmd).Output()
	if exiterr, ok := err.(*exec.ExitError); ok {
		// nuosql has exited with an non zero exit code; generate better error
		// message by including the command stderr
		require.NoError(t, err, string(exiterr.Stderr))
	} else {
		require.NoError(t, err)
	}
```

PR #255, opened 2021-11-24
Increase leaderAssignmentTimeout default value

sivanov-nuodb:
**Issue**

In certain cases, databases with 3+ Storage Managers can fail to calculate leader assignment within the configured default timeout due to additional SM Pod startup delays and scheduling deviation between different SM pods.

**Changes**
- Increase the default value for `leaderAssignmentTimeout` to `900000` (15 min.) which should be more than enough for most databases

### DIFF
--- a/stable/admin/values.yaml
+++ b/stable/admin/values.yaml
@@ -175,9 +175,10 @@ admin:
     # force reconnecting processes that are not in the domain state to shut
     # themselves down
     evictUnknownProcesses: true
-    # increase leader assignment timeout to account for maximum scheduling
-    # delay of 5 minutes in K8s
-    leaderAssignmentTimeout: "300000"
+    # increase leader assignment timeout to account for maximum scheduling delay
+    # of 5 minutes in K8s; set the default value to 15 minutes (3x back-off
+    # delay) which should be enough for most databases
+    leaderAssignmentTimeout: "900000"
 
   ## nuodb-admin resource requests and limits
   ## Ref: http://kubernetes.io/docs/user-guide/compute-resources/


bkelly-ndb (approved):

adriansuarez:
I'm not sure if I understand the comment about 3x5 minutes for 3 SMs. If the maximum restart interval is 5 minutes, then shouldn't all SMs restart roughly within that window? I understand having some margin for error in case SMs take a long time to start up _after_ being scheduled, but I don't think the `leaderAssignmentTimeout` scales with the maximum restart interval linearly in the way that you describe.

Consider the following picture, which ignores variability in SM start up itself (time is advancing from left to right and an asterisk is the SM being scheduled):

```
    0          5          15
SM0 *          *          *
SM1    *          *          *
SM2          *          *          *
```

Regardless of how they are staggered, all of the SMs should be scheduled within the same 5 minute window, right?

Like I said, I do agree that we should increase the value to account for variability in SM start up. But I'm not sure I agree with the comment about how the value is chosen, and I suspect that the 3x5 minutes value is more than enough even if the number of SMs is larger than 3.

adriansuarez (approved):
Looks good, but I think the comment might need revising.

sivanov-nuodb:
> If the maximum restart interval is 5 minutes, then shouldn't all SMs restart roughly within that window?

@adriansuarez After containers in a Pod exit, the kubelet restarts them with an exponential back-off delay. The maximum container back-off delay is 5 min. For example, let's suppose that one of the SMs can't be scheduled due to node unavailability. This will force the back-off delay to be capped at 5 min. for the other two SMs. 

My understanding is that those SM containers will run for the configured `leaderAssignmentTimeout` as the admin counts the timeout separately for each SM and then kubelet will delay the next scheduling for 5 min. So in theory we can have something like this for `leaderAssignmentTimeout=300000`.

```
    0          5         15         20
SM0 |++++++++++|----------|++++++++++|
SM1 ---------|++++++++++|----------|++
SM2 +++|----------|++++++++++|--------
```

* the `+` indicates that the SM container is running and blocked by the admin in STARTED state
* the `-` indicates that the SM container is not scheduled due to a back-off delay

In the timing above, we don't have a timeslot where all 3 SMs are running. Maybe in practice, such bad timing is less likely and SM2 will start with a smaller back-off delay.

adriansuarez:
@sivanov-nuodb: Good point about all SMs not necessarily being running at the same time due to them timing out at different times. The fact that all the SMs may not be running at the same time does not actually prevent leader assignment from taking place. If an SM comes up and supplies its archive history to the admin and then times out, leader assignment can still be performed in its absence. All SMs do not have to be running at the same time for leader assignment to be performed, the admin layer just needs to have an up-to-date archive history from all of them.

The problem with leader assignment being performed in the absence of some SMs is that the leader of the unpartitioned storage group can actually be an SM that is not currently running. In that case, all of the other SMs will be blocked waiting for the unpartitioned leader to come, which may take a long time if it is in CrashLoopBackOff, and then when they timeout and shutdown, they have to supply archive histories the next time they are start up and a new round of leader assignment has to take place. I think that situation becomes less likely as the number of SMs grows, since it becomes less likely that the SM that is not running is the unpartitioned leader.

I think we should just set `leaderAssignmentTimeout` to a very high value. The one you've chosen is probably good, but I actually can't think of any reason not to set it even higher.

sivanov-nuodb:
Thanks @adriansuarez for pointing out that the recent archive histories are "remembered" so that leader assignments can succeed event with an absence of an SM. This will also reduce significantly the database cold start time in such cases. 
I agree that the timeout doesn't scale linearly with the number of SMs and there are also other factors here. I will leave the value of 15 min. and update the commit message to reflect your comments above.

PR #256, opened 2021-11-25
Resurrect archive only if it will be reused

sivanov-nuodb:
**Issue**

In rare cases, a database running NuoDB 4.1.x can have `AWAITING_ARCHIVE_HISTORIES_INC` state if the PVC for some of the SMs needs to be recreated. For example:

- scale down SM StatefulSet to 0
- delete and recreate the archive PVC which creates a new PV (change in the storage class for example)
- scale up SM StatefulSet to the previous replica count

**Changes**

Resurrect the archive metadata which was previously bound to an SM Pod only if it will be reused. Use the previously fetched `archiveId` when resurrecting it. 

**Testing**

- `TestChangingJournalLocationWithMultipleSMs` was flaky because of this issue
- `TestDomainResync` exercise the archive metadata reuse after SM scale down and then up
- regression testing

### DIFF
--- a/stable/database/files/nuosm
+++ b/stable/database/files/nuosm
@@ -145,12 +145,15 @@ function isRestoreSourceAvailable() {
 #
 function resurrectRemovedArchive() {
   trace "resurrecting removed archive metadata"
-  # Find archiveId associated with process which address is the same as this pod hostname
-  # It assumes that there is only one archive object associated with the current host
-  removed_archive=$( nuocmd show archives --db-name $DB_NAME --removed --removed-archive-format "archive-id: {id}" | sed -En "/^archive-id: / {N; /$HOSTNAME/ s/^archive-id: ([0-9]+).*$/\1/; T; p}" | head -n 1 )
-  if [ -n "$removed_archive" ]; then
-    log "Resurrecting removed archiveId=${removed_archive}"
-    nuocmd create archive --db-name $DB_NAME --archive-path $DB_DIR --is-external --restored --archive-id $removed_archive
+  # Resurrect the archiveId associated which this pod if the previously fetched
+  # archiveId is removed
+  if [ -n "$myArchive" ] && \
+    nuocmd show archives \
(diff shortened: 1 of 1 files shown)


stable/database/files/nuosm
```diff
@@ -761,9 +761,10 @@ else
 
   wotIdid=""
 
-  # resurrect my archive if it is removed
-  # this will likely never execute but is kept to be sure
-  resurrectRemovedArchive
+  # resurrect my archive only if info.json file is available, otherwise
+  # nuodocker will create a new one; this is needed because KAA will remove
+  # archives for scaled down SM statefulsets
+  [ -f "$DB_DIR/info.json" ] && resurrectRemovedArchive
```
adriansuarez, line 767: I just looked at the code for `resurrectRemovedArchive` and it is doing screen-scraping of `nuocmd show archives` and filtering on the hostname to obtain the archive ID, which is readily available in `info.json`. I think we should just update `resurrectRemovedArchive` to get the archive ID from `info.json`.

adriansuarez (approved):
I'll approve this since it is an improvement over what we were doing, but we should get the archive ID from `info.json` and not by scraping `nuocmd show archives` output using an unreadable bash one-liner.

stable/database/files/nuosm
sivanov-nuodb: Good call! Yes, we need to fix the `resurrectRemovedArchive` so that it uses the already available `$myArchive` variable instead of fetching the re-fetching the archived. I added another commit to address that.

@acabrele reminded me that this "screen-scraping" won't work if the SM pod was previously in _crashloop_ state having some startup problem causing the admin to display `<UNKNOWN ADDRESS>` since the engine never passed `CONFIGURED` state. This won't happen when NuoDB 4.2+ is used, however, this change in `nuosm` should fix it when using previous versions of the product.

adriansuarez (approved):

bkelly-ndb (approved):

PR #257, opened 2021-11-26
Update AWS annotations for the LoadBalancer service

sivanov-nuodb:
**Issue**

Starting from aws-load-balancer-controller v2.2.0, the `service.beta.kubernetes.io/aws-load-balancer-internal` LoadBalancer service annotation is deprecated. The new `service.beta.kubernetes.io/aws-load-balancer-scheme` annotation should instead. The v2.2.0+ controller creates NLB with "internal" scheme by default which allows traffic from the VPC only. 

This prevents external access to the NuoDB domain when `admin.externalAccess.internalIP` and `database.te.externalAccess.internalIP` are set to `false`.

**Changes**

- added the needed new [AWS annotations](https://docs.aws.amazon.com/eks/latest/userguide/network-load-balancing.html#network-load-balancer) for external access to the `admin` and `database` _LoadBalancer_ services
- refactored the external service annotations rendering in named template
- added `admin.externalAccess.annotations` and `database.te.externalAccess.annotations` support for user-defined annotations. This is to allow custom configuration based on the load balancer controller version if needed.

**Testing**

- added more integration tests
- regression testing
- @bkelly-ndb could you please give it another run in EKS whenever possible? Thanks!

### DIFF
--- a/stable/admin/templates/_helpers.tpl
+++ b/stable/admin/templates/_helpers.tpl
@@ -266,3 +266,36 @@ Renders the admin service name for external access based on the service type
 {{ .Values.admin.domain }}
   {{- end }}
 {{- end }}
+
+{{/*
+Renders the annotations for the LoadBalancer admin service
+*/}}
+{{- define "admin.externalAccessAnnotations" -}}
+  {{- if eq (default "LoadBalancer" .Values.admin.externalAccess.type) "LoadBalancer" }}
+    {{- if .Values.admin.externalAccess.annotations }}
+{{ toYaml .Values.admin.externalAccess.annotations | trim }}
+    {{- else -}}
+      {{- if .Values.cloud.provider }}
+        {{- if eq .Values.cloud.provider "amazon" }}
+          {{- if .Values.admin.externalAccess.internalIP }}
+service.beta.kubernetes.io/aws-load-balancer-internal: "true"
+service.beta.kubernetes.io/aws-load-balancer-scheme: "internal"
+          {{- else }}
+service.beta.kubernetes.io/aws-load-balancer-type: "external"
+service.beta.kubernetes.io/aws-load-balancer-nlb-target-type: "ip"
+service.beta.kubernetes.io/aws-load-balancer-scheme: "internet-facing"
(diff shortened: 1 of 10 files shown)


stable/admin/templates/_helpers.tpl
```diff
@@ -265,4 +265,37 @@ Renders the admin service name for external access based on the service type
...
+{{ toYaml .Values.admin.externalAccess.annotations | trim }}
+    {{- else -}}
+      {{- if .Values.cloud.provider }}
+        {{- if eq .Values.cloud.provider "amazon" }}
+          {{- if .Values.admin.externalAccess.internalIP }}
+service.beta.kubernetes.io/aws-load-balancer-internal: "true"
+service.beta.kubernetes.io/aws-load-balancer-scheme: "internal"
+          {{- else }}
+service.beta.kubernetes.io/aws-load-balancer-type: "external"
+service.beta.kubernetes.io/aws-load-balancer-nlb-target-type: "ip"
```
sivanov-nuodb, line 285: This will work with the [Amazon EKS VPC CNI plugin](https://docs.aws.amazon.com/eks/latest/userguide/pod-networking.html) only. For alternative CNI plugins, it should be set to `instance`. For example:

```yaml
admin:
  externalAccess:
    enabled: true
    annotations:
      "service.beta.kubernetes.io/aws-load-balancer-type": "external"
      "service.beta.kubernetes.io/aws-load-balancer-nlb-target-type": "instance"
      "service.beta.kubernetes.io/aws-load-balancer-scheme": "internet-facing"
```

acabrele (approved):
I guess it would also be possible, with these changes, to deprecate externalAccess.internalIP and just allow the user to supply externalAccess.annotations whether it be for internal or external, and regardless of the cloud.provider.  But, good to keep for backward compatibility.

sivanov-nuodb:
> deprecate externalAccess.internalIP

Good point! Maybe we could deprecate it in v3.5.0 assuming that this change goes into v3.4.0.

adriansuarez (approved):
Looks good to me.

PR #260, opened 2021-12-03
Avoid recursive chmod if all directories already have 770

adriansuarez:
Avoid performing recursive on `/var/opt/nuodb` and `/var/log/nuodb` if all files and subdirectories one level deep already have mode 770. For SMs, descend two levels to ensure that all files `/var/opt/nuodb/archive` and `/var/opt/nuodb/journal` have mode 770 (this is to account for the fact that previously we did not perform recursive chmod at all, so it is possible for those directories to contain files that are not writeable by the nuodb user).

### DIFF
--- a/stable/database/templates/statefulset.yaml
+++ b/stable/database/templates/statefulset.yaml
@@ -437,16 +437,9 @@ spec:
       - name: init-disk
         image: {{ template "init.image" . }}
         imagePullPolicy: {{ default "" .Values.busybox.image.pullPolicy | quote }}
-        command:
-          - 'chmod'
-          - '-R'
-          - '770'
-          - '/var/opt/nuodb/archive'
-          - '/var/opt/nuodb/backup'
-          - '/var/log/nuodb'
-{{- if eq (include "defaultfalse" .Values.database.sm.hotCopy.journalPath.enabled) "true"}}
-          - '/var/opt/nuodb/journal'
-{{- end }}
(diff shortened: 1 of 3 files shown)


stable/database/templates/statefulset.yaml
```diff
@@ -68,15 +68,9 @@ spec:
...
-        command:
-          - 'chmod'
-          - '-R'
-          - '770'
-          - '/var/opt/nuodb/archive'
-          - '/var/log/nuodb'
-  {{- if eq (include "defaultfalse" .Values.database.sm.noHotCopy.journalPath.enabled) "true"}}
-          - '/var/opt/nuodb/journal'
-  {{- end }}
+        command: ['sh', '-c', '
```
bkelly-ndb, line 71: Same change needs to be applied on the hotcopy init-disk container below

bkelly-ndb (changes requested):

stable/database/templates/statefulset.yaml
sivanov-nuodb: +1

sivanov-nuodb (approved):
Thanks for adding this improvement!

stable/database/templates/statefulset.yaml
adriansuarez: Good catch.

bkelly-ndb (approved):
Looks good, thanks for fixing so quick!

PR #262, opened 2021-12-20
Request incremental hot copy if journal hot copy is temporary disabled

sivanov-nuodb:
```
NuoDB 4.3+ will temporary suspend the journal hot copy during archive sync.
When coordinated hot copy operation is performed, the error is shown either in
the "message" attribute of the hot copy response or when the hot copy is
initiated depending on the choice of the coordinator SM

For example:
- "message": "Hot copy failed: Node 6: Journal hot copy has been suspended due
  to archive sync. Full or incremental hot copy must be completed before
 journal hot copy."
- 'backup database' failed: Failure while performing hot-copy: Error while
  sending engine request: Journal hot copy has been suspended due to archive
  sync. Full or incremental hot copy must be completed before journal hot copy.

If full or incremental hot copy is already requested, the error message will be
different: "Journal hot copy has been suspended due to archive sync. The
running incremental hot copy must complete before journal hot copy can start."

A new incremental hot copy is requested now if the journal hot copy fails
because it is temporary disabled.
```


stable/database/files/nuobackup
```diff
@@ -87,8 +87,18 @@ function execute_backup() {
...
                --value "${failed} ${backupset}" --expected-value "${failed}"
          fi
-      fi 
-      exit $retval
+      fi
+
+      if [ "$type" = "journal" ] && echo "$output" | grep -q "Full or incremental hot copy must be completed before journal hot copy"; then
+         echo "Executing incremental hot copy as journal hot copy is temporarily suspended"
+         execute_backup "incremental" || return $?
+         # retry the journal backup which should suceed now
```
adriansuarez, line 95: Typo: suceed->succeed

stable/database/files/nuobackup
```diff
@@ -87,8 +87,18 @@ function execute_backup() {
...
+         execute_backup "journal"
```
adriansuarez, line 96: This function is executing itself. Is there a risk that we keep issuing journal backups forever because `execute_backup "incremental"` did not return the correct code to signal that it failed? I would set a variable to prevent more than one "nested" invocation of `execute_backup "journal"` being generated.

adriansuarez (approved):
I had only a couple of comments. I think we should be more defensive about the recursive invocation of `execute_backup` because shell scripting brittle.

stable/database/files/nuobackup
sivanov-nuodb: Goot catch!

adriansuarez (approved):
Looks good.

stable/database/files/nuobackup
```diff
@@ -87,8 +89,17 @@ function execute_backup() {
                --key "$NUODB_BACKUP_KEY/$db_name/${backup_group}/failed" \
                --value "${failed} ${backupset}" --expected-value "${failed}"
          fi
-      fi 
-      exit $retval
+      fi
+      if [ -n "$retry_failed" ] && [ "$type" = "journal" ] && echo "$output" | grep -q "Full or incremental hot copy must be completed before journal hot copy"; then
```
acabrele, line 93: I think the portion of the message being handled is a little misleading when seen in isolation, but I understand why you are not handling only the text about journal being suspended (because the same text appears in the other message).  Instead, can we just put comment here that this is related to the journal hotcopy suspension?

acabrele (approved):

bkelly-ndb (approved):

stable/database/files/nuobackup
sivanov-nuodb: Will do.

PR #263, opened 2022-01-04
Allow Pod security context to be specified

adriansuarez:
- Allow security context to be specified for all Pods created by NuoDB.
  `database.securityContext.enabled` and `admin.securityContext.enabled`
  allow the `runAsUser` and `fsGroup` values to be specified, while
  explicitly supplying a `runAsGroup` value of 0.
  `*.securityContext.runAsNonRootGroup` causes the `runAsUser` and
  `runAsGroup` values for the security context to each be set to 1000.
  This will be supported by the NuoDB image version 4.3. All versions of
  the NuoDB image before 4.3 require the user to have gid 0.
  `*.securityContext.fsGroupOnly` causes a security context to be
  created that only specifies the `fsGroup`.
- Add `admin.initContainers.runInitDisk` and
  `database.initContainers.runInitDisk` values to allow the init
  container that changes permissions and ownership of files within
  storage volumes to be omitted. If a security context is specified,
  this init container is usually redundant, because `fsGroup` implements
  the same behavior natively to Kubernetes. There are some
  storage-classes that do not support this behavior of `fsGroup` such as
  hostpath, which is used by Minikube.  In that case, it makes sense to
  keep the init container enabled even if a security context is
  specified. Since the init container cannot actually change permissions
  and ownership unless it runs as root, it now specifies its own
  security context by default to override the Pod security context. The
  init container security context can be omitted by specifying
  `*.initContainers.runInitDiskAsRoot=false`.
- Make init container compatible with a user with a non-0 gid by
  changing ownership to 1000. Only uid 1000 can have a non-0 gid, and
  the files are also made group-accessible, so that any user with gid 0
  can access them as well. This also fixes the command so that it adds
  the required permissions rather than explicitly setting them to 770,
  which may actually remove permissions, e.g. if the file or directory
  is world-accessible.
- Improve log collection on test failure so that if the default
  container ("admin" or "engine") is not running, we fetch the output of
  its last invocation, and if it does not have a last invocation because
  the Pod failed in the init container, we fetch the output of the init
  container.
- Improve logging of scheduled Pods to show the expected number and the
  names of the scheduled Pods.
- Enable TestUpgradeHelmFullDB in Bamboo, which does not seem to be
  failing anymore after these changes.

### DIFF
--- a/stable/admin/README.md
+++ b/stable/admin/README.md
@@ -170,8 +170,15 @@ The following tables list the configurable parameters for the `admin` option of 
 | `logPersistence.size` | Amount of disk space allocated for log storage | `60Gi` |
 | `logPersistence.storageClass` | Storage class for volume backing log storage | `-` |
 | `envFrom` | Import ENV vars from one or more configMaps | `[]` |
-| `options` | Set optons to be passed to nuoadmin as arguments | `{}` |
-| `securityContext.capabilities` | add capabilities to the container | `[]` |
+| `options` | Set options to be passed to nuoadmin as arguments | `{}` |
+| `initContainers.runInitDisk` | Whether to run the `init-disk` init container to set volume permissions | `true` |
+| `initContainers.runInitDiskAsRoot` | Whether to run the `init-disk` init container as root | `true` |
+| `securityContext.fsGroupOnly` | Creates a security context for Pods containing only the `securityContext.fsGroup` value | `false` |
+| `securityContext.runAsNonRootGroup` | Creates a security context for Pods containing a non-root user and group (1000:1000) along with the `securityContext.fsGroup` value | `false` |
+| `securityContext.enabled` | Creates a security context for Pods containing the `securityContext.runAsUser` and `securityContext.fsGroup` values | `false` |
+| `securityContext.runAsUser` | The user ID for the Pod security context created if `securityContext.enabled` is `true`. | `1000` |
+| `securityContext.fsGroup` | The `fsGroup` for the Pod security context created if any of `securityContext.fsGroupOnly`, `securityContext.runAsNonRootGroup`, or `securityContext.enabled` are `true`. | `1000` |
+| `securityContext.capabilities` | Capabilities to add to admin container security context | `[]` |
 | `tlsCACert.secret` | TLS CA certificate secret name | `nil` |
 | `tlsCACert.key` | TLS CA certificate secret key | `nil` |
 | `tlsKeyStore.secret` | TLS keystore secret name | `nil` |
--- a/stable/restore/templates/_helpers.tpl
+++ b/stable/restore/templates/_helpers.tpl
@@ -127,6 +127,76 @@ imagePullSecrets:
 {{- end -}}
 
 {{/*
+Get Pod securityContext
+*/}}
+{{- define "securityContext" -}}
+{{- if eq (include "defaultfalse" .Values.database.securityContext.enabled) "true" }}
+securityContext:
+  runAsUser: {{ default 1000 .Values.database.securityContext.runAsUser }}
+  runAsGroup: 0
+  fsGroup: {{ default 1000 .Values.database.securityContext.fsGroup }}
+  {{- include "sc.fsGroupChangePolicy" . }}
+{{- else if eq (include "defaultfalse" .Values.database.securityContext.runAsNonRootGroup) "true" }}
+securityContext:
+  runAsUser: 1000
+  runAsGroup: 1000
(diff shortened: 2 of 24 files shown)


stable/restore/templates/_helpers.tpl
```diff
@@ -126,6 +126,76 @@ imagePullSecrets:
...
+  {{- include "sc.fsGroupChangePolicy" . }}
+{{- end }}
+{{- end -}}
+
+{{/*
+Get fsGroupChangePolicy if Kubernetes version supports it
+*/}}
+{{- define "sc.fsGroupChangePolicy" -}}
+{{- if semverCompare ">=1.20" .Capabilities.KubeVersion.Version }}
+  fsGroupChangePolicy: OnRootMismatch
```
sivanov-nuodb, line 157: It looks like the default value for this is `Always`. Are we specifying this to overcome the slow `chmod -R` for the archive volume?

stable/admin/README.md
```diff
@@ -170,8 +170,15 @@ The following tables list the configurable parameters for the `admin` option of
...
 | `logPersistence.storageClass` | Storage class for volume backing log storage | `-` |
 | `envFrom` | Import ENV vars from one or more configMaps | `[]` |
-| `options` | Set optons to be passed to nuoadmin as arguments | `{}` |
-| `securityContext.capabilities` | add capabilities to the container | `[]` |
+| `options` | Set options to be passed to nuoadmin as arguments | `{}` |
+| `initContainers.runInitDisk` | Whether to run the `init-disk` init container to set volume permissions | `true` |
+| `initContainers.runInitDiskAsRoot` | Whether to run the `init-disk` init container as root | `true` |
+| `securityContext.fsGroupOnly` | Creates a security context for Pods containing only the `securityContext.fsGroup` value | `false` |
+| `securityContext.runAsNonRootGroup` | Creates a security context for Pods containing a non-root user and group (1000:1000) along with the `securityContext.fsGroup` value | `false` |
+| `securityContext.enabled` | Creates a security context for Pods containing the `securityContext.runAsUser` and `securityContext.fsGroup` values | `false` |
```
sivanov-nuodb, line 178: My understanding is that `securityContext.enabled`, `securityContext.fsGroupOnly` and `securityContext.runAsNonRootGroup` are mutual exclusive options. Would it be better if we have something like below?

```yaml
securityContext:
  # disable securityContext
  enabled: false
  # One of runAsRootGroup|runAsNonRootGroup|fsGroupOnly
  runAsPreset: "runAsRootGroup"
  runAsUser: 1000
  runAsGroup: 0
  fsGroup: 1000
  ...
```

sivanov-nuodb (approved):
@adriansuarez Thanks for allowing NuoDB to run with non-root GID!

stable/restore/templates/_helpers.tpl
adriansuarez: Yes, this is supposed to avoid the issue in `fsGroup` analogous to what @bkelly-ndb discovered with the init container. The reason why it is conditionalized on version is that it does not exist before Kubernetes 1.18 and is in alpha before 1.20.

stable/admin/README.md
adriansuarez: The reason why I went with this approach is that it I didn't see any pattern for validating options for cases where one of the well-known constants is misspelled. I guess the user could also misspell the key, e.g. `--set admin.securityContext.runAsNonRootGropu=true`, and that would also silently fail because the values file does not have a schema (that I'm aware of).

I still have a slight preference for this approach because it allows the user to enable the different modes by setting only one value, assuming that the values file does not set any of these values to `true`. For example, to enable `fsGroup`, the user can do

```
... --set admin.securityContext.fsGroupOnly=true
```

as opposed to

```
... --set admin.securityContext.enabled=true --set admin.securityContext.runAsPreset=fsGroupOnly
```

sivanov-nuodb: Works for me.

> I didn't see any pattern for validating options for cases where one of the well-known constants is misspelled

I don't think that Helm can do schema validation. We have poor man's logic for validating `autoRestore.source`:

```
{{- define "autoRestore.source" -}}
{{- if hasPrefix ":" .Values.database.autoRestore.source }}
{{- $valid := list ":latest" ":group-latest" "" }}
{{- if not (has .Values.database.autoRestore.source $valid) }}
{{- fail (printf "Invalid autorestore source: %s" .Values.database.autoRestore.source) }}
{{- end }}
{{- end -}}
{{- with .Values.database.autoRestore.source }}
{{- . }}
{{- end -}}
{{- end -}}
```

bkelly-ndb (approved):

PR #265, opened 2022-01-07
Backup groups enhancements

sivanov-nuodb:
**Changes**

- automatically create backup groups based on `database.sm.hotCopy.replicas` with one HCSM in each group
- support for user-defined backup groups using labels and schedule overwrites 
- fail database restore early if a _special_ restore source can't be resolved
- the `latest` value in the KV store is updated with the corresponding backup group name on _any_ successful hot copy
- store all pod names that participated in a successful _full_ hot copy in the KV store
- the `backup` label on the HCSMs is not removed for backward compatibility but it's is not used anymore

**Enhancements**

- new `database.sm.hotCopy.backupGroupPrefix` option
- new special values can be used as a `restore.source` - `<backup group>:latest` and `<backup group>:<index>`
- new `database.sm.hotCopy.journalBackup.journalSchedule` option
- new `nuobackup --type report-groups` option
- new `nuobackup --type report-backups` option

**UX changes**

- now `restore.labels` or `restore.archive-ids` should be used when a specific backup set is configured as a `restore.source` because not all HCSMs are part of the same backup group by default anymore
- `restore.labels` default value is changed from `{}` to `""`. Providing _map_ is still supported for backward compatibility
- change backup cronjob names to include the backup group name
- scaling up HCSM statefulset should be done using `helm upgrade` so that the cronjobs for the new backup groups are created

**Removed**

- deprecate `:group-latest` special value for `restore.source` option
- deprecate `stream:http://...` special value for `restore.source` option
- remove `nuobackup --semaphore` flag
- remove `nuorestore --semaphore` flag
- remove `database.sm.hotCopy.journalBackup.intervalMinutes` option
- remove `database.sm.hotCopy.backupGroup` option

**Examples**

Leaving the default value for `database.sm.hotCopy.backupGroups`, NuoDB Helm Charts will automatically create backup groups based on `database.sm.hotCopy.replicas` with one HCSM in each group. For example:

```shell
helm install database nuodb/database \
  --namespace nuodb \
  --set cloud.cluster.name=cluster0 \
  --set database.sm.hotCopy.replicas=2 \
  --set database.sm.noHotCopy.replicas=1 \
  --set database.sm.hotCopy.fullSchedule='?/3 * * * *' \
  --set database.sm.hotCopy.incrementalSchedule='?/2 * * * *' \
  --set database.sm.hotCopy.journalBackup.enabled=true \
  --set database.sm.hotCopy.journalBackup.journalSchedule='?/1 * * * *' 
```

```shell
$ kubectl get cronjobs.batch
NAME                                  SCHEDULE      SUSPEND   ACTIVE   LAST SCHEDULE   AGE
full-hotcopy-demo-cluster0-0          ?/3 * * * *   False     0        3m              10m
full-hotcopy-demo-cluster0-1          ?/3 * * * *   False     0        3m              10m
incremental-hotcopy-demo-cluster0-0   ?/2 * * * *   False     0        2m              10m
incremental-hotcopy-demo-cluster0-1   ?/2 * * * *   False     0        2m              10m
journal-hotcopy-demo-cluster0-0       ?/1 * * * *   False     0        60s             10m
journal-hotcopy-demo-cluster0-1       ?/1 * * * *   False     0        60s             10m
```

```shell
$ kubectl -ti sm-database-nuodb-cluster0-demo-hotcopy-0 -- bash

bash-4.4$ nuobackup --type report-groups
cluster0-1 sm-database-nuodb-cluster0-demo-hotcopy-1
cluster0-0 sm-database-nuodb-cluster0-demo-hotcopy-0

bash-4.4$ nuobackup --type report-backups
cluster0-1:1 20220106T113252 sm-database-nuodb-cluster0-demo-hotcopy-1
cluster0-1:2 20220106T113312 sm-database-nuodb-cluster0-demo-hotcopy-1
cluster0-1:3 20220106T113338 sm-database-nuodb-cluster0-demo-hotcopy-1
cluster0-0:1 20220106T113258 sm-database-nuodb-cluster0-demo-hotcopy-0
cluster0-0:2 20220106T113312 sm-database-nuodb-cluster0-demo-hotcopy-0
cluster0-0:3 20220106T113337 sm-database-nuodb-cluster0-demo-hotcopy-0
```

If there is a requirement for all the HCSMs in a single cluster to be part of the same backup group (as it is before Helm Charts v3.5.0), the following configuration should be used:

```shell
helm install database nuodb/database \
  --namespace nuodb \
  --set cloud.cluster.name=cluster0 \
  --set database.sm.hotCopy.replicas=2 \
  --set database.sm.noHotCopy.replicas=1 \
  --set database.sm.hotCopy.backupGroups.cluster0.labels='backup cluster0'
```

**Testing**

- new integration tests
- new E2E restore test with multiple automatically created backup groups
- `TestKubernetesRestoreWithStorageGroups ` has been modified so that all SMs are part of one backup group because they serve disjoint storage groups
- specify `restore.labels` for some of the restore tests that are using backup set as `restore.source`
- regression testing


stable/database/files/nuosm
```diff
@@ -395,34 +395,48 @@ function completeRestoreRequest(){
...
+  echo "${group:-UNKNOWN_BACKUP_GROUP}"
+}
+
 #=======================================
 # function - resolves :latest or :group-latest backup source and stores it
-# directly into restore_source variable
+# directly into restore_source variable. It is needed only when performing auto
+# restore as nuorestore script already resolved these
 #
 function resolveLatestSource() {
```
sivanov-nuodb, line 412: This function will be used now only when _autoRestore_ is performed.

stable/database/templates/statefulset.yaml
```diff
@@ -523,14 +523,13 @@ spec:
             secretKeyRef:
               name: {{ .Values.database.name }}.nuodb.com
               key: database-backup-credentials
-        - { name: NUODB_DOMAIN,        value: "{{ .Values.admin.domain }}" }
-        - { name: NUOCMD_API_SERVER,   value: "{{ template "admin.address" . }}:8888" }
-        - { name: PEER_ADDRESS,        value: "{{ template "admin.address" . }}" }
-        - { name: NUODB_BACKUP_GROUP,  value: "{{ include "hotcopy.group" . }}" }
```
sivanov-nuodb, line 529: The only change for this block of code is the removal of `NUODB_BACKUP_GROUP`

stable/database/templates/_helpers.tpl
```diff
@@ -185,12 +185,53 @@ Return options as $key $value
...
+Return labels for a specific backup group. If there is no user-defined backup
+groups, return the pod name of a single HCSM by extracting the pod ordinal from
+the automatically generated backup group name. Otherwise return the configured 
+backup group labels or empty value (representing all HCSMs in the database).
+*/}}
+{{- define "hotcopy.group.labels" -}}
+  {{- if .Values.database.sm.hotCopy.backupGroups -}}
+    {{- $group := index .Values.database.sm.hotCopy.backupGroups (print .backupGroup) -}}
+    {{- if $group -}}
+{{- default "" $group.labels -}}
```
adriansuarez, line 228: The indentation throws me off. This is just a value that is supposed to be expanded. Doesn't the `{{- ... -}}` cause the leading and trailing whitespace to be collapsed?

stable/database/templates/_helpers.tpl
```diff
@@ -185,12 +185,53 @@ Return options as $key $value
...
+    {{- end -}}
+  {{- else -}}
+    {{- $groupPrefix := include "hotcopy.groupPrefix" . -}}
+{{ printf "pod-name sm-%s-hotcopy%s" (include "database.fullname" .) (trimPrefix $groupPrefix .backupGroup) }}
```
adriansuarez, line 232: Same comment here. I think you should be able to indent normally and use `{{- printf ... }}`.

stable/restore/files/nuorestore
```diff
@@ -125,12 +144,60 @@ done
...
+   # "<backup_group>:<index>"; fail the database restore early if we are not
+   # able to do that
+   if echo "$restore_source" | grep -Eq ":latest$"; then
+      backup_group="${restore_source%:latest}"
+      if [ -z "$backup_group" ]; then
+         backup_group=$(nuocmd get value --key "$NUODB_BACKUP_KEY/$db_name/latest")
+      fi
+      backup_index=$(nuocmd get value --key "$NUODB_BACKUP_KEY/$db_name/$backup_group/latest")
+      if [ -z "$backup_index" ]; then
+         die 1 "Unable to find the latest backup for backup group '${backup_group}'"
```
adriansuarez, line 158: Single-quoting `'${backup_group}'` specifies it as a literal. Is that what you want?

adriansuarez: Same comment on lines 164 and 173.

stable/database/README.md
```diff
@@ -228,15 +228,20 @@ The following tables list the configurable parameters of the `database` chart an
...
 | `sm.hotCopy.failureHistory` | Number of failed jobs to keep | `5` |
 | `sm.hotCopy.backupDir` | Directory path where backupsets will be stored | `/var/opt/nuodb/backup` |
-| `sm.hotCopy.backupGroup` | Name of the backup group | `{{ .Values.cloud.cluster.name }}` |
-| `sm.hotCopy.fullSchedule` | cron schedule for FULL hotcopy jobs | `35 22 * * 6` |
-| `sm.hotCopy.incrementalSchedule` | cron schedule for INCREMENTAL hotcopy jobs | `35 22 * * 0-5` |
+| `sm.hotCopy.backupGroupPrefix` | Prefix for the automatically generated backup groups | `{{ .Values.cloud.cluster.name }}` |
+| `sm.hotCopy.backupGroups` | Backup groups configuration. By default a backup group per HCSM is created automatically | `{}` |
+| `sm.hotCopy.backupGroups.<name>.labels` | Space separated process labels used to select the Storage Managers which are part of this backup group. _Any_ label key and value should match for the SM to be selected | `nil` |
+| `sm.hotCopy.backupGroups.<name>.overwrites.fullSchedule` | cron schedule for _FULL_ hot copy performed by this backup group. If not defined the `sm.hotCopy.fullSchedule` setting will be used | `nil` |
```
adriansuarez, line 234: Can you replace `overwrites` with `overrides`? Or what about just `sm.hotCopy.backupGroups.<name>.fullSchedule`, `sm.hotCopy.backupGroups.<name>.incrementalSchedule`, etc. defining the override? I'm not sure if that introduces ambiguity, but it seems intuitive enough to me that I don't think we need to nest the value under `overwrites`/`overrides`.

adriansuarez (approved):
Looks good to me.

I have one comment about the `overwrites` vs `overrides` vs implicitly defining overrides by the presence of the backup schedule under the backup group name.

stable/database/files/nuobackup
```diff
@@ -43,7 +39,6 @@ backup_type="full"
 db_name=$DB_NAME
 backup_group=""
 labels=""
-semaphore=""
```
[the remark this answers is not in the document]
sivanov-nuodb: Well, I will update the commit message to reflect this.

stable/database/templates/_helpers.tpl
sivanov-nuodb: I think that in this case we can indent it properly because of the `{{-`. I will test and update it.

stable/restore/files/nuorestore
sivanov-nuodb: Actually, the shell with expanding the variable as the whole string is in double quotes. This is a quick test to demonstrate it.

```shell
bash-3.2$ backup_group=cluster0
bash-3.2$ die 1 "Unable to find the latest backup for backup group '${backup_group}'"
Unable to find the latest backup for backup group 'cluster0'
```

And here are some tests with the actual script:

```shell
$ k logs restore-demo-n2qff -f
Unable to connect to http://nuodb.nuodb.svc:8888: HTTPConnectionPool(host='nuodb.nuodb.svc', port=8888): Max retries exceeded with url: /api/1/kvstore/%2Fnuodb%2Fnuobackup%2Fdemo%2Ffoo%2F2 (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fbd50f56a10>: Failed to establish a new connection: [Errno -2] Name or service not known',))
2022-01-12T11:24:07.459+0000 Unable to find a valid backup set for restore source 'foo:2'
```

```shell
$ k logs restore-demo-6nrqq -f
2022-01-12T11:29:10.704+0000 Unable to find the latest backup for backup group 'foo'
```

stable/database/README.md
sivanov-nuodb: I agree with you. It would be easier if the structure is flat and remove `overwrites`.

sivanov-nuodb:
@adriansuarez I pushed some changes which should address your comments.

stable/restore/files/nuorestore
adriansuarez: I see. It seems like I never stop learning new and surprising things about shell scripting.

adriansuarez:
> @adriansuarez I pushed some changes which should address your comments.

Thanks for update. Looks good to me.

stable/database/values.yaml
```diff
@@ -323,7 +340,7 @@ database:
       ## settings for the journal backup
       journalBackup:
         enabled: false
-        intervalMinutes: 15
+        journalSchedule: "?/15 * * * *"
```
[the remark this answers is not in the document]
sivanov-nuodb: > Could you please explain the benefit of changing the journalBackup.intervalMinutes to a cron schedule?

This is done so that all backup schedule options are consistent and it's easy for us and for the users to spot them.
We already have _cron schedules_ exposed for other backup types so it is not something new that the user should take into consideration and learn how to define.  It also aligns better when used with named backup groups. I.e 

```
backupGroups:
  foo:
    labels: "..."
    fullSchedule: "00 03 * * 6"
    incrementalSchedule: "00 03 * * 0-5"
    journalSchedule: "?/5 * * * *"
```

I see that Brian also likes the change but if someone else thinks that this is not appropriate, I am fine with having a further discussion.

> I am a firm believer in giving users controls which are reasonably self-explanatory, and which prevent - as much as is reasonable - the users from setting an insane value.

According to me, the new option name of `journalSchedule` is as descriptive as the old one. I don't think that the option name of `intervalMinutes` prevents the user to configure "insane" values (you can still put 120 or more). Maybe 99% of the time there won't be a need to change the default value. If we want to make it clear that journal backup should be taken often then we need to document it properly in the `values.yaml` and the `README.md`. I've updated them so that it's more clear.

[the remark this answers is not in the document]
NikTJ777: The value of a cron schedule has 5 fields - of which one, and only one, (minutes) is valid for journal backups.

If the user sets a value _in any of the other 4 fields_ the result will most likely be crashing SMs.

If you give users fields to play with which are, in fact, illegal to use - they are going to ask why they were given that ability in the first place.

>I don't think that the option name of intervalMinutes prevents the user to configure "insane" values (you can still put 120 or more).
It is very easy to put a check into the helm charts to limit the values of intervalMinutes to be between 5 and some upper value. (what should that upper value be?)
It will be much more effort to the build a check that does the same for a cron schedule value - you would need to pull the cron schedule apart, and throw an error if any field other than minutes is set _and_ check that the minutes filed is within sane limits.

> This is done so that all backup schedule options are consistent
But it masks the very real fact that the appropriate values for journal backups are _NOT_ consistent with the other backups
There is no situation in which a schedule other than some multiple of minutes is reasonable or safe for the journal backup period.

Note the following inconsistencies of journal backups vs full or incremental backups:
* if an incremental backups fail then the database will continue running with no problems;
* if a full backup fails, then the database will continue running - unless that failure impacts journal backups;
* if a journal backup fails repeatedly, then the backup SMs will run out of journal disk space and crash;
   - this may bring the entire database down, or eliminate its HA component;
 * if a full or incremental backup is configured with an infrequent period, then the result is relatively few backups with large gaps between them;
 * if a journal backup is configured with an infrequent period - the backup SMs crash.

In short, I don't believe that consistency overrules practicality.
A consistently bad implementation is still a bad implementation.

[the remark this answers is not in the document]
NikTJ777: ... All that being said, I guess that the user-settable parameters of the helm charts become less of an issue as we move towards DBaaS...

PR #267, opened 2022-01-25
Deploy NuoDB in more restricted environments

sivanov-nuodb:
**Changes**

- include the ability to specify `resources` for all NuoDB containers
- create `securityContext` for all NuoDB containers (disabled by default)
- include the ability to specify `capabilities.drop` for NuoDB containers
- included `privileged` and `allowPrivilegeEscalation` for the container security context
- skip `nuodb` service account creation if needed (disabled by default). If `nuodb.serviceAccount` is `nil` or empty, then the default service account in the namespace will be used to run NuoDB pods.

**Testing**

- added new integration tests
- added new E2E test for restricted database deployment


stable/admin/templates/job.yaml
```diff
@@ -22,7 +22,9 @@ spec:
         release: {{ .Release.Name | quote }}
       name: job-lb-policy-{{ .Values.admin.lbPolicy | lower }}
     spec:
+      {{- if .Values.nuodb.serviceAccount }}
```
adriansuarez, line 25: Is it sufficient to specify `--set nuodb.addServiceAccount=false` or do we also have to specify `--set nuodb.serviceAccount=''` to omit the service account name?

stable/database/templates/_sidecar.tpl
```diff
@@ -5,6 +5,9 @@
   image: {{ template "nuocollector.image" . }}
   imagePullPolicy: {{ .Values.nuocollector.image.pullPolicy }}
   tty: true
+  resources:
+{{ toYaml .Values.nuocollector.resources | trim | indent 4 }}
```
adriansuarez, line 9: I think you should be able to intend these normally if you use `{{- ... }}`.

test/integration/template_admin_test.go
```diff
@@ -696,3 +767,65 @@ func TestAdminInitContainers(t *testing.T) {
...
+
+		// nuodb ServiceAccount is not created
+		_, err := helm.RenderTemplateE(t, options, helmChartPath, "release-name", []string{"templates/serviceaccount.yaml"})
+		assert.NotNil(t, err)
+		assert.Contains(t, err.Error(), "could not find template templates/serviceaccount.yaml in chart")
+
+		// the default ServiceAccount for the namespace will be used
+		output := helm.RenderTemplate(t, options, helmChartPath, "release-name", []string{"templates/statefulset.yaml"})
+		for _, obj := range testlib.SplitAndRenderStatefulSet(t, output, 1) {
+			assert.Empty(t, obj.Spec.Template.Spec.ServiceAccountName)
```
adriansuarez, line 827: Isn't this the same as the previous test case, but with the empty string replacing "nuodb"? In other words, we are still injecting `nuodb.serviceAccount` verbatim, but its value just happens to be `''`.

test/integration/template_collector_test.go
```diff
@@ -18,6 +18,8 @@ import (
 func checkSidecarContainers(t *testing.T, containers []v1.Container, options *helm.Options, chartPath string) {
 	require.NotEmpty(t, containers)
 	found := 0
+	securityContextEnabled := options.SetValues["admin.securityContext.enabledOnContainer"] == "true" ||
+		options.SetValues["database.securityContext.enabledOnContainer"] == "true"
```
adriansuarez, line 22: Should this be using the form that checks that the key is present, i.e. `val, ok := options.SetValues["admin.securityContext.enabledOnContainer"]`?

stable/admin/templates/job.yaml
adriansuarez: Following up on this, I can't tell if the tests actually cover this change. If `nuodb.serviceAccount` is explicitly specified as the empty string, and the default value of `serviceAccountName` is the empty string, then it doesn't seem like we need to conditionalize this, unless we are actually trying to suppress the injection of `nuodb` as a default, in which case, wouldn't the following by better?

```
  serviceAccountName: {{ default "" .Values.nuodb.serviceAccount }}
```

adriansuarez (approved):
Looks good, but I left a couple of comments on the conditionalizing of the `serviceAccountName`, which I'm not sure about.

test/integration/template_collector_test.go
sivanov-nuodb: The values for non-existing keys will be an empty string so this short form will work.

stable/admin/templates/job.yaml
sivanov-nuodb: > Is it sufficient to specify --set nuodb.addServiceAccount=false or do we also have to specify --set nuodb.serviceAccount='' to omit the service account name?

We would like to support the following use cases:
1. Automatically create a service account (the default). `nuodb.addServiceAccount=true` and `nuodb.serviceAccount='nuodb'`
2. Don't create the service account but use preprovisioned one.  `nuodb.addServiceAccount=false` and `nuodb.serviceAccount='foo'`
3.  Don't create the service account and use the default one. `nuodb.addServiceAccount=false` and (`nuodb.serviceAccount=''` or `nuodb.serviceAccount=null` or `nuodb.serviceAccount='default'`)


> If nuodb.serviceAccount is explicitly specified as the empty string, and the default value of serviceAccountName is the empty string, then it doesn't seem like we need to conditionalize this, unless we are actually trying to suppress the injection of nuodb as a default, in which case, wouldn't the following by better (with no {{ if }} block conditionalizing it)?

This is an interesting question. It looks like if the `serviceAccountName` is left empty in the Pod template, Kubernetes will inject the `default` SA at runtime. So your proposal will work and it won't change the default behaviour because the `nuodb.serviceAccount=nuodb` is set as default in the `values.yaml` provided with the charts.

stable/database/templates/_sidecar.tpl
sivanov-nuodb: I will try with `{{- toYaml ... | nindent 4 }}` so that we keep the leading new line.

test/integration/template_admin_test.go
sivanov-nuodb: When we parse the yaml to Kubernetes object, the `serviceAccountName` is initialized to it's zero value ("empty string"). Please see my comment above regarding runtime behaviour.

bkelly-ndb (approved):

stable/admin/templates/serviceaccount.yaml
```diff
@@ -1,4 +1,6 @@
+{{- if eq (include "defaulttrue" .Values.nuodb.addServiceAccount) "true" }}
```
adriansuarez, line 1: Should this be conditionalized on `nuodb.serviceAccount` being defined, that way we do not create a service account that does not actually get used?

sivanov-nuodb: This makes sense. I will fix it.

PR #269, opened 2022-02-01
Delete pods reported as CrashLoopBackOff

sivanov-nuodb:
The `TestKubernetesRestoreMultipleBackupGroups` test is doing multiple restore
operations in a sequence which increases the possibility for a container to be
reported as `CrashLoopBackOff`. This increases the testing time due to
container restart back-off and can fail the test. Monitor for such pods and
restart them to reset the back-off timer.

Database processes that are not selected for restore may experience schedule
delay and start after the restore coordinator SM already restored the database.
In that case _"Waiting for database restore to complete"_ message won't be see
in their log.

### DIFF
--- a/test/testlib/minikube_utilities.go
+++ b/test/testlib/minikube_utilities.go
@@ -1346,3 +1346,46 @@ func RemoveOrphanNamespaces(t *testing.T) {
 		}
 	}
 }
+
+func DeleteCrashLoopBackOffPods(t *testing.T, namespaceName string, collectLog bool) {
+	toDelete := make(map[string]corev1.ContainerStatus)
+	// find all Pods that have containers reported as CrashLoopBackOff
+	for _, pod := range FindAllPodsInSchema(t, namespaceName) {
+		for _, status := range pod.Status.ContainerStatuses {
+			if status.State.Waiting != nil && status.State.Waiting.Reason == "CrashLoopBackOff" {
+				toDelete[pod.Name] = status
+			}
+		}
+	}
+	for podName, previousStatus := range toDelete {
+		if collectLog {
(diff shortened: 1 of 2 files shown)


test/testlib/minikube_utilities.go
```diff
@@ -1346,3 +1346,46 @@ func RemoveOrphanNamespaces(t *testing.T) {
...
+	for _, pod := range FindAllPodsInSchema(t, namespaceName) {
+		for _, status := range pod.Status.ContainerStatuses {
+			if status.State.Waiting != nil && status.State.Waiting.Reason == "CrashLoopBackOff" {
+				toDelete[pod.Name] = status
+			}
+		}
+	}
+	for podName, previousStatus := range toDelete {
+		if collectLog {
+			GetAppLog(t, namespaceName, podName, "-crashloopbackoff", &corev1.PodLogOptions{})
```
adriansuarez, line 1362: Should this be moved down to right before the `DeletePod()` invocation so that we don't needlessly collect logs?

adriansuarez (approved):
Looks good.

sivanov-nuodb:
All local testing passed, however, running [the test multiple times](https://app.circleci.com/pipelines/github/nuodb/nuodb-helm-charts/890/workflows/c3129af8-6e9e-4102-87b1-34b2d9552306/jobs/4147) in CircleCI still fails with timing issues.

The pod restart logic may act on the old Kubernetes state and restart a pod concurrently with kubelet restarting a `CrashLoopBackOff` container. Deleting `CrashLoopBackOff` pods is not reliable in such tests because we often check the logs of the pod for certain messages like restoring an archive. For these reasons I'll abandon this PR.

PR #271, opened 2022-02-08
Support Ingress with SSL passthrough

sivanov-nuodb:
**Changes**

- Deploy Ingress resources for NuoDB Admin REST service, NuoDB Admin SQL, and TEs
- Automatically inject `external-address` and `external-port` TE labels if `database.te.ingress.enabled` is `true`
- Bump `k8s.io/api` to `v0.19.0` to get support for Ingress v1

**Testing**

- new integration tests
- new E2E test enabled with NuoDB image version 4.2.3+
- regression testing


go.sum
```diff
@@ -38,9 +38,11 @@ github.com/Azure/azure-sdk-for-go v38.1.0+incompatible/go.mod h1:9XXNKU+eRnpl9mo
 github.com/Azure/go-ansiterm v0.0.0-20170929234023-d6e3b3328b78/go.mod h1:LmzpDX56iTiv29bbRTIsUNlaFfuhWRQBWjQdVyAevI8=
 github.com/Azure/go-autorest/autorest v0.9.0/go.mod h1:xyHB1BMZT0cuDHU7I0+g046+BFDTQ8rEZB0s4Yfa6bI=
 github.com/Azure/go-autorest/autorest v0.9.3/go.mod h1:GsRuLYvwzLjjjRoWEIyMUaYq8GNUx2nRB378IPt/1p0=
+github.com/Azure/go-autorest/autorest v0.9.6/go.mod h1:/FALq9T/kS7b5J5qsQ+RSTUdAmGFqi0vUdVNNx8q630=
```
sivanov-nuodb, line 41: This file is automatically updated by `go`.

stable/admin/templates/ingress.yaml
```diff
@@ -0,0 +1,83 @@
...
+  - host: {{ .Values.admin.ingress.sql.hostname | quote }}
+    http:
+      paths:
+      - path: /
+        pathType: Prefix
+        backend:
+          service:
+            name: {{ .Values.admin.domain }}-{{ .Values.admin.serviceSuffix.clusterip }}
+            port:
+              name: 48004-tcp
```
adriansuarez, line 79: Do you know if this can be omitted? It does not seem to make any sense since we are actually not exposing a REST service on port 48004, but maybe we need this to force the load balancer to correctly dispatch connections with `SNI={{ .Values.admin.ingress.sql.hostname}}` to service.

stable/admin/templates/ingress.yaml
```diff
@@ -0,0 +1,83 @@
...
+    {{- end -}}
+  {{- end }}
+  {{- if .Values.admin.ingress.api.className }}
+  ingressClassName: {{ .Values.admin.ingress.api.className | quote }}
+  {{- end }}
+  rules:
+  - host: {{ .Values.admin.ingress.api.hostname | quote }}
+    http:
+      paths:
+      - path: {{ default "/" .Values.admin.ingress.api.path }}
```
adriansuarez, line 33: Would any path other than `/` work with `ssl-passthrough=true` for the haproxy-ingress? Also, what would be the use case here, `nuocmd` clients connecting from outside the Kubernetes cluster?

stable/admin/values.yaml
```diff
@@ -126,6 +126,48 @@ admin:
...
+      className: ""
+      # custom annotations that are set on the Ingress resource
+      annotations: 
+        ingress.kubernetes.io/ssl-passthrough: "true"
+      # Enable TLS termination in the ingress controller; it is recommended to
+      # use SSL Passthrough feature instead
+      tls: false
+      # the name of the secret used by Ingress controller to terminate the TLS
+      # traffic on port 443
+      secretName: ""
```
adriansuarez, line 169: How would this work? Would the Ingress controller communicate over plaintext TCP with the admin?

stable/admin/templates/ingress.yaml
sivanov-nuodb: We need to expose NuoAdmin port 48004 used by the SQL clients to connect to the admin. This section of the ingress resource is configuring the `backend` service. I don't know if we can commit the `path` and the `pathType` as normally they are not interpreted by the Ingress controller when SSL passthrough is activated.

stable/admin/templates/ingress.yaml
sivanov-nuodb: > Would any path other than / work with ssl-passthrough=true for the haproxy-ingress?

Correct. The `path` doesn't have any effect when `ssl-passthrough=true` because the TLS can't be decrypted by the controller. However, for the REST service ssl-passthrough is not mandatory as it is just HTTP service so someone might want to disable in the NuoAdmin, use the Ingress controller to terminate TLS with the client, and use basic authentication. I decided to not assume the configuration but make it as flexible as possible.

> Also, what would be the use case here, nuocmd clients connecting from outside the Kubernetes cluster?

Correct. This exposes the Admin REST service.

stable/admin/values.yaml
sivanov-nuodb: Using a secret to terminate the SQL client TLS connection on the HAProxy with `ssl-passthrough=false` doesn't work nowadays probably because the ingress controller assumes that the traffic is HTTP and tries to inspect the headers, etc. 

I didn't test with different ingress controllers so I decided to keep these values to allow maximum configuration flexibility and support of a variety of use cases in the future.

stable/admin/templates/ingress.yaml
```diff
@@ -0,0 +1,83 @@
+{{- if .Values.admin.ingress }}
+{{- if eq (include "defaultfalse" .Values.admin.ingress.enabled) "true" }}
+{{- if .Values.admin.ingress.api }}
+{{- if .Values.admin.ingress.api.hostname }}
+apiVersion: networking.k8s.io/v1
```
sivanov-nuodb, line 5: We need to document that Ingress can be enabled only with Kubernetes v1.19+.
I didn't include the `networking.k8s.io/v1beta1` version of the Ingress resource.

stable/admin/templates/ingress.yaml
adriansuarez: So does it make sense to have a path other than `/`? If `nuocmd` is being issued by an external client, then it would have to have the path prefix embedded in the `--api-server` value or something, e.g. `https://<ingress host>/<path>`. I doubt that this would actually work, since I'm sure that there are places where we assume that `--api-server` has the form `<protocol>://<host>:<port>`, `<host>:<port>`, or just `<host>`.

stable/admin/values.yaml
adriansuarez: My preference would be to only expose configuration options that we know work. The fact that we are exposing these options and providing some documentation implies that we have validated somehow, but we only know that they definitely _don't_ work with HAProxy.

stable/admin/templates/ingress.yaml
sivanov-nuodb: That's a good point. I only tested with the default path (`/`), Admin without TLS, and TLS termination on the Ingress controller.
If we have some assumptions for the URL in the client, then it would be better to not expose the path.

stable/admin/values.yaml
sivanov-nuodb: Okay, I was having the same dilemma. Making it flexible in terms of configuration or exposing only those options that we have tested with. We will assume that the SSL passthough feature is turned `on` don't expose the following variables at all:
* `admin.ingress.api.path`
* `admin.ingress.api.tls`
* `admin.ingress.api.secretName`
* `admin.ingress.sql.tls`
* `admin.ingress.sql.secretName`
* `database.te.ingress.tls`
* `database.te.ingress.secretName`

sivanov-nuodb:
@adriansuarez I pushed another commit that removes the values that we discussed above.

adriansuarez (approved):
Looks good.

bkelly-ndb (approved):

PR #272, opened 2022-02-14
Add runAs in the container security context

sivanov-nuodb:
**Changes**

- include `runAsUser` and `runAsGroup` into the container security context
- specify `runAsNonRoot` if `runAsUser` != 0 and (`initContainers.runInitDisk` == false or `initContainers.runInitDiskAsRoot` == false)

**Testing**
- new integration tests
- regression testing


stable/admin/templates/_helpers.tpl
```diff
@@ -135,22 +135,30 @@ imagePullSecrets:
...
+
+{{/*
+Get security context runAsUser and runAsGroup
+*/}}
+{{- define "sc.runAs" -}}
+{{- if eq (include "defaultfalse" .Values.admin.securityContext.enabled) "true" }}
+runAsUser: {{ default 1000 .Values.admin.securityContext.runAsUser }}
+runAsGroup: 0
+  {{- if and (or (eq (include "defaulttrue" .Values.admin.initContainers.runInitDisk) "false") (eq (include "defaulttrue" .Values.admin.initContainers.runInitDiskAsRoot) "false")) (ne (toString (default 1000 .Values.admin.securityContext.runAsUser)) "0")  }}
+runAsNonRoot: true
```
adriansuarez, line 154: Does this actually have to be specified? It seems totally redundant, but I guess if the security rules require it then we have to specify it.

sivanov-nuodb: This is needed to comply with [Require runAsNonRoot policy](https://kyverno.io/policies/pod-security/restricted/require-run-as-nonroot/require-run-as-nonroot/). Kubernetes does some extra checks if this is set to true.

bkelly-ndb (approved):

PR #275, opened 2022-03-01
Change permissions on backup volume

adriansuarez:
Change permissions on backup volume and add test of SM init container.


bkelly-ndb (approved):

sivanov-nuodb (approved):
Thanks for fixing this!

PR #280, opened 2022-05-19
Set publishNotReadyAddresses to true for TE services

sivanov-nuodb:
**Issue**

A SQL application will fail to connect to a TE group that doesn't have _ready_ TE pods with the following error message:

```
Connection failed: connection failed: 6: connection to 10.97.49.14:48006 failed: Connection refused (61)
```

This issue affects external database access in scenarios where multiple TE groups are running in different availability zones or Kubernetes clusters. Normally the SQL application will be configured with Load balancer policies that prefer the TE group in the local AZ or cluster. 

The switch to the remote TE group happens instantly if the local one fails or is unavailable, however, the problem is seen when the previously failed TE group is brought up again. The AP will return a local TE external address (the `LoadBalancer` service external address) as the TE will be in `RUNNING` state, however, the service will not route the traffic to unready endpoints preventing the connection to a TE. Until at least one TE pod in the local TE group becomes ready, the SQL app won't be able to obtain a new database connection (typically around 30sec).

**Change**

Set `publishNotReadyAddresses=true` on TE group services so that incoming connections can be sent to unready endpoints. This should be safe because the AP won't advertise a TE that is not `RUNNING`.

### DIFF
--- a/stable/database/templates/service-clusterip.yaml
+++ b/stable/database/templates/service-clusterip.yaml
@@ -33,6 +33,7 @@ metadata:
   annotations:
     description: |
       Cluster IP service targeting specific Helm release permitting us to load balance within the cluster.
+    "service.alpha.kubernetes.io/tolerate-unready-endpoints": "true"
   labels:
     app: {{ template "database.fullname" . }}
     group: nuodb
--- a/stable/database/templates/service-clusterip.yaml
+++ b/stable/database/templates/service-clusterip.yaml
@@ -42,6 +42,7 @@ metadata:
     release: {{ .Release.Name | quote }}
   name: {{ template "database.fullname" . }}-{{ default .Values.admin.serviceSuffix.clusterip .Values.database.serviceSuffix.clusterip }}
 spec:
+  publishNotReadyAddresses: true
   ports:
   - { name: 48006-tcp,  port: 48006,  protocol: TCP,  targetPort: 48006 }
   selector:
--- a/stable/database/templates/service.yaml
+++ b/stable/database/templates/service.yaml
@@ -14,6 +14,7 @@ metadata:
     release: {{ .Release.Name | quote }}
   name: {{ template "database.externalServiceName" . }}
 spec:
+  publishNotReadyAddresses: true
   ports:
   - { name: 48006-tcp,  port: 48006,  protocol: TCP,  targetPort: 48006 }
   selector:
--- a/stable/database/templates/service.yaml
+++ b/stable/database/templates/service.yaml
@@ -4,6 +4,7 @@ kind: Service
 metadata:
   annotations:
     description: "Service (and load-balancer) for TE pods."
+    "service.alpha.kubernetes.io/tolerate-unready-endpoints": "true"
 {{- include "database.externalAccessAnnotations" . | indent 4 }}
   labels:
     app: {{ template "database.fullname" . }}
--- a/test/integration/template_database_test.go
+++ b/test/integration/template_database_test.go
@@ -85,6 +85,7 @@ func TestDatabaseClusterServiceRenders(t *testing.T) {
 			// This is the cluster IP targeting only TEs in this TE group
 			assert.Equal(t, "te", obj.Spec.Selector["component"])
 			assert.Equal(t, "release-name-nuodb-cluster0-demo-database", obj.Spec.Selector["app"])
+			assert.True(t, obj.Spec.PublishNotReadyAddresses)
 		}
 	}
 }
@@ -126,6 +127,7 @@ func TestDatabaseServiceRenders(t *testing.T) {
 	for _, obj := range testlib.SplitAndRenderService(t, output, 1) {
 		assert.Equal(t, "release-name-nuodb-cluster0-demo-database-balancer", obj.Name)
 		assert.Equal(t, v1.ServiceTypeLoadBalancer, obj.Spec.Type)
+		assert.True(t, obj.Spec.PublishNotReadyAddresses)
 		assert.Equal(t, "release-name-nuodb-cluster0-demo-database", obj.Spec.Selector["app"])
 		assert.Equal(t, "te", obj.Spec.Selector["component"])
 		assert.Contains(t, obj.Annotations, "service.beta.kubernetes.io/aws-load-balancer-internal")


adriansuarez (approved):
Looks good.

stable/database/templates/service-clusterip.yaml
```diff
@@ -42,6 +42,7 @@ metadata:
     release: {{ .Release.Name | quote }}
   name: {{ template "database.fullname" . }}-{{ default .Values.admin.serviceSuffix.clusterip .Values.database.serviceSuffix.clusterip }}
 spec:
+  publishNotReadyAddresses: true
```
acabrele, line 45: In the example I experienced, with a multi-cluster deployment, both deployments are primary -  a secondary was not used.  Don't we need to add it to primary too?

acabrele (approved):
Our admin services also include the deprecated "service.alpha.kubernetes.io/tolerate-unready-endpoints": "true" annotation.  Do we want that here also, for consistency?

stable/database/templates/service-clusterip.yaml
sivanov-nuodb: If I'm not mistaken, in your test you have used external access using NLBs. This will utilize the "xxx-balancer" services for the TEs, which is now configured with `publishNotReadyAddresses=true` (see _service.yaml_). I also changed the TE group `ClusterIP` services because they are used when external access using `Ingress` is configured. 

The `demo-clusterip` will most probably be used only for `direct` TE access. (i.e. not a TE returned by the admin) so I think that we shouldn't publish unready endpoints.

sivanov-nuodb:
> Do we want that here also, for consistency?

Done.

sivanov-nuodb:
> This should be safe because the AP won't advertise a TE that is not RUNNING.

@adriansuarez , @acabrele This is correct, however, we don't have one-to-one mapping between TEs and Network Load Balancers (NLBs).
The proposed change will _not_ work correctly in a scenario where more than 1 TE is provisioned in a TE group. As both TEs will have the same external address (the same NLB), Kubernetes service may load-balance the incoming traffic to a TE in the same group that is not `RUNNING`.

I think I will abandon this PR in lieu of https://github.com/nuodb/nuodb-helm-charts/pull/281 which doesn't solve the problem entirely but minimizes the delay to 10 sec.

acabrele:
> 

So, the scenario is that the pod is still running, but the TE inside it is not, in such a way that has not caused it to be restarted by k8s in a reasonable time?  I guess we are trading issues on one side for the other.  The question is, which is more likely to be a noticeable problem.

sivanov-nuodb:
I was thinking about scenario where we have several healthy TEs in a group and just one "flaky" TE that got constantly restarted by Kubernetes. This will disrupt the connectivity to the whole TE group.

From example see below how we get random connectivity failures:

```
te-groupA-nuodb-cluster0-demo-585c9d86c6-hn2sf       0/1     Error       4          110s
te-groupA-nuodb-cluster0-demo-585c9d86c6-npd4z       1/1     Running     0          2m22s
te-groupA-nuodb-cluster0-demo-585c9d86c6-qzjkb       1/1     Running     0          2m22s
te-groupB-nuodb-cluster0-demo-5564696fb9-ns587      1/1     Running     1          60m
```

```
2022-05-23T15:31:20+0300 Connected to group A
2022-05-23T15:31:21+0300 2022-05-23T15:31:21.635+0300 [12814] Failed to resolve address 10.111.60.178 to a hostname in 0.001 seconds, error 8: nodename nor servname provided, or not known
Connection failed: connection failed: 6: connection to 10.111.60.178:48006 failed: Connection refused (61)
2022-05-23T15:31:22+0300 2022-05-23T15:31:22.663+0300 [12836] Failed to resolve address 10.111.60.178 to a hostname in 0.001 seconds, error 8: nodename nor servname provided, or not known
Connection failed: connection failed: 6: connection to 10.111.60.178:48006 failed: Connection refused (61)
2022-05-23T15:31:23+0300 2022-05-23T15:31:23.692+0300 [12862] Failed to resolve address 10.111.60.178 to a hostname in 0.001 seconds, error 8: nodename nor servname provided, or not known
Connection failed: connection failed: 6: connection to 10.111.60.178:48006 failed: Connection refused (61)
2022-05-23T15:31:24+0300 Connected to group A
2022-05-23T15:31:25+0300 Connected to group A
2022-05-23T15:31:26+0300 Connected to group A
2022-05-23T15:31:27+0300 Connected to group A
2022-05-23T15:31:28+0300 Connected to group A
2022-05-23T15:31:29+0300 Connected to group A
2022-05-23T15:31:30+0300 2022-05-23T15:31:30.930+0300 [13060] Failed to resolve address 10.111.60.178 to a hostname in 0.001 seconds, error 8: nodename nor servname provided, or not known
Connection failed: connection failed: 6: connection to 10.111.60.178:48006 failed: Connection refused (61)
2022-05-23T15:31:31+0300 Connected to group A
2022-05-23T15:31:33+0300 Connected to group A
2022-05-23T15:31:34+0300 2022-05-23T15:31:34.084+0300 [13150] Failed to resolve address 10.111.60.178 to a hostname in 0.001 seconds, error 8: nodename nor servname provided, or not known
Connection failed: connection failed: 6: connection to 10.111.60.178:48006 failed: Connection refused (61)
2022-05-23T15:31:35+0300 Connected to group A
2022-05-23T15:31:36+0300 2022-05-23T15:31:36.145+0300 [13203] Failed to resolve address 10.111.60.178 to a hostname in 0.001 seconds, error 8: nodename nor servname provided, or not known
Connection failed: connection failed: 6: connection to 10.111.60.178:48006 failed: Connection refused (61)
2022-05-23T15:31:37+0300 Connected to group A
2022-05-23T15:31:38+0300 Connected to group A
2022-05-23T15:31:39+0300 Connected to group A
2022-05-23T15:31:40+0300 Connected to group A
2022-05-23T15:31:41+0300 2022-05-23T15:31:41.346+0300 [13335] Failed to resolve address 10.111.60.178 to a hostname in 0.001 seconds, error 8: nodename nor servname provided, or not known
Connection failed: connection failed: 6: connection to 10.111.60.178:48006 failed: Connection refused (61)
2022-05-23T15:31:42+0300 Connected to group A
2022-05-23T15:31:43+0300 Connected to group A
2022-05-23T15:31:44+0300 2022-05-23T15:31:44.464+0300 [13418] Failed to resolve address 10.111.60.178 to a hostname in 0.001 seconds, error 8: nodename nor servname provided, or not known
Connection failed: connection failed: 6: connection to 10.111.60.178:48006 failed: Connection refused (61)
2022-05-23T15:31:45+0300 Connected to group A
2022-05-23T15:31:46+0300 Connected to group A
2022-05-23T15:31:47+0300 Connected to group A
```

acabrele:
Right, I get it.  The backoff would make it less of a problem over time, but it is a problem for sure.

Maybe we can make this defaulted but configurable instead?  I guess the other timing changes in 281 wil go most of the way to improve it, anyway.

sivanov-nuodb:
Correct the backoff will make the problem appear every 5 min. At this point I will leave it as is because enabling the change might cause more problems than good. Maybe we can think of making the readiness times configurable.

PR #282, opened 2022-05-23
Remove port 48005 from AP service used for external access

adriansuarez:
This change removes port 48005 from the service created by
admin.externalAccess.enabled=true.

Port 48005 is used for internal communication among APs and not by
external clients. When this port is exposed to external access using
Network Load Balancer in AWS, stacktraces are generated by the Thrift
server that listens on that port due to the load balancer pinging the AP
on that port, even though there is no reason for the port to be exposed
through a load balancer.


sivanov-nuodb (approved):
Looks good.

bkelly-ndb (approved):

PR #284, opened 2022-05-26
Update image used by CircleCI

adriansuarez:
This change updates the Ubuntu image used to run tests using CircleCI,
updates the version of Minikube used, and removes configuration file
wrangling that is superfluous now that `minikube start` is invoked as
the non-root user.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -79,7 +79,7 @@ jobs:
         type: string
 
     machine:
-      image: ubuntu-1604:202007-01
+      image: ubuntu-2004:202201-02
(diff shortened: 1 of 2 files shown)


adriansuarez:
@sivanov-nuodb: The multi-cluster tests are failing apparently due to some kind of resource constraint issue in AKS. I can't tell how this would be related to updating the image used to run the tests.

sivanov-nuodb (approved):
Looks good. Thanks for doing this!

bkelly-ndb (approved):

PR #287, opened 2022-06-01
Allow priorityClassName to be specified for admin, SM, and TE pods

adriansuarez:


bkelly-ndb (approved):
Looks good, what happens here if we set a priority class name which doesn't exists, the helm deployment will fail, right? If not should we validate the priorityClass exists?

adriansuarez:
> Looks good, what happens here if we set a priority class name which doesn't exists, the helm deployment will fail, right? If not should we validate the priorityClass exists?

It doesn't seem to prevent the Helm charts from installing, but it prevents the StatefulSets/Deployments from creating Pods. This is the error message that I see:

```
$ kubectl -n testkubernetespriorityclassnonexistent-vmfidj describe statefulsets.apps admin-omed1g-nuodb-cluster0
...
Events:
  Type     Reason            Age                 From                    Message
  ----     ------            ----                ----                    -------
  Normal   SuccessfulCreate  20s                 statefulset-controller  create Claim raftlog-admin-omed1g-nuodb-cluster0-0 Pod admin-omed1g-nuodb-cluster0-0 in StatefulSet admin-omed1g-nuodb-cluster0 success
  Warning  FailedCreate      10s (x12 over 20s)  statefulset-controller  create Pod admin-omed1g-nuodb-cluster0-0 in StatefulSet admin-omed1g-nuodb-cluster0 failed error: pods "admin-omed1g-nuodb-cluster0-0" is forbidden: no PriorityClass with name high-priority was found
```

test/minikube/minikube_priority_class_test.go
```diff
@@ -0,0 +1,136 @@
...
+	admin0 := adminStatefulSet + "-0"
+
+	defer testlib.Teardown(testlib.TEARDOWN_DATABASE)
+	testlib.StartDatabaseNoWait(t, namespaceName, admin0, &options)
+	testlib.Await(t, func() bool {
+		filter := metav1.ListOptions{FieldSelector: "type=Warning"}
+		events := testlib.GetEvents(t, namespaceName, filter)
+		for _, event := range events {
+			if !strings.Contains(event.Message, "no PriorityClass with name high-priority was found") {
+				return false
```
sivanov-nuodb, line 59: Am I understanding this correctly? Returning `false` inside the loop means that we should have _only_ events with the expected string. Can we have other warnings in the system?

Maybe it's better to count all occurrences of the expected event and wait until they become 4.

sivanov-nuodb:
Looks good to me. I left a minor comment in the functional test.

sivanov-nuodb (approved):

stable/admin/templates/statefulset.yaml
```diff
@@ -44,6 +44,7 @@ spec:
         group: nuodb
     spec:
       serviceAccountName: {{ default "" .Values.nuodb.serviceAccount }}
+      priorityClassName: {{ default "" .Values.admin.priorityClass }}
```
oswcab, line 47: Question: any reason why is this different from the ones for the database? I mean, couldn't   be inside an if conditional as below?

test/minikube/minikube_priority_class_test.go
adriansuarez: Events are namespace-bound, so I wouldn't expect there to be any events of type `Warning` that are not generated by the failure to create pods. I could make the check more specific to account for that and also to check that the event is associated with a pod from one of our StatefulSets and Deployments and has reason `FailedCreate`.

stable/admin/templates/statefulset.yaml
adriansuarez: The admin chart only has the admin StatefulSet, while the database chart has the SM StatefulSets and TE Deployment. The database chart nests both values under `database.priorityClasses`, which may be absent entirely, so we guard against that by checking that `database.priorityClasses` exists. This isn't necessary with the admin chart because there is no nesting.

test/minikube/minikube_priority_class_test.go
sivanov-nuodb: Thanks for updating the test case!

PR #288, opened 2022-06-20
Use batch/v1/CronJob if available

adriansuarez:
This change conditionally defines the API version for the backup
cronjobs as batch/v1 if kind batch/v1/CronJob is available, and
otherwise falls back to batch/v1beta1.


sivanov-nuodb (approved):
Looks good to me.

butson (approved):
looks good

PR #289, opened 2022-06-20
Dbutson/rolling upgrade

butson:
two changes.

1. rolling upgrade if configFiles changed.
2. cronjob apiversion to v1 (betav1 is deprecated).

### DIFF
--- a/stable/admin/templates/statefulset.yaml
+++ b/stable/admin/templates/statefulset.yaml
@@ -27,8 +27,15 @@ spec:
   template:
     metadata:
       name: {{ template "admin.fullname" . }}
-      {{- if .Values.admin.podAnnotations }}
       annotations:
+      {{- if .Values.admin.configFiles }}
+        {{- with .Values.admin.configFiles }}
+        checksum/config: {{ toYaml . | sha256sum }}
+        {{- end }}
+      {{- else }}
+        checksum/config: 0
+      {{- end }}
+      {{- if .Values.admin.podAnnotations }}
 {{ toYaml .Values.admin.podAnnotations | trim | indent 8}}
       {{- end }}
       labels:
--- a/stable/database/templates/cronjob.yaml
+++ b/stable/database/templates/cronjob.yaml
@@ -13,8 +13,13 @@
 {{- range $i, $backupGroup := $backupGroups -}}
   {{- with $globalScope }}
     {{ $backupGroupLabels := include "hotcopy.group.labels" (merge (dict "backupGroup" $backupGroup) .) }}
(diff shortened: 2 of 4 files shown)


stable/admin/templates/statefulset.yaml
```diff
@@ -27,8 +27,15 @@ spec:
   template:
     metadata:
       name: {{ template "admin.fullname" . }}
-      {{- if .Values.admin.podAnnotations }}
       annotations:
+      {{- if .Values.admin.configFiles }}
+        {{- with .Values.admin.configFiles }}
+        checksum/config: {{ toYaml . | sha256sum }}
```
adriansuarez, line 33: Is the issue here is that changes to volume mounts are not detected?

stable/database/templates/cronjob.yaml
```diff
@@ -13,8 +13,13 @@
 {{- range $i, $backupGroup := $backupGroups -}}
   {{- with $globalScope }}
     {{ $backupGroupLabels := include "hotcopy.group.labels" (merge (dict "backupGroup" $backupGroup) .) }}
+{{- if .Capabilities.APIVersions.Has "batch/v1" -}}
+ {{- $cronJobVersion := "batch/v1" -}}
+{{- else -}}
+ {{- $cronJobVersion := "batch/v1beta1" -}}
+{{- end }}
```
adriansuarez, line 20: This isn't quite right. The `batch/v1` API version can exist without the `batch/v1/CronJob` kind existing. I already have a PR that addresses this: https://github.com/nuodb/nuodb-helm-charts/pull/288

stable/database/templates/deployment.yaml
```diff
@@ -24,8 +24,16 @@ spec:
     type: RollingUpdate
   template:
     metadata:
-      {{- if .Values.database.podAnnotations }}
+      {{- with .Values.database.te.nodeSelector }}
```
adriansuarez, line 27: Why was this moved up here?

stable/admin/templates/statefulset.yaml
sivanov-nuodb, line 33: What if we have more than 1 config file? Shouldn't we include the config file "key" in the annotation key?

sivanov-nuodb: Yes, I think that this is a known issue - https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments

stable/database/templates/deployment.yaml
butson: This was a mistake.  I've put it back to where it belongs.

PR #292, opened 2022-06-21
added pod topology constraints

butson:
For Medidata, I added topologySpreadConstraints to TE and SM to allow an even distribution of pods across availability zones.   In order to use:

database:
   te:
     topologySpreadConstraints: |-
      - maxSkew: 1
        topologyKey: topology.kubernetes.io/zone
        whenUnsatisfiable: DoNotSchedule
        labelSelector:
          matchLabels:
            component: te

not that we use `tpl` to format.  this is similar to what is done for affinity.

### DIFF
--- a/stable/database/templates/deployment.yaml
+++ b/stable/database/templates/deployment.yaml
@@ -37,6 +37,10 @@ spec:
       priorityClassName: {{ default "" .Values.database.priorityClasses.te }}
       {{- end }}
       {{- include "securityContext" . | indent 6 }}
+      {{- if .Values.database.te.topologySpreadConstraints }}
+      topologySpreadConstraints:
+{{ tpl .Values.database.te.topologySpreadConstraints . | trim | indent 8 }}
+      {{- end }}
       {{- with .Values.database.te.nodeSelector }}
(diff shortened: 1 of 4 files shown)


stable/database/templates/deployment.yaml
```diff
@@ -37,6 +37,10 @@ spec:
       priorityClassName: {{ default "" .Values.database.priorityClasses.te }}
       {{- end }}
       {{- include "securityContext" . | indent 6 }}
+      {{- if .Values.database.te.topologySpreadConstraints }}
+      topologySpreadConstraints:
+{{ tpl .Values.database.te.topologySpreadConstraints . | trim | indent 8 }}
```
sivanov-nuodb, line 42: You could preserve the indentation with:
`{{- tpl .Values.database.te.topologySpreadConstraints . | trim | nindent 8 }}`

butson: indent is used in other areas for similar sections such as affinity and tolerations.

sivanov-nuodb (approved):
Looks good to me.

PR #293, opened 2022-06-22
allow database.env.NUODB_ARCHIVEDIR and NUODB_JOURNALDIR to be set to…

butson:
… override the default of /var/opt/nuodb/archive/<domain>/<dbname>.  This was original intend and then broken when support for restore was added to nuosm.   This change will break any upgrade that currently sets these variables.  I don't believe that case exists,  if so then the variable settings will need to modified when upgrading.

### DIFF
--- a/stable/database/files/nuosm
+++ b/stable/database/files/nuosm
@@ -9,9 +9,9 @@
 
 . ${NUODB_HOME}/etc/nuodb_setup.sh
 
-: ${NUODB_ARCHIVEDIR:=/var/opt/nuodb/archive}
+: ${NUODB_ARCHIVEDIR:=/var/opt/nuodb/archive/${NUODB_DOMAIN}/${DB_NAME}}
 : ${NUODB_BACKUPDIR:=/var/opt/nuodb/backup}
-: ${NUODB_JOURNALDIR:=/var/opt/nuodb/journal}
+: ${NUODB_JOURNALDIR:=/var/opt/nuodb/journal/${NUODB_DOMAIN}/${DB_NAME}}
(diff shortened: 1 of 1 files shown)


sivanov-nuodb:
@butson This breaks some of the restore tests.

adriansuarez (approved):
Looks fine, now that the tests are green. I'll defer to @sivanov-nuodb on whether this is a safe fix.

stable/database/files/nuosm
```diff
@@ -9,9 +9,9 @@
...
 . ${NUODB_HOME}/etc/nuodb_setup.sh
 
-: ${NUODB_ARCHIVEDIR:=/var/opt/nuodb/archive}
+: ${NUODB_ARCHIVEDIR:=/var/opt/nuodb/archive/${NUODB_DOMAIN}/${DB_NAME}}
 : ${NUODB_BACKUPDIR:=/var/opt/nuodb/backup}
-: ${NUODB_JOURNALDIR:=/var/opt/nuodb/journal}
+: ${NUODB_JOURNALDIR:=/var/opt/nuodb/journal/${NUODB_DOMAIN}/${DB_NAME}}
 : ${NUODB_STORAGE_PASSWORDS_DIR:=/etc/nuodb/tde}
 : ${NUODB_DOMAIN:="nuodb"}
 : ${DB_NAME:="demo"}
```
sivanov-nuodb, line 17: For clarity, you may want to move these two lines before defaulting the `${NUODB_ARCHIVEDIR}`.
It works now because `${NUODB_DOMAIN}` and `${DB_NAME}` are both set in the templates.

sivanov-nuodb (approved):
Looks good to me.

PR #294, opened 2022-06-27
Dbutson/latency check

butson:
allows via env variable (NUOADMIN_DISTRIBUTION_POLICY) to set --admin-distribution-policy when calling docker start sm (nuosm) or docker start te (nuote).   This variable is currently also used in (nuote and nuosm) to instead find nearest admin and set server-id.  This code nuolatency.py should probably be in nuodocker as a distribution policy.  But for now is a handled in helm scripting (nuodocker requires change to image or cloning of nuodocker.py)

### DIFF
--- a/stable/database/files/nuolatency.py
+++ b/stable/database/files/nuolatency.py
@@ -20,7 +20,7 @@ def latency_point(host, port = 48004, timeout = 5):
         s.connect((host, int(port)))
         s.shutdown(socket.SHUT_RD)
     except:
-        return 99999
+        raise ValueError("Connection Timeout")
 
     # Stop Timer
     s_runtime = (time() - s_start) * 1000
@@ -28,9 +28,12 @@ def latency_point(host, port = 48004, timeout = 5):
 
 def measure(host,port=48005,runs=9):
     total = 0
-    for n in range(0,runs):
-        total += latency_point(host)
-    return total / float(runs)
+    try:
+        for n in range(0,runs):
+            total += latency_point(host)
+        return total / float(runs)
+    except Value:
+        return 99999
 
 class LatencyCommands(nuodb_cli.AdminCommands):
 
--- a/stable/database/files/nuolatency.py
+++ b/stable/database/files/nuolatency.py
@@ -1,0 +1,54 @@
+import socket, pprint
+from time import time
+
+from pynuoadmin import nuodb_cli
+
+def latency_point(host, port = 48004, timeout = 5):
+    """
+    Calculate a latency point using sockets. If error return large number
+    """
+
+    # New Socket and Time out
+    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    s.settimeout(timeout)
+
+    # Start a timer
+    s_start = time()
+
+    # Try to Connect
+    try:
+        s.connect((host, int(port)))
+        s.shutdown(socket.SHUT_RD)
+    except:
+        return 99999
+
+    # Stop Timer
+    s_runtime = (time() - s_start) * 1000
+    return float(s_runtime)
+
+def measure(host,port=48005,runs=9):
+    total = 0
+    for n in range(0,runs):
+        total += latency_point(host)
+    return total / float(runs)
+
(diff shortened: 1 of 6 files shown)


stable/database/files/nuolatency.py
```diff
@@ -0,0 +1,54 @@
...
+    @nuodb_cli.subcommand
+    def find_closest_admin(self):
+        """
+        Find the closest (based upon latency) admin process.  This should be in same zone
+        """
+
+        closest = None
+        closest_measurement = 99999
+        
+        for s in self.conn._get_all('peers'):
```
adriansuarez, line 46: What's wrong with `AdminConnection.get_servers()`?

stable/database/files/nuolatency.py
```diff
@@ -0,0 +1,54 @@
...
+    # Try to Connect
+    try:
+        s.connect((host, int(port)))
+        s.shutdown(socket.SHUT_RD)
+    except:
+        return 99999
+
+    # Stop Timer
+    s_runtime = (time() - s_start) * 1000
+    return float(s_runtime)
```
adriansuarez, line 27: This is supposed to return milliseconds?

stable/database/files/nuolatency.py
```diff
@@ -0,0 +1,54 @@
...
+
+    # Start a timer
+    s_start = time()
+
+    # Try to Connect
+    try:
+        s.connect((host, int(port)))
+        s.shutdown(socket.SHUT_RD)
+    except:
+        return 99999
```
adriansuarez, line 23: Do you want the returned latency to be large that we do not use the admin that generated the exception? I would just propagate the exception and let the loop that iterates over the admins catch the exception and skip it.

If you don't want to remove the admin from consideration completely in case the others also generate exceptions, then I guess you could do this, but it would be tricky to distinguish between admins that have high latency (higher than 5 seconds) and ones that are completely unresponsive.

adriansuarez:
My concern with this change is that it may cause database processes to take longer to start up. Do you know what the impact is on start up times?

stable/database/files/nuolatency.py
butson: Will get servers also output to screen.   _get_all is quiet

stable/database/files/nuolatency.py
butson: Yes it probably does not help to return a large number and keep iterating.

butson:
For my test, the impact is probably 5-6 seconds or so.  Given that we are talking about .300 ms in worse case.  I'm not sure that a slow start is a bad thing given the benefits.  We can reduce the number of iterations from 9 to something smaller. 

It's also conditioned now on setting the variable NUOADMIN_DISTRIBUTION_POLICY=nearest.  So if you don't want to live with the extra cost and okay with connecting to round robin admin then okay.   I don't know how long it takes to start with having to coordinate things via raft either as is done now.

stable/database/files/nuolatency.py
butson: @adriansuarez  I pushed a change that throws exception and then handle the exception and exit the loop with 99999

adriansuarez:
This also needs to be conditionalized on the version of the image having support for `NUOCMD_PLUGINS`, otherwise we break all image versions that do not support it. I guess it is off if the "nearest" distribution policy is not specified, but it would be better to generate a message that is comprehensible to the user.

butson:
@adriansuarez   how would we check the version of the image and what is the condition.   I guess the easiest thing to do would be to conditional it on whether NUOCMD_PLUGIN is already set.  As it's not set in previous versions and will be set if supported as it's a builtin env of the image.

adriansuarez:
> @adriansuarez how would we check the version of the image and what is the condition. I guess the easiest thing to do would be to conditional it on whether NUOCMD_PLUGIN is already set. As it's not set in previous versions and will be set if supported as it's a builtin env of the image.

Good point. Yes, doing `env | grep -q NUOCMD_PLUGINS` seems like an easy way to check.

PR #295, opened 2022-06-27
Dbutson/label volumes

butson:
add use label to pvc.  not necessarily needed but,  felt it would be useful.   Also added label to backup volume to associate with backup group.  Added annotations and labels to backup cronjobs.  All of this is done to assist in watching for backup jobs to complete and being able to snapshot volumes on completion of backup job.

### DIFF
--- a/stable/admin/templates/statefulset.yaml
+++ b/stable/admin/templates/statefulset.yaml
@@ -1,4 +1,4 @@
-apiVersion: "apps/v1"
+admin/templates/statefulset.yamlapiVersion: "apps/v1"
 kind: StatefulSet
 metadata:
   annotations:
@@ -251,6 +251,7 @@ spec:
   - metadata:
       name: raftlog
       labels:
+        use: raftlog-volume
         {{- include "admin.resourceLabels" . | nindent 8 }}
     spec:
     {{- if .Values.admin.persistence.storageClass }}
@@ -271,6 +272,7 @@ spec:
   - metadata:
       name: log-volume
       labels:
+        use: log-volume
         {{- include "admin.resourceLabels" . | nindent 8 }}
     spec:
     {{- if .Values.admin.logPersistence.storageClass }}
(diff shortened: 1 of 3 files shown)


stable/admin/templates/statefulset.yaml
```diff
@@ -258,6 +258,7 @@ spec:
   - metadata:
       name: raftlog
       labels:
+        use: raftlog-volume
```
sivanov-nuodb, line 261: > add use label to pvc. not necessarily needed but, felt it would be useful.

The PVC section is immutable. Any new labels will fail the rolling upgrade.
If these labels are not absolutely necessary, I prefer not to add breaking changes in a minor version of the Helm charts.

sivanov-nuodb: My concern is only with the `use` and `groupPrefix` labels. The rest of the labels and annotations looks good to me.
If we can somehow go without them, we won't need to worry about breaking upgrade.

> What I need to be able to do is find the backup volumes associated with completion of a hotcopy job

We now use a CronJob per backup group. Usually, with just `UNPARTIONED` in use, the `PVC <-> backupGroup` is a one-to-one mapping, however, I think that we shouldn't limit the backup manager to a single SG use case.

`nuobackup` has some reporting machinery that you could use to find which SMs are part of a backup group. Of course, it is available on SMs only but I think that the backup manager has permission to exec into the pods. For example:

```
# Get SMs to backup group mapping
$ k exec -ti sm-database-nuodb-cluster0-demo-hotcopy-0 -- \
  nuobackup --type report-groups --db-name demo
cluster0-0 sm-database-nuodb-cluster0-demo-hotcopy-0

# Find the PVC for each SM
$ k get pod sm-database-nuodb-cluster0-demo-hotcopy-0  \
  -o jsonpath='{.spec.volumes[?(@.name=="backup-volume")].persistentVolumeClaim.claimName}'
backup-volume-sm-database-nuodb-cluster0-demo-hotcopy-0

# Create the VolumeSnapshot resource ...
```

PR #296, opened 2022-06-30
add annotation to job as method to record what was backed up.

butson:
This seems to be all that is needed for the backup manager.   I've added annotations to the jobTemplate spec for backup to record type of backup performed and the backupGroupLabels used.


sivanov-nuodb (approved):
Looks good to me.

PR #301, opened 2022-08-15
Ac/cloudbeaver

acabrele:
Add an incubator helm chart to configure and deploy [Cloudbeaver](https://cloudbeaver.io/) for use with NuoDB.


incubator/cloudbeaver/Chart.yaml
```diff
@@ -0,0 +1,18 @@
...
+- nuodb
+- database
+- sql
+home: https://github.com/nuodb/nuodb-helm-charts/tree/master/incubator/cloudbeaver
+maintainers:
+- name: NuoDB
+  email: support@nuodb.com
+engine: gotpl
+icon: https://raw.githubusercontent.com/nuodb/nuodb-helm-charts/master/images/nuodb.svg
+appVersion: 4.3.2
```
sivanov-nuodb, line 16: We might want to change this to the CloudBeaver version that we tested with. I.e `22.1.4`.

incubator/cloudbeaver/README.md
```diff
@@ -0,0 +1,156 @@
...
+         <file  type="jar"  path="drivers/nuodb"/>
+       </driver>
+       ```
+
+### Deployment
+
+Install the chart with helm, for example:
+
+```sh
+helm upgrade cloudbeaver ~/git/nuodb-helm-charts/incubator/cloudbeaver/ --namespace cloudbeaver --install --create-namespace
```
sivanov-nuodb, line 98: To align with the README.md files for the other charts, we may want to omit the `~/git/nuodb-helm-charts/` prefix.

incubator/cloudbeaver/values.yaml
```diff
@@ -0,0 +1,121 @@
...
+  #These values populate the files in files/cloudbeaver/*
+  #Refer to this page for information on what these do: https://github.com/dbeaver/cloudbeaver/wiki/Server-configuration
+  configs:
+
+    #Optionally the overrides can be disbaled, instead using all cloudbeaver defaults.
+    override: true
+
+    #If any of these sections are not set, Cloudbeaver will take you through the setup on the first start.
+    serverName: Cloudbeaver
+    serverURL: http://localhost:8978
```
sivanov-nuodb, line 37: Should the URL be changed if a loadbalancer service or different port is used?

incubator/cloudbeaver/files/cloudbeaver/initial-data-sources.conf
```diff
@@ -0,0 +1,26 @@
...
+		"nuodb-{{ $initialDataSources.databaseName }}": {
+			"provider": "generic",
+			"driver": "nuodb",
+			"name": "NuoDB: {{ $initialDataSources.adminDomain }}.{{ $initialDataSources.databaseName }}",
+			"save-password": false,
+			"show-system-objects": true,
+			"read-only": false,
+			"template": true,
+			"configuration": {
+				"host": "{{ $initialDataSources.adminDomain }}-clusterip.{{ $initialDataSources.adminNamespace }}.svc",
```
sivanov-nuodb, line 14: Nice!

incubator/cloudbeaver/files/nuodb-config.sh
```diff
@@ -0,0 +1,31 @@
...
+  echo "Downloading NuoDB Driver $NUODB_DRIVER_NAME from maven";
+  wget -O /opt/cloudbeaver/drivers/nuodb/$NUODB_DRIVER_NAME https://repo1.maven.org/maven2/com/nuodb/jdbc/nuodb-jdbc/$NUODB_DRIVER_VERSION/$NUODB_DRIVER_NAME;
+else
+  #Decode it from the configmap
+  echo "Copying NuoDB Driver $NUODB_DRIVER_NAME from configmap";
+  cp /opt/cloudbeaver/config/nuodb/$NUODB_DRIVER_NAME /opt/cloudbeaver/drivers/nuodb/$NUODB_DRIVER_NAME;
+fi;
+
+#Replace the two plugin.xml files containing the additional NuoDB config
+CLOUDBEAVER_BUNDLE_NAME=$(find /opt/cloudbeaver/cloudbeaver-jars/ -name "io.cloudbeaver.resources.drivers.base*");
```
sivanov-nuodb, line 23: Should the script fail if it doesn't find the bundle?

incubator/cloudbeaver/templates/statefulsets.yaml
```diff
@@ -0,0 +1,159 @@
...
+spec:
+  serviceName: cloudbeaver
+  replicas: 1
+  selector:
+    matchLabels:
+      app: cloudbeaver
+  template:
+    metadata:
+      labels:
+        app: cloudbeaver
```
sivanov-nuodb, line 14: I'm not sure if we want multiple installations of this chart. If we do, then we should set the `app` to `{{ include "cloudbeaver.fullname" . }}` instead.

incubator/cloudbeaver/templates/statefulsets.yaml
```diff
@@ -0,0 +1,159 @@
...
+    spec:
+      serviceAccountName: {{ include "cloudbeaver.serviceAccountName" . }}
+      initContainers:
+        - name: source-jars
```
sivanov-nuodb, line 18: Nitpick: Do we need to also set the `workingDir: /opt/cloudbeaver` similar to the second init container?

incubator/cloudbeaver/values.yaml
```diff
@@ -0,0 +1,121 @@
...
+    repository: openjdk
+    tag: 11-jdk
+    pullPolicy: IfNotPresent
+
+cloudbeaver:
+  #Cloudbeaver image details
+  image:
+    registry: docker.io
+    repository: dbeaver/cloudbeaver
+    tag: latest
```
sivanov-nuodb, line 14: I think that we should pin the version of CloudBeaver that we tested with. If you pin the version in `Chart.yaml`, then this becomes `"{{ .Chart.AppVersion }}"`.

incubator/cloudbeaver/templates/statefulsets.yaml
```diff
@@ -0,0 +1,159 @@
...
+kind: StatefulSet
+metadata:
+  name: {{ include "cloudbeaver.fullname" . }}
+spec:
+  serviceName: cloudbeaver
+  replicas: 1
+  selector:
+    matchLabels:
+      app: cloudbeaver
+  template:
```
sivanov-nuodb, line 11: Should we also add annotation with config files checksum to trigger rolling upgrade on configuration changes?

sivanov-nuodb (approved):
Excellent work! Thanks for making NuoDB database connectivity event easier.

incubator/cloudbeaver/Chart.yaml
acabrele: Good spot, copy-n-paste :-). Fixed.

incubator/cloudbeaver/templates/statefulsets.yaml
acabrele: Since this one uses the cloudbeaver image, the working dir is already /opt/cloudbeaver as set in their docker file - but it could be added to overcome any changes that might happen in the cloudbeaver image.

incubator/cloudbeaver/templates/statefulsets.yaml
acabrele: I thought about doing this, but the thing is that Cloudbeaver only uses these configs on first start.  Later changes won't have any effect.  

That said... I did in the end add the option to _not_ persist configuration, in which case this could be useful.  I'll also add commentary to that effect.

incubator/cloudbeaver/README.md
acabrele: Done!

incubator/cloudbeaver/values.yaml
acabrele: As discussed, removed this from being configurable. As per Cloudbeaver docs it is only used for 3rd party integration links.

incubator/cloudbeaver/files/nuodb-config.sh
acabrele: Added check/fail for no find result

incubator/cloudbeaver/templates/statefulsets.yaml
acabrele: Done!

incubator/cloudbeaver/templates/statefulsets.yaml
acabrele: Done!

incubator/cloudbeaver/values.yaml
acabrele: Done!

incubator/cloudbeaver/templates/statefulsets.yaml
acabrele: Done!

incubator/cloudbeaver/templates/services.yaml
```diff
@@ -0,0 +1,17 @@
...
+{{- end }}
+spec:
+  type: {{ .Values.cloudbeaver.service.type }}
+  ports:
+    - port: {{ .Values.cloudbeaver.service.port }}
+      targetPort: 8978
+      protocol: TCP
+      name: http
+  selector:
+    app: cloudbeaver
```
sivanov-nuodb, line 17: This should also change to `app: {{ include "cloudbeaver.fullname" . }}` right?

adriansuarez (approved):
Looks good.

PR #307, opened 2023-01-30
Add database resync multi-cluster test

sivanov-nuodb:
**Testing**

- Added conditional sub-test which checks the automatic database removal in the existing E2E multi-cluster test. 
- Updated the database testing utilities to work in cases where the database Helm release has been removed already.


test/multicluster/minikube_base_multicluster_test.go
```diff
@@ -134,4 +140,73 @@ func TestKubernetesBasicMultiCluster(t *testing.T) {
 			verifyNuoSQL(t, context.Namespace, admin0, "demo")
 		})
 	})
+
+	testlib.RunOnNuoDBVersionCondition(t, ">=5.0.2", func(version *semver.Version) {
```
sivanov-nuodb, line 144: I'll update the version to `5.0.3` in case the feature doesn't make it for NuoDB 5.0.2.

kontaras (approved):

test/multicluster/minikube_base_multicluster_test.go
```diff
@@ -134,4 +140,73 @@ func TestKubernetesBasicMultiCluster(t *testing.T) {
...
+			)
+			// negative test: check that the database is not deleted
+			testlib.ExecuteInAllClusters(t, func(context *testlib.ClusterDeploymentContext) {
+				admin0 := fmt.Sprintf("%s-nuodb-%s-0", context.AdminReleaseName, context.ThisCluster.Name)
+				testlib.AwaitDatabaseUp(t, context.Namespace, admin0, "demo", 2)
+				testlib.CheckArchives(t, namespaceName, admin0, "demo", 1, 0)
+				db, err := testlib.GetDatabaseE(t, namespaceName, admin0, "demo")
+				require.NoError(t, err)
+				require.NotEqual(t, "TOMBSTONE", db.State)
+			})
```
adriansuarez, line 165: Is this actually a negative test? I consider a negative test to be checking that bad input is correctly rejected.

test/testlib/nuodb_database_utilities.go
```diff
@@ -190,9 +199,13 @@ func StartDatabaseTemplate(t *testing.T, namespaceName string, adminPod string,
 		if opt.DbPrimaryRelease {
 			smPodName0 := GetPodName(t, namespaceName, smPodNameTemplate)
 			AddTeardown(TEARDOWN_DATABASE, func() {
-				for _, smPod := range GetPodNames(t, namespaceName, smPodNameTemplate) {
-					t.Logf("Getting log from SM pod: %s", smPod)
-					go GetAppLog(t, namespaceName, smPod, "", &v12.PodLogOptions{Follow: true})
+				pods, err := findPods(t, namespaceName, smPodNameTemplate)
+				if err != nil {
```
adriansuarez, line 203: It looks like `findPods()` just returns non-`nil` error when the list of pods is empty. Wouldn't the subsequent loop just do nothing in that case?

If there are any pods that are somehow returned by `findPods()`, we would want to collect their logging regardless of the error, right?

adriansuarez (approved):
Looks good to me. I left a couple of questions in the review.

test/multicluster/minikube_base_multicluster_test.go
sivanov-nuodb: I called it "negative" because the database should not be removed automatically as not all conditions are met. In this case, the domain state is the input so according to your description it is not testing a "bad" input.

test/testlib/nuodb_database_utilities.go
sivanov-nuodb: I haven't checked the `findPods` implementation in detail but used the usual _go_ error handling which guards against returned `nil` slice. As you pointed out, we could omit the error handling in this specific case.

PR #310, opened 2023-02-21
Update TestAdminScaleDown to delete PVC before checking for Raft leader

adriansuarez:
Deletion of the PVC for the scaled down APs will become a requirement for future versions of the product. This change updates TestAdminScaleDown so that it works with the new behavior, and also the existing behavior, which applies the consensus override immediately after the pod is deleted after being scaled down.

### DIFF
--- a/test/minikube/minikube_base_database_test.go
+++ b/test/minikube/minikube_base_database_test.go
@@ -338,7 +338,7 @@ func TestKubernetesStartDatabaseShrinkedAdmin(t *testing.T) {
 	k8s.RunKubectl(t, kubectlOptions, "scale", "statefulset", admin, "--replicas=2")
 	admin2 := admin + "-2"
 	testlib.AwaitServerState(
-		t, namespaceName, admin0, admin2, "Disconnected", true, 60*time.Second)
+		t, namespaceName, admin0, admin2, "Disconnected", 60*time.Second)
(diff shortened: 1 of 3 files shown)


kontaras (approved):

test/minikube/minikube_base_database_test.go
```diff
@@ -334,10 +334,13 @@ func TestKubernetesStartDatabaseShrinkedAdmin(t *testing.T) {
 		testlib.DescribePods(t, namespaceName, admin)
 	})
 
-	// scale down the APs to 2; KAA will automatically evict the shut down AP
+	// scale down the APs to 2 and delete PVC for scaled down AP; KAA will automatically shrink membership to 2
 	k8s.RunKubectl(t, kubectlOptions, "scale", "statefulset", admin, "--replicas=2")
+	admin2 := admin + "-2"
 	testlib.AwaitServerState(
-		t, namespaceName, admin0, fmt.Sprintf("%s-2", admin), "Disconnected", true, 60*time.Second)
+		t, namespaceName, admin0, admin2, "Disconnected", 60*time.Second)
```
sivanov-nuodb, line 341: That way the test coverage for the previous behavior is lost.
I guess it's fine as we don't expect users to scale down the AP just to override the consensus.

sivanov-nuodb (approved):
Looks good. Thanks for updating the tests!

PR #311, opened 2023-02-28
Incremental backup after SM restart fails if the previous full was unsuccessful

sivanov-nuodb:
**Issue**

If the _journal_ hot copy is temporarily disabled due to SM sync, an incremental backup will be requested. The requested incremental will fail if the previous _full_ backup was not successful.

**Changes**

Refactor the `nuobackup` script and create a function for each backup type. This makes it easy to understand the prerequisites and post-failure actions for each hot copy type.

**Testing**

- changed the `TestKubernetesJournalBackupSuspended` E2E test so that the scenario is covered.
- added `testlib.BackupDatabaseE` function which doesn't fail the test if the requested database backup fails

### DIFF
--- a/stable/database/files/nuobackup
+++ b/stable/database/files/nuobackup
@@ -157,7 +157,7 @@ function full_exist() {
       --labels "${labels}" 2>&1)"
    if [ $? -ne 0 ] && echo "$error" | grep -q "No existing backup path found"; then
       # Check if full backup has been started already for all archiveIds. It
-      # won't be the case in the following scenarious:
+      # won't be the case in the following scenarios:
       # * initial database backup request
       # * backup request after the database have been restored
       # * new HC SM has been started in the backup group since the last full
--- a/stable/database/files/nuobackup
+++ b/stable/database/files/nuobackup
@@ -42,6 +42,7 @@ labels=""
 timeout=""
 backup_root=$BACKUP_DIR
 retry_failed="True"
+backup_database_output=""
 
 function get_current_backup() {
    output=$( nuodocker get current-backup \
@@ -56,7 +57,6 @@ function execute_backup() {
    local current
    local next
    local backupset
-   local output
    local failed
    local pods
 
@@ -65,7 +65,7 @@ function execute_backup() {
 
    # call nuodocker to perform the actual backup
    echo "Starting ${type} backup for database ${db_name} on processes with labels '${labels}' ..."
-   output=$(nuodocker backup database \
+   backup_database_output=$(nuodocker backup database \
       --db-name "${db_name}" \
       --type "${type}" \
       --backup-root "${backup_root}" \
@@ -73,7 +73,7 @@ function execute_backup() {
       "${extra_flags[@]}" 2>&1)
 
    retval=$?
-   echo "$output"
+   echo "$backup_database_output"
    if [ $retval != 0 ]; then
       echo >&2 "Error running hotcopy $retval"
       # Store failed backupset in KV store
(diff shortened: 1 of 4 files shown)


stable/database/files/nuobackup
```diff
@@ -161,6 +149,86 @@ function execute_backup() {
    fi
 }
 
+function full_exist() {
+   local error
+
+   error="$( nuodocker get current-backup \
+      --db-name "${db_name}" \
+      --labels "${labels}" 2>&1)"
```
bkelly-ndb, line 157: Will labels always be passed into this script? I assume so but just checking since this could this fail if they are not

test/testlib/minikube_utilities.go
```diff
@@ -1313,6 +1313,12 @@ func ScaleStatefulSet(t *testing.T, namespaceName string, name string, replicas
 	k8s.RunKubectl(t, options, "scale", "statefulset", name, fmt.Sprintf("--replicas=%d", replicas))
 }
 
+func ScaleDeployment(t *testing.T, namespaceName string, name string, replicas int) {
```
kontaras, line 1316: Nit: variable `name` is a bit vague for someone trying to understand the function from the signature. Maybe rename it to `deploymentName`?

kontaras (approved):

stable/database/files/nuobackup
sivanov-nuodb: The script defaults the labels for `role hotcopy` if none are provided. Please see line 268.

test/testlib/minikube_utilities.go
sivanov-nuodb: I presumed that as the function is called `ScaleDeployment`, then it is clear that the parameter `name` corresponds to the deployment name. The `ScaleStatefulSet` is having the same signature.

bkelly-ndb (approved):

stable/database/files/nuobackup
```diff
@@ -161,6 +149,86 @@ function execute_backup() {
...
+   if [ $? -ne 0 ] && echo "$error" | grep -q "No existing backup path found"; then
+      # Check if full backup has been started already for all archiveIds. It
+      # won't be the case in the following scenarious:
```
adriansuarez, line 160: Nitpick: Misspelled "scenarios".

stable/database/files/nuobackup
```diff
@@ -161,6 +149,86 @@ function execute_backup() {
...
+      execute_full
+   else 
+      execute_backup "journal"
+      rc=$?
+      if [ $rc -ne 0 ]; then
+         if echo "$backup_database_output" | grep -q "Full or incremental hot copy must be completed before journal hot copy"; then
+            # journal hot copy has been temporarly suspended due to archive sync;
+            # perform incremental hot copy first
+            echo "Executing incremental hot copy as journal hot copy is temporarily suspended"
+            execute_incremental || return $?
```
adriansuarez, line 222: Will journal hotcopy fail even if there is a full, because there are gaps in the journal (e.g. because all SMs were restarted)? If so, then it looks like you are relying on incremental to cover that gap before performing journal hotcopy. Is that correct?

test/testlib/nuodb_database_utilities.go
```diff
@@ -189,8 +186,11 @@ func StartDatabaseTemplate(t *testing.T, namespaceName string, adminPod string,
 		})
 		// the TEs will become RUNNING after the SMs as they need an entry node
 		// so use the same ready timeout for both
-		readyTimeout := AdjustPodTimeout(tePodName, 300*time.Second)
-		AwaitPodUp(t, namespaceName, tePodName, readyTimeout)
+		readyTimeout := AdjustPodTimeout(tePodNameTemplate, 300*time.Second)
```
adriansuarez, line 189: Why don't we move this into the `if`-statement?

adriansuarez (approved):
Looks good.

test/testlib/nuodb_database_utilities.go
sivanov-nuodb: The same variable is used to wait for the SMs in another _if_ block later (please see line 205).

stable/database/files/nuobackup
sivanov-nuodb: > Will journal hotcopy fail even if there is a full, because there are gaps in the journal (e.g. because all SMs were restarted)?

If the _full_ has been done after the restarted SM finished syncing, then we won't get into this _if_ block.

If some of the SMs in the group have been restarted and went to syncing, then the journal hot copy for all SMs in the group will be temporarily suspended. A _full_ or _incremental_ hot copy must be requested to form a new baseline for subsequent _journal_ hot hopy.

PR #314, opened 2023-03-13
Storage groups support

sivanov-nuodb:
**Changes**

- users can now enable Table Partitions and Storage Groups (TPSG) and supply SG name (by default the Helm release name) when installing the _database_ Helm chart. This requires NuoDB v5.0.3 or later.
- A one-to-one mapping between Helm release and NuoDB storage group simplifies deployment and operations. Storage group reconfiguration is not supported.
- Additional storage groups are added by installing the _database_ chart multiple times with `database.primaryRelease=false` option.
- added `database.te.enablePod` option which can be used to disable TE Deployment when additional storage groups are added.
- when TPSG is enabled the SMs serving storage group will be labeled with `sg <sgName>`.
- fix an issue in `nuobackup` where new backup groups doesn't get registered if it is a substring of an existing one

**Provision Database With Storage Groups (example)**

The steps below will deploy NuoDB database with 3 storage groups. Each SG will be served by 2 SMs, one regular and one HCSM enabling backup.

> **Note** It is recommended that storage groups are installed with `database.primaryRelease=false` option. This allows the user to remove any storage group without removing resources installed by the primary database release.

Install the primary _database_ release. In this example, it's used to deploy the TEs, however, several [TE groups](https://github.com/nuodb/nuodb-helm-charts/blob/master/docs/HowToConnectExternally.md#transaction-engine-groups) can be enabled as well.

```
helm upgrade --install database stable/database \
  --set database.sm.hotCopy.replicas=0 \
  --set database.sm.noHotCopy.replicas=0 \
  --set database.sm.backupGroups.main-cluster.labels="role hotcopy" \
  -f values.yaml
```

Install a secondary _database_ release per storage group.

```sh
helm upgrade --install sg1 stable/database \
  --set database.sm.storageGroup.enabled=true \
  --set database.primaryRelease=false \
  --set database.sm.noHotCopy.replicas=1 \
  --set database.te.enablePod=false \
  -f values.yaml

helm upgrade --install sg2 stable/database \
  --set database.sm.storageGroup.enabled=true \
  --set database.primaryRelease=false \
  --set database.sm.noHotCopy.replicas=1 \
  --set database.te.enablePod=false \
  -f values.yaml

helm upgrade --install sg3 stable/database \
  --set database.sm.storageGroup.enabled=true \
  --set database.primaryRelease=false \
  --set database.sm.noHotCopy.replicas=1 \
  --set database.te.enablePod=false \
  -f values.yaml
```

Check the database storage groups.

```sh
$ nuocmd get storage-groups --db-name demo
StorageGroup(archive_states={}, db_name=demo, id=1, leader_candidates=[], name=ALL, process_states={}, state=Available)
StorageGroup(archive_states={0: ADDED, 1: ADDED}, db_name=demo, id=10, leader_candidates=[0, 2], name=SG1, process_states={0: RUNNING, 2: RUNNING}, state=Available)
StorageGroup(archive_states={2: ADDED, 3: ADDED}, db_name=demo, id=11, leader_candidates=[1, 4], name=SG2, process_states={1: RUNNING, 4: RUNNING}, state=Available)
StorageGroup(archive_states={4: ADDED, 5: ADDED}, db_name=demo, id=12, leader_candidates=[3, 5], name=SG3, process_states={3: RUNNING, 5: RUNNING}, state=Available)
StorageGroup(archive_states={}, db_name=demo, id=2, leader_candidates=[0, 1, 2, 3, 4, 5], name=UNPARTITIONED, process_states={0: RUNNING, 1: RUNNING, 2: RUNNING, 3: RUNNING, 4: RUNNING, 5: RUNNING}, state=Available)
```

Spread table partitions over all SGs.

```sql
CREATE TABLE codes ( sg char(4), zip char(5) ) PARTITION BY RANGE (zip) ( 
	PARTITION p_sg1 VALUES LESS THAN ('2000') STORE IN SG1 
	PARTITION p_sg2 VALUES LESS THAN ('3000') STORE IN SG2 
	PARTITION p_sg3 VALUES LESS THAN ('4000') STORE IN SG3;

INSERT INTO codes VALUES ('sg1', '1001'), ('sg2', '2001'), ('sg3', '3001');
```

Check the row count in each partition.

```sql
SELECT 
	pids.PARTITIONNAME, 
	pids.STORAGEGROUP, 
	count(0) as COUNT 
FROM 
	codes c 
		INNER JOIN system.PARTITIONIDS pids 
			ON c._record_partitionid = pids.PARTITIONID 
GROUP BY pids.PARTITIONNAME, pids.STORAGEGROUP;
```

```
 PARTITIONNAME  STORAGEGROUP  COUNT
 -------------- ------------- ------

     P_SG1           SG1        1
     P_SG2           SG2        1
     P_SG3           SG3        1
```

**Delete Storage Group (example)**

Delete the partitions stored in SG2.

```sql
ALTER TABLE codes DROP PARTITION p_sg2;
```

Delete the corresponding Helm release.

```sh
helm delete sg2
```

Delete the storage group.

```sh
nuocmd delete storage-group --db-name demo --sg-name sg2
```

**Testing**

- integration tests for the new helper templates
- update the `TestKubernetesRestoreWithStorageGroups` test and use the new options to configure storage groups

### DIFF
--- a/stable/database/README.md
+++ b/stable/database/README.md
@@ -270,6 +270,9 @@ The following tables list the configurable parameters of the `database` chart an
 | `sm.tolerations` | Tolerations for NuoDB SM | `[]` |
 | `sm.topologySpreadConstraints` | Topology spread constraints for NuoDB SM | `[]` |
 | `sm.readinessTimeoutSeconds` | SM readiness probe timeout, sometimes needs adjusting depending on environment and pod resources | `5` |
+| `sm.storageGroup.enabled` | Enable Table Partitions and Storage Groups (TPSG) for all SMs in this database Helm release | `false` |
+| `sm.storageGroup.name` | The name of the storage group. Only alphanumeric and underscore ('_') characters are allowed. By default the Helm release name is used | `.Release.Name` |
+| `te.enablePod` | Create deployment for TEs | `true` |
 | `te.externalAccess.enabled` | Whether to deploy a Layer 4 service for the database | `false` |
 | `te.externalAccess.internalIP` | Whether to use an internal (to the cloud) or external (public) IP address for the load balancer. Only applies to external access of type `LoadBalancer` | `nil` |
 | `te.externalAccess.type` | The service type used to enable external database access. The supported types are `NodePort` and `LoadBalancer` (defaults to `LoadBalancer`) | `nil` |
--- a/stable/database/README.md
+++ b/stable/database/README.md
@@ -272,7 +272,7 @@ The following tables list the configurable parameters of the `database` chart an
 | `sm.readinessTimeoutSeconds` | SM readiness probe timeout, sometimes needs adjusting depending on environment and pod resources | `5` |
 | `sm.storageGroup.enabled` | Enable Table Partitions and Storage Groups (TPSG) for all SMs in this database Helm release | `false` |
 | `sm.storageGroup.name` | The name of the storage group. Only alphanumeric and underscore ('_') characters are allowed. By default the Helm release name is used | `.Release.Name` |
-| `te.enablePod` | Create deployment for TEs | `true` |
+| `te.enablePod` | Create deployment for TEs | `nil` |
 | `te.externalAccess.enabled` | Whether to deploy a Layer 4 service for the database | `false` |
 | `te.externalAccess.internalIP` | Whether to use an internal (to the cloud) or external (public) IP address for the load balancer. Only applies to external access of type `LoadBalancer` | `nil` |
 | `te.externalAccess.type` | The service type used to enable external database access. The supported types are `NodePort` and `LoadBalancer` (defaults to `LoadBalancer`) | `nil` |
--- a/stable/database/README.md
+++ b/stable/database/README.md
@@ -227,7 +227,7 @@ The following tables list the configurable parameters of the `database` chart an
 | `sm.logPersistence.size` | Amount of disk space allocated for log storage | `60Gi` |
 | `sm.logPersistence.storageClass` | Storage class for volume backing log storage.  This storage class must be pre-configured in the cluster | `-` |
 | `sm.hotCopy.replicas` | SM replicas with hot-copy enabled | `1` |
-| `sm.hotCopy.enablePod` | Create DS/SS for hot-copy enabled SMs | `true` |
+| `sm.hotCopy.enablePod` | Create StatefulSet for hot-copy enabled SMs | `true` |
 | `sm.hotCopy.enableBackups` | Enable full/incremental/journal backups | `true` |
 | `sm.hotCopy.deadline` | Deadline for a hotcopy job to start (seconds) | `1800` |
 | `sm.hotCopy.timeout` | Timeout for a started `full` or `incremental` hotcopy job to complete (seconds). The default timeout of "0" will force the backup jobs to wait forever for the requested hotcopy operation to complete | `0` |
@@ -257,7 +257,7 @@ The following tables list the configurable parameters of the `database` chart an
 | `sm.hotCopy.journalPath.size` | Amount of disk space allocated for SM journal | `20Gi` |
 | `sm.hotCopy.journalPath.storageClass` | Storage class for SM journal.  This storage class must be pre-configured in the cluster | `-` |
 | `sm.noHotCopy.replicas` | SM replicas with hot-copy disabled | `0` |
-| `sm.noHotCopy.enablePod` | Create DS/SS for non-hot-copy SMs | `true` |
+| `sm.noHotCopy.enablePod` | Create StatefulSet for non-hot-copy SMs | `true` |
 | `sm.noHotCopy.journalPath.enabled` | Whether to enable separate SM journal directory. For more info, read the [Journal HowTo](../../docs/HowToArchiveJournal.md) | `false` |
 | `sm.noHotCopy.journalPath.accessModes` | Volume access modes enabled (must match capabilities of the storage class) | `ReadWriteOnce` |
 | `sm.noHotCopy.journalPath.size` | Amount of disk space allocated for SM journal | `20Gi` |
--- a/stable/database/README.md
+++ b/stable/database/README.md
@@ -272,7 +272,7 @@ The following tables list the configurable parameters of the `database` chart an
 | `sm.readinessTimeoutSeconds` | SM readiness probe timeout, sometimes needs adjusting depending on environment and pod resources | `5` |
 | `sm.storageGroup.enabled` | Enable Table Partitions and Storage Groups (TPSG) for all SMs in this database Helm release | `false` |
 | `sm.storageGroup.name` | The name of the storage group. Only alphanumeric and underscore ('_') characters are allowed. By default the Helm release name is used | `.Release.Name` |
-| `te.enablePod` | Create deployment for TEs | `nil` |
+| `te.enablePod` | Create deployment for TEs. By default, the TE Deployment is disabled if TP/SG is enabled and this is a "secondary" release. | `nil` |
 | `te.externalAccess.enabled` | Whether to deploy a Layer 4 service for the database | `false` |
 | `te.externalAccess.internalIP` | Whether to use an internal (to the cloud) or external (public) IP address for the load balancer. Only applies to external access of type `LoadBalancer` | `nil` |
 | `te.externalAccess.type` | The service type used to enable external database access. The supported types are `NodePort` and `LoadBalancer` (defaults to `LoadBalancer`) | `nil` |
--- a/stable/database/templates/_helpers.tpl
+++ b/stable/database/templates/_helpers.tpl
@@ -509,6 +509,9 @@ Renders the storage groups option for nuosm script
 {{- if has (upper .) $invalid }}
 {{- fail (printf "Invalid storage group name: %s" .) }}
 {{- end }}
+{{- if contains " " (trim .) }}
+{{- fail (printf "Multiple storage group names provided: %s" .) }}
+{{- end }}
 {{- end }}
 - "--storage-groups"
 - {{ .Values.database.sm.storageGroup.name | default .Release.Name | trim | quote }}
--- a/stable/database/templates/_helpers.tpl
+++ b/stable/database/templates/_helpers.tpl
@@ -497,3 +497,32 @@ Renders the name of the Secret for this database
 {{- define "database.secretName" -}}
 {{ .Values.admin.domain }}-{{ .Values.database.name }}
 {{- end -}}
+
+{{/*
+Renders the storage groups option for nuosm script
+*/}}
+{{- define "database.storageGroup.args" -}}
+{{- if .Values.database.sm.storageGroup }}
+{{- if .Values.database.sm.storageGroup.enabled }}
+{{- with .Values.database.sm.storageGroup.name }}
+{{- $invalid := list "ALL" "UNPARTITIONED" }}
+{{- if has (upper .) $invalid }}
+{{- fail (printf "Invalid storage group name: %s" .) }}
+{{- end }}
+{{- end }}
+- "--storage-groups"
+- {{ .Values.database.sm.storageGroup.name | default .Release.Name | trim | quote }}
+{{- end }}
+{{- end }}
+{{- end -}}
+
+{{/*
+Renders the storage group domain process label
+*/}}
+{{- define "database.storageGroup.label" -}}
+{{- if .Values.database.sm.storageGroup }}
+{{- if .Values.database.sm.storageGroup.enabled }}
+{{- printf "sg %s" (.Values.database.sm.storageGroup.name | default .Release.Name | trim) -}}
+{{- end }}
+{{- end }}
+{{- end -}}
--- a/stable/database/templates/_helpers.tpl
+++ b/stable/database/templates/_helpers.tpl
@@ -529,3 +529,19 @@ Renders the storage group domain process label
 {{- end }}
 {{- end }}
 {{- end -}}
+
+{{/*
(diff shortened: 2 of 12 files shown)


test/integration/template_database_test.go
```diff
@@ -1787,3 +1800,126 @@ func TestDatabaseTopologyConstraints(t *testing.T) {
...
+	})
+
+	t.Run("testStorageGroupWithName", func(t *testing.T) {
+		// Path to the helm chart we will test
+		helmChartPath := testlib.DATABASE_HELM_CHART_PATH
+		options := &helm.Options{
+			SetValues: map[string]string{
+				"database.sm.nohotCopy.replicas":   "1",
+				"database.sm.storageGroup.enabled": "true",
+				"database.sm.storageGroup.name":    "SG1",
```
kontaras, line 1832: Should you pick a SG name that is not the default?

test/testlib/nuodb_database_utilities.go
```diff
@@ -425,6 +436,27 @@ func CheckArchives(t *testing.T, namespaceName string, adminPod string, dbName s
...
 }
 
+func CheckStorageGroup(t *testing.T, namespaceName, adminPod, dbName, sgName, expectedState string) *NuoDBStorageGroup {
+	options := k8s.NewKubectlOptions("", "", namespaceName)
+	output, err := k8s.RunKubectlAndGetOutputE(t, options, "exec", adminPod, "-c", "admin", "--",
+		"nuocmd", "--show-json", "get", "storage-groups", "--db-name", dbName)
+	require.NoError(t, err, output)
+	err, sgs := UnmarshalStorageGroups(output)
+	require.NoError(t, err)
+	// the admin convert all storage group names to upper case
```
kontaras, line 446: Nit: "convert" -> "converts"

kontaras (approved):

test/integration/template_database_test.go
sivanov-nuodb: The default SG name will be the Helm release name which is `release-name` in this test.

stable/database/README.md
```diff
@@ -270,6 +270,9 @@ The following tables list the configurable parameters of the `database` chart an
 | `sm.tolerations` | Tolerations for NuoDB SM | `[]` |
 | `sm.topologySpreadConstraints` | Topology spread constraints for NuoDB SM | `[]` |
 | `sm.readinessTimeoutSeconds` | SM readiness probe timeout, sometimes needs adjusting depending on environment and pod resources | `5` |
+| `sm.storageGroup.enabled` | Enable Table Partitions and Storage Groups (TPSG) for all SMs in this database Helm release | `false` |
+| `sm.storageGroup.name` | The name of the storage group. Only alphanumeric and underscore ('_') characters are allowed. By default the Helm release name is used | `.Release.Name` |
+| `te.enablePod` | Create deployment for TEs | `true` |
```
adriansuarez, line 275: `enablePod` sounds like a weird name for the value, since this is actually enabling/disabling the TE deployment. Creation of TE pods can also be inhibited by setting `te.replicas=0`, so `enableDeployment` sounds more descriptive.

I guess this is consistent with `enablePod` which already exists for SMs. Also, I noticed that the `sm.*.enablePod` rows both mention DaemonSets, which we no longer support.

stable/database/README.md
```diff
@@ -270,6 +270,9 @@ The following tables list the configurable parameters of the `database` chart an
 | `sm.tolerations` | Tolerations for NuoDB SM | `[]` |
 | `sm.topologySpreadConstraints` | Topology spread constraints for NuoDB SM | `[]` |
 | `sm.readinessTimeoutSeconds` | SM readiness probe timeout, sometimes needs adjusting depending on environment and pod resources | `5` |
+| `sm.storageGroup.enabled` | Enable Table Partitions and Storage Groups (TPSG) for all SMs in this database Helm release | `false` |
```
adriansuarez, line 273: Would it be better if the `te.enablePod` default value was `nil` and we would switch it on if (1) `sm.storageGroup.enabled=true` and (2) `primaryRelease=false`?

stable/database/templates/_helpers.tpl
```diff
@@ -497,3 +497,32 @@ Renders the name of the Secret for this database
...
+{{- if .Values.database.sm.storageGroup }}
+{{- if .Values.database.sm.storageGroup.enabled }}
+{{- with .Values.database.sm.storageGroup.name }}
+{{- $invalid := list "ALL" "UNPARTITIONED" }}
+{{- if has (upper .) $invalid }}
+{{- fail (printf "Invalid storage group name: %s" .) }}
+{{- end }}
+{{- end }}
+- "--storage-groups"
+- {{ .Values.database.sm.storageGroup.name | default .Release.Name | trim | quote }}
```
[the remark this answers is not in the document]
adriansuarez: I guess this is supposed to accept a particular storage group rather than comma-separated list.

test/minikube/minikube_long_restore_test.go
```diff
@@ -333,7 +333,7 @@ func TestKubernetesRestoreWithStorageGroups(t *testing.T) {
 	if os.Getenv("NUODB_LICENSE") != "ENTERPRISE" && os.Getenv("NUODB_LICENSE_CONTENT") == "" {
```
adriansuarez, line 333: Didn't realize we had coverage of SGs in these tests. I guess this was to exercise restore specifically, which was designed to work with SGs.

test/minikube/minikube_long_restore_test.go
```diff
@@ -351,38 +351,64 @@ func TestKubernetesRestoreWithStorageGroups(t *testing.T) {
...
+	databaseOptions.SetValues["database.primaryRelease"] = "false"
+	databaseOptions.SetValues["database.te.enablePod"] = "false"
+	databaseOptions.SetValues["database.sm.storageGroup.name"] = "sg1"
+	sg1ReleaseName := testlib.StartDatabase(t, namespaceName, admin0, &databaseOptions)
+
+	// set the total number of engines as in the test utilities they are
+	// inferred from the values
+	databaseOptions.SetValues["database.te.enablePod"] = "true"
+	databaseOptions.SetValues["database.te.replicas"] = "1"
+	databaseOptions.SetValues["database.sm.hotCopy.replicas"] = "2"
```
adriansuarez, line 384: What is this for? It looks like we're updating the Helm values after we have done the deployment. Is this to prevent checks from breaking that inspect the Helm values?

test/minikube/minikube_long_restore_test.go
```diff
@@ -351,38 +351,64 @@ func TestKubernetesRestoreWithStorageGroups(t *testing.T) {
...
+	// install database secondary release which serves sg1
+	databaseOptions.SetValues["database.primaryRelease"] = "false"
+	databaseOptions.SetValues["database.te.enablePod"] = "false"
+	databaseOptions.SetValues["database.sm.storageGroup.name"] = "sg1"
+	sg1ReleaseName := testlib.StartDatabase(t, namespaceName, admin0, &databaseOptions)
+
+	// set the total number of engines as in the test utilities they are
+	// inferred from the values
+	databaseOptions.SetValues["database.te.enablePod"] = "true"
+	databaseOptions.SetValues["database.te.replicas"] = "1"
```
adriansuarez, line 383: This didn't change, right?

adriansuarez (approved):
Looks good to me.

I made a couple of minor comments having to do with the Helm values. I also think the test needs to be updated to wait for SGs to be added to the SMs.

stable/database/README.md
sivanov-nuodb: > I guess this is consistent with enablePod which already exists for SMs

Right. I personally prefer `enabled` but decided to stay consistent.

> Also, I noticed that the sm.*.enablePod rows both mention DaemonSets (DS), which we no longer support

Thanks for pointing this out. I will fix it.

test/minikube/minikube_long_restore_test.go
sivanov-nuodb: Right. The test was setting up SGs using `nuocmd` and not it's more declarative.

test/minikube/minikube_long_restore_test.go
sivanov-nuodb: Right. I just put it for completeness.

test/minikube/minikube_long_restore_test.go
```diff
@@ -351,38 +351,64 @@ func TestKubernetesRestoreWithStorageGroups(t *testing.T) {
...
+		require.GreaterOrEqual(t, len(sg.LeaderCandidates), 1)
+		for archiveId, state := range sg.ArchiveStates {
+			require.Equal(t, "ADDED", state, "Unexpected state for archiveId=%s", archiveId)
+		}
+		for nodeId, state := range sg.ProcessStates {
+			require.Equal(t, "RUNNING", state, "Unexpected state for nodeId=%s", nodeId)
+		}
+	}
+	verifyStorageGroup("sg0")
+	verifyStorageGroup("sg1")
```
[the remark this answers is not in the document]
sivanov-nuodb: Good point! Will do that.

stable/database/templates/_helpers.tpl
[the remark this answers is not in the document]
sivanov-nuodb: This is a good point! The product supports passing multiple SGs as space space-separated list but we would like to avoid this. We want to enforce one-to-one mapping between Helm release and NuoDB storage group which simplifies deployment and operations. I will add a check for not allowing spaces.

stable/database/README.md
sivanov-nuodb: I'm not sure yet what NuoDB database deployment will look like for the scale-out use case. It may be that you would want to have new TE + SM per SG. This is why I didn't want to code assumptions for `database.te.enablePod` value but rather leave this to the user.

Your proposal will work but `database.te.enablePod=nil` is not consistent with other `enablePod` values and I'm not sure how to set `nil` value for a property in the `values.yaml` (maybe `prop=null`).

adriansuarez: I meant just commenting out the property in the Helm values, which should allow us to detect that it was explicitly specified.

sivanov-nuodb: I've implemented this in [e00b6ff](https://github.com/nuodb/nuodb-helm-charts/pull/314/commits/e00b6ff02748525822c723ea715189344c30efdc). It's a bit tricky to check for `nil` though.

stable/database/templates/_helpers.tpl
```diff
@@ -529,3 +529,19 @@ Renders the storage group domain process label
 {{- end }}
 {{- end }}
 {{- end -}}
+
+{{/*
+Checks if the TE Deployment should be enabled. If the default value is absent,
+the TE Deployment is disabled if TPSG is enabled and this is a secondary release.
+*/}}
+{{- define "database.te.enablePod" -}}
+  {{- if kindIs "invalid" .Values.database.te.enablePod -}}
```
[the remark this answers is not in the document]
sivanov-nuodb: This is the strange way to check for `nil`. If bogus value is supplied, then the below error will be shown.
I will add a test for this.

```
 Error: template: database/templates/deployment.yaml:1:11: executing "database/templates/deployment.yaml" at <include "database.te.enablePod" .>: error calling include: template: database/templates/_helpers.tpl:545:8: executing "database.te.enablePod" at <include "defaulttrue" .Values.database.te.enablePod>: error calling include: template: database/templates/_helpers.tpl:299:4: executing "validate.boolString" at <fail (printf "Invalid boolean value: %!s(MISSING)" .)>: error calling fail: Invalid boolean value: foo
```

test/integration/template_database_test.go
```diff
@@ -328,8 +328,39 @@ func TestDatabaseDeploymentRenders(t *testing.T) {
...
+		}
+
+		// Run RenderTemplate to render the template and capture the output.
+		output := helm.RenderTemplate(t, options, helmChartPath, "release-name", []string{"templates/deployment.yaml"})
+
+		for _, obj := range testlib.SplitAndRenderDeployment(t, output, 1) {
+			assert.Equal(t, "te", obj.Spec.Selector.MatchLabels["component"])
+			assert.Equal(t, "te", obj.Spec.Template.ObjectMeta.Labels["component"])
+		}
+	})
```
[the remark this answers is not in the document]
sivanov-nuodb: I can add a test for completeness.

stable/database/values.yaml
```diff
@@ -430,7 +430,25 @@ database:
...
+    # time. The SM will emit warning messages if there is a mismatch between the
+    # initially configured SG and the currently served SGs.
+    storageGroup: 
+      enabled: false
+      # The name of the storage group. Only alphanumeric and underscore ('_')
+      # characters are allowed. By default the Helm release name is used.
+      name: ""
+
   te:
+    # Provision Transaction Engine (TE) Deployment.
```
acabrele, line 449: A note about how this behaves with SGs / secondary deployment might be useful here, as well as / instead of in the helpers file where users are less likely to look.

acabrele:
Surprisingly efficient change enabled by the previous work of primary / secondary deployments!  

As per our slack chat, one area that I think needs some documentation is on how backup and restore work in a multi-SG deployment, specifically around the admin validation for all SGs being selected, what is and isn't supported, co-ordinated backup etc.  Perhaps a future PR can extend on docs/BackupAndRestore.md

sivanov-nuodb:
> As per our slack chat, one area that I think needs some documentation is on how backup and restore work in a multi-SG deployment, specifically around the admin validation for all SGs being selected, what is and isn't supported, co-ordinated backup etc. Perhaps a future PR can extend on docs/BackupAndRestore.md

Right. Updating the existing document will come in a separate PR.

PR #315, opened 2023-04-11
Expose process filter for database and restore charts

sivanov-nuodb:
**Changes**

- allow users to define the process filter to select the SMs part of a backup group
- allow users to supply process filter to select the archives to restore
- update the `nuorestore` -> `nuodocker request restore` CLI options plumbing to match the expected format. Label keys/values are passed as separate options. Process filter is passed as one option.

**Testing**

- change release names for some integration tests so that they don't contain upper case (newer Helm binaries don't allow them)
- add new integration tests
- add E2E tests for database restore with custom backup groups


stable/database/files/nuobackup
```diff
@@ -264,8 +271,15 @@ do
 done
 
 backup_type=$(echo $backup_type | tr '[:upper:]' '[:lower:]')
-# include all HCSMs if no labels are provided
-[ -z "$labels" ] && labels="role hotcopy"
+# determine the HCSMs selector based on the input
+if [ -n "$labels" ]; then
+   selector+=("--labels" "$labels")
+elif [ -n "$process_filter" ]; then
```
kontaras, line 277: Do you want to throw an error or print a warning if the user provides both a label and a filter?

stable/database/values.yaml
```diff
@@ -334,6 +334,7 @@ database:
       backupGroups: {}
       #  cluster0-0:
       #    labels: "pod-name sm-database-nuodb-cluster0-demo-hotcopy-0"
+      #    processFilter: "label(pod-name sm-database-nuodb-cluster0-demo-hotcopy-0)"
```
kontaras, line 337: How about providing a sample value that is not just the label filter, to give an example of when the user will want to use the option?

kontaras (approved):

stable/database/README.md
```diff
@@ -238,6 +238,7 @@ The following tables list the configurable parameters of the `database` chart an
 | `sm.hotCopy.backupGroupPrefix` | Prefix for the automatically generated backup groups | `{{ .Values.cloud.cluster.name }}` |
 | `sm.hotCopy.backupGroups` | Backup groups configuration. By default a backup group per HCSM is created automatically | `{}` |
 | `sm.hotCopy.backupGroups.<name>.labels` | Space separated process labels used to select the Storage Managers which are part of this backup group. _Any_ label key and value should match for the SM to be selected | `nil` |
+| `sm.hotCopy.backupGroups.<name>.processFilter` | LBQuery expression used to select the Storage Managers which are part of this backup group. For more information on LBQuery expression syntax, see <https://doc.nuodb.com/nuodb/latest/client-development/load-balancer-policies/#lbquery-expression-syntax>. | `nil` |
```
adriansuarez, line 241: Can you use a link with text so that it does not fill up the field? e.g.
```
For more information, see [LBQuery expression syntax](https://doc.nuodb.com/nuodb/latest/client-development/load-balancer-policies/#lbquery-expression-syntax).
```

stable/database/files/nuobackup
adriansuarez: `nuodocker backup database` enforces mutual exclusion, but this does suppress that check and only supply one of the two supplied arguments, which we don't want to do.

I would change this so that we plumb everything through to `nuodocker backup database` and let it decide what to do with the arguments. That could be done by replacing `elif` with `if` and explicitly checking the `selectors` array when deciding whether to inject the default.

adriansuarez (approved):
Looks good. Not sure if the test failures are related.

stable/database/README.md
sivanov-nuodb: Good point!

stable/database/values.yaml
sivanov-nuodb: I've intentionally added commented-out filter which matches the label selector to illustrate how the same use case is implemented either by `labels` or `processFilter`. Currently, we don't have a lot of distinct labels on SMs so that we can leverage fully the power of LBQuery syntax, however, there is a plan to "resurrect" the _placement_ labels like _zone_, _region_, _node_, etc.
I will take your comment into consideration when documenting backup groups in the user docs.

sivanov-nuodb:
> Looks good. Not sure if the test failures are related.

Not actually. I opened and merged other PRs to address them.

PR #317, opened 2023-04-21
Fix flaky tests

sivanov-nuodb:
Upgrading the Kubernetes version exposed some random test failures.

### DIFF
--- a/test/testlib/nuodb_database_utilities.go
+++ b/test/testlib/nuodb_database_utilities.go
@@ -419,26 +419,29 @@ func GetLatestBackup(t *testing.T, namespaceName string, podName string,
 	return backupset
 }
 
-func CheckArchives(t *testing.T, namespaceName string, adminPod string, dbName string, numExpected int, numExpectedRemoved int) (archives []NuoDBArchive, removedArchives []NuoDBArchive) {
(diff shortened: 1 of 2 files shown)


bkelly-ndb (approved):

test/testlib/nuodb_database_utilities.go
```diff
@@ -275,7 +275,13 @@ func UpgradeDatabase(t *testing.T, namespaceName string, helmChartReleaseName st
 	if upgradeOptions.TePodShouldGetRecreated {
 		AwaitPodObjectRecreated(t, namespaceName, tePod, 30*time.Second)
 	}
-	AwaitPodUp(t, namespaceName, tePodName, 180*time.Second)
+	// the TE ReplicaSet can be recreated with a different name so fetch the pod
+	// using a template continuously
+	Await(t, func() bool {
+		pod, err := findPod(t, namespaceName, tePodNameTemplate)
+		require.NoError(t, err, tePodNameTemplate)
```
adriansuarez, line 282: Is it correct to do an assertion in this function that is returning a bool indicating if the check was successful?

adriansuarez (approved):
Looks good.

test/testlib/nuodb_database_utilities.go
sivanov-nuodb: Good point! We don't need to fail fast in this case.

PR #319, opened 2023-05-30
Add liveness probe and improve readiness probe plumbing

adriansuarez:
- Add liveness probe that uses `nuocmd check server --check-converged` to check the health of the nuoadmin container. By default, this is enabled and causes the nuoadmin container to restart after 10 minutes of probe failures. This replaces the existing liveness probe, which was a TCP probe on port 8888 that probably did nothing.
- Expose parameters for existing readiness probe as Helm values and lower the default failure threshold. Also use a small timeout on --check-converged to guard against concurrent commits causing false positives where the commit index is temporarily stale.

### DIFF
--- a/stable/admin/README.md
+++ b/stable/admin/README.md
@@ -208,7 +208,17 @@ The following tables list the configurable parameters for the `admin` option of 
 | `serviceSuffix.balancer` | The suffix to use for the LoadBalancer service name | `balancer` |
 | `serviceSuffix.clusterip` | The suffix to use for the ClusterIP service name | `clusterip` |
 | `serviceSuffix.nodeport` | The suffix to use for the NodePort service name | `nodeport` |
-| `readinessTimeoutSeconds` | Admin readiness probe timeout, sometimes needs adjusting depending on environment and pod resources | `1` |
+| `livenessProbe.enabled` | Whether to enable the liveness probe for the nuoadmin container | `true` |
+| `livenessProbe.initialDelaySeconds` | The initial delay for the liveness probe | `60` |
+| `livenessProbe.periodSeconds` | The period at which the liveness probe is invoked | `60` |
+| `livenessProbe.failureThreshold` | The number of times that the liveness probe must fail consecutively before the nuoadmin container is restarted | `10` |
+| `livenessProbe.timeoutSeconds` | The timeout for liveness probe invocations | `20` |
+| `readinessProbe.initialDelaySeconds` | The initial delay for the readiness probe | `10` |
+| `readinessProbe.periodSeconds` | The period at which the readiness probe is invoked | `15` |
+| `readinessProbe.failureThreshold` | The number of times that the readiness probe must fail consecutively before the nuoadmin container is considered not ready | `4` |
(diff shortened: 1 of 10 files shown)


test/minikube/minikube_rolling_upgrade_test.go
```diff
@@ -67,28 +71,74 @@ func TestAdminReadinessProbe(t *testing.T) {
...
-		// make sure readinessprobe fails
-		output, err = k8s.RunKubectlAndGetOutputE(t, options, "exec", admin1, "-c", "admin", "--", "readinessprobe")
-		code, err := shell.GetExitCodeForRunCommandError(err)
-		require.NoError(t, err)
-		if code == 0 {
-			return false
+		// make sure readinessprobe and livenessprobe fail by invoking
+		// them directly
+		for _, probe := range []string{"readinessprobe", "livenessprobe"} {
+			output, err = k8s.RunKubectlAndGetOutputE(t, options, "exec", admin1, "-c", "admin", "--", probe)
```
sivanov-nuodb, line 104: Is there a timing issue here because the liveness probe might have restarted the container already due to `admin.livenessProbe.failureThreshold=1`?

stable/admin/README.md
```diff
@@ -208,7 +208,17 @@ The following tables list the configurable parameters for the `admin` option of
...
+| `livenessProbe.enabled` | Whether to enable the liveness probe for the nuoadmin container | `true` |
+| `livenessProbe.initialDelaySeconds` | The initial delay for the liveness probe | `60` |
+| `livenessProbe.periodSeconds` | The period at which the liveness probe is invoked | `60` |
+| `livenessProbe.failureThreshold` | The number of times that the liveness probe must fail consecutively before the nuoadmin container is restarted | `10` |
+| `livenessProbe.timeoutSeconds` | The timeout for liveness probe invocations | `20` |
+| `readinessProbe.initialDelaySeconds` | The initial delay for the readiness probe | `10` |
+| `readinessProbe.periodSeconds` | The period at which the readiness probe is invoked | `15` |
+| `readinessProbe.failureThreshold` | The number of times that the readiness probe must fail consecutively before the nuoadmin container is considered not ready | `4` |
+| `readinessProbe.successThreshold` | The number of times that the readiness probe must succeed consecutively before the nuoadmin container is considered ready | `1` |
```
sivanov-nuodb, line 219: It was `2` before. Are we changing this to `1` because of the internal timeout of 5sec?

sivanov-nuodb (approved):
Looks good.
I've just added a question in the test.

test/minikube/minikube_rolling_upgrade_test.go
adriansuarez: Good point. I will just remove these checks since they are redundant with the status checks that are performed below.

stable/admin/README.md
adriansuarez: I'm not sure why it was ever 2. The fact that we are now using `nuocmd check server ... --check-converged` makes the probe more resilient to spurious failures or successes.

kontaras (approved):

PR #320, opened 2023-06-01
Limit CronJob and StatefulSet name length

sivanov-nuodb:
**Issues**

Using long identifiers for cluster, domain, and database names can cause issues in creating Kubernetes primitives. StatefulSet names can't exceed 52 chars until https://github.com/kubernetes/kubernetes/issues/64023 is fixed. Truncating CronJob names to 52 charts may omit the backup group name and lead to duplicate resources.

**Changes**

- Limit the _database_ StatefulSet names to 52 chars.  If the name exceeds the limit, add a 7 chars SHA256 hash suffix to avoid collisions.
- Limit the backup CronJob names to 52 chars. If the name exceeds the limit, add a 7 chars SHA256 hash suffix to avoid collisions.
- Added `component` label to all workloads (Deployment, StatefulSet) and `role` label to SM StatefulSets. This will simplify filtering workloads per component type.

**Example**

```
$ helm install admin  stable/admin \
  -f values.yaml \
  --set admin.domain=superlongdomainname
$ helm install database stable/database \
    -f values.yaml \
    --set admin.domain=superlongdomainname \
    --set database.name=superlongdatabasename

$ kubectl get pods
NAME                                                              READY   STATUS             RESTARTS      AGE
admin-superlongdomainname-cluster0-0                              1/1     Running            0             2m11s
sm-database-superlongdomainname-cluster0-sup-088aace-0            1/1     Running            0             85s
te-database-superlongdomainname-cluster0-superlongdat-786fhz5rv   1/1     Running            0             85s

$ kubectl get cronjobs
NAME                                                   SCHEDULE      SUSPEND   ACTIVE   LAST SCHEDULE   AGE
full-hotcopy-superlongdomainname-superlongda-ff68eea   35 22 * * 6   False     0        <none>          101s
incremental-hotcopy-superlongdomainname-supe-66c725e   ?/2 * * * *   False     1        78s             101s
journal-hotcopy-superlongdomainname-superlon-1f8ea1e   ?/1 * * * *   False     2        18s             101s
```

**Testing**

- new integration tests
- regression testing

### DIFF
--- a/stable/database/templates/_helpers.tpl
+++ b/stable/database/templates/_helpers.tpl
@@ -26,6 +26,13 @@ and we have to allow for added suffixes including "-hotcopy" and "-NN" where NN 
 {{- end -}}
 
 {{/*
+The StatefulSet name is limited to 52 charts (https://github.com/kubernetes/kubernetes/issues/64023).
+*/}}
+{{- define "database.statefulset.name" -}}
+{{ template "truncWithHash" (list . 52) }}
+{{- end -}}
+
+{{/*
 Create chart name as used by the chart label.
 */}}
 {{- define "database.chart" -}}
--- a/stable/database/templates/_helpers.tpl
+++ b/stable/database/templates/_helpers.tpl
@@ -261,7 +261,8 @@ into account any schedule overrides configured per backup group.
 Renders the name of the HotCopy CronJob
 */}}
 {{- define "hotcopy.cronjob.name" -}}
-{{ printf "%s-hotcopy-%s-%s-%s" .hotCopyType .Values.admin.domain .Values.database.name .backupGroup | trunc 52 | trimSuffix "-" }}
+{{- $name := printf "%s-hotcopy-%s-%s-%s" .hotCopyType .Values.admin.domain .Values.database.name .backupGroup -}}
+{{ template "truncWithHash" (list $name 52) }}
 {{- end -}}
 
 {{/*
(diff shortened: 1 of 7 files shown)


stable/database/templates/_helpers.tpl
```diff
@@ -261,7 +268,8 @@ into account any schedule overrides configured per backup group.
 Renders the name of the HotCopy CronJob
 */}}
 {{- define "hotcopy.cronjob.name" -}}
-{{ printf "%s-hotcopy-%s-%s-%s" .hotCopyType .Values.admin.domain .Values.database.name .backupGroup | trunc 52 | trimSuffix "-" }}
+{{- $name := printf "%s-hotcopy-%s-%s-%s" .hotCopyType .Values.admin.domain .Values.database.name .backupGroup -}}
```
adriansuarez, line 271: I'm always confused about how these values get evaluated. How can we evaluate `.hotCopyType` and also have access to the root context to be able to get `.Values.admin.domain` and `.Values.database.name`?

stable/database/templates/_helpers.tpl
```diff
@@ -616,3 +624,23 @@ rendered, false otherwise.
...
+length exceeds the supplied max length, a 7 characters SHA256 hash is
+calculated and appended to the truncated string to make it unique.
+*/}}
+{{- define "truncWithHash" -}}
+{{- $s := index . 0 -}}
+{{- $max := index . 1 -}}
+{{- if not (typeIs "string" $s) -}}
+{{- fail (printf "truncWithHash template failed: not supported type '%s'" (typeOf $s)) -}}
+{{- end -}}
+{{- if gt (len $s) $max -}}
```
adriansuarez, line 639: Is there a minimum max value we should enforce? Obviously, you can't pass in max less than or equal to 8 since we reserve space for the hash suffix.

adriansuarez (approved):
Looks good.

stable/database/templates/_helpers.tpl
sivanov-nuodb: Right. Lacking the ability to define custom functions (or at lest I don't know how to do it) forces us to declare named templates and use them as they were functions. The problem is that there is no way to explicitly define the input variables. Another downside is that you can't use these templates with the Helm pipeline syntax.
I found an interesting [blog post](https://itnext.io/use-named-templates-like-functions-in-helm-charts-641fbcec38da) with more tricks.

stable/database/templates/_helpers.tpl
sivanov-nuodb: Good point! I will add validation for that.

bkelly-ndb (approved):
Nice change!

kontaras (approved):

PR #323, opened 2023-08-28
Use release-tools submodule containing relman.py

adriansuarez:
Delete release/ directory and replace it with a submodule pointing to the github.com/nuodb/release-tools repository. This enables usage of the tool in different repositories.


sivanov-nuodb (approved):
Looks good.

PR #324, opened 2023-09-06
Update sample archive protocol version to 5.0

sivanov-nuodb:
Update the protocol version of the sample archive used in the restore tests to 5.0.


bkelly-ndb (approved):

adriansuarez (approved):
Thanks for updating it.

PR #326, opened 2023-09-20
Install restore ConfigMap for secondary releases

sivanov-nuodb:
**Changes**

```
Previously it wasn't possible to define different autoImport sources for
secondary Helm releases as the `restore` ConfigMap was installed only by the
primary Helm release. This prevented automatic database import from working
with storage groups.

Install restore ConfigMap, which has the autoImport and autoRestore
configuration, by secondary Helm release.  This allows auto-importing database
backups/archives when TPSG is enabled. The backup URL from the database archive
servicing a storage group must be configured as `database.autoImport.source`
when installing the corresponding storage group Helm release. When performing
database auto-import, the `database.te.replicas=0` must be set which won't
start Transaction engines (TEs) initially concurrently with adding storage
groups to the database. Once all storage groups transition to an Available
state, the TEs can be scaled up.

The `database.autoImport.credentials` are set only once as they are stored in
the database Secret installed by the primary release. This change doesn't
introduce another Secret resource specific to restore as it is unlikely for
backups from different storage groups to be kept in different backup
repositories.
```


bkelly-ndb (approved):
Looks good. Tested with both autoimport and autorestore and both worked fine. For autoimport on installation TE replicas had to be set to 0. We can handle that in the other review in the troubleshooting section.

PR #329, opened 2023-10-04
License related test change for NuoDB 6

NexFlare:
Changes 

1. VerifyLincenseIsCommunity deprecated and VerifyLicense function added. This function needs an additional parameter `expected` which is of type LicenseType.
2. ApplyNuoDBLicense deprecated and ApplyLicense function added. This function needs an additional paramter `licenseType` which is of type LicenseType.
3. Added a new env variable NUODB_LIMITED_LICENSE_CONTENT for limited license.
4. Added a new function `RunOnNuoDBVersionFromOptions` which also checks the version provided in `nuodb.image.tag`.


test/testlib/minikube_utilities.go
```diff
@@ -835,7 +836,28 @@ func VerifyLicenseIsCommunity(t *testing.T, namespace string, podName string) {
...
 	require.NoError(t, err)
-	require.True(t, strings.Contains(output, "server license: Community"), output)
+
+	RunOnNuoDBVersionCondition(t, ">=6.0.0", func(version *semver.Version) {
+		require.True(t, strings.Contains(output, "server license: Limited"), output)
+	})
+
+	RunOnNuoDBVersionCondition(t, "<6.0.0", func(version *semver.Version) {
+		require.True(t, strings.Contains(output, "server license: Community"), output)
+	})
```
sivanov-nuodb, line 846: There is a behavior change that might break the test calling this function. E.g. before 6.0.0, the test will succeed without installing a license but now a license needs to be installed to get "server license: Limited".
I would suggest deprecating/deleting the `VerifyLicenseIsCommunity` function and replacing it with `VerifyLicense` which accepts the license type. For example:

```go
const (
	UNLICENSED LicenseType = "Unlicensed"
	LIMITED    LicenseType = "Limited"
	ENTERPRISE LicenseType = "Enterprise"
)

func VerifyLicense(t *testing.T, namespace string, podName string, expected LicenseType) {
	options := k8s.NewKubectlOptions("", "", namespace)

	output, err := k8s.RunKubectlAndGetOutputE(t, options, "exec", "-c", "admin", podName, "--", "nuocmd", "show", "domain")
	require.NoError(t, err)

	expectedString := fmt.Sprintf("server license: %s", expected)
	if expected != ENTERPRISE {
		RunOnNuoDBVersionCondition(t, "<6.0.0", func(version *semver.Version) {
			// Prior to NuoDB v6.0.0, Community and Enterprise license types are
			// supported
			expectedString = "server license: Community"
		})
	}
	require.True(t, strings.Contains(output, expectedString), output)
}
```

sivanov-nuodb:
@NexFlare I think that the tests will still fail with NuoDB 6.0.0 as no license is installed by default which will prevent processes from starting. You would need to install a limited license (either via `valuesInject.yaml` or programmatically) for all tests.

NexFlare:
> @NexFlare I think that the tests will still fail with NuoDB 6.0.0 as no license is installed by default which will prevent processes from starting. You would need to install a limited license (either via `valuesInject.yaml` or programmatically) for all tests.

The limited or a community license is provided in valuesInject.yaml file from NuoDB repo now when test are invoked in Bamboo.

sivanov-nuodb:
> > @NexFlare I think that the tests will still fail with NuoDB 6.0.0 as no license is installed by default which will prevent processes from starting. You would need to install a limited license (either via `valuesInject.yaml` or programmatically) for all tests.
> 
> The limited or a community license is provided in valuesInject.yaml file from NuoDB repo now when test are invoked in Bamboo. Would it be better to follow the concept similar to having license as env variable an applying it to test that require them.

I would suggest that the `testlib.ApplyNuoDBLicense` is called in `testlib.StartAdminTemplate` unconditionally as since NuoDB 6.0.0 no engines could be started otherwise.

test/testlib/minikube_utilities.go
```diff
@@ -830,11 +831,29 @@ func VerifyLicenseFile(t *testing.T, namespace string, podName string, expectedL
...
+		RunOnNuoDBVersionCondition(t, "<6.0.0", func(version *semver.Version) {
+			// Prior to NuoDB v6.0.0, Community and Enterprise license types are
+			// supported
+			expectedString = "server license: Community"
+		})
+	}
+	require.True(t, strings.Contains(output, expectedString), output)
+}
+
 func VerifyLicenseIsCommunity(t *testing.T, namespace string, podName string) {
```
sivanov-nuodb, line 851: Please include a note in the method doc string that it is deprecated. For example:

```go
// Deprecated: The VerifyLicenseIsCommunity function has been deprecated. When testing with NuoDB v6.0.0, use the VerifyLicense function instead.
```

sivanov-nuodb (approved):
Looks good.
I left a minor comment about leaving a deprecated note for `VerifyLicenseIsCommunity` function.

test/testlib/nuodb_admin_utilities.go
```diff
@@ -142,6 +143,18 @@ func StartAdminTemplate(t *testing.T, options *helm.Options, replicaCount int, n
 		}
 	}
 
+	admin0 := fmt.Sprintf("%s-nuodb-cluster0-0", helmChartReleaseName)
+
+	RunOnNuoDBVersionCondition(t, ">=6.0.0", func(version *semver.Version) {
```
sivanov-nuodb, line 148: Nit: The `RunOnNuoDBVersionCondition` will be called for every test now.  It creates a pod to check the NuoDB version which takes some time. You could cache the fetched version in a global variable to minimize the performance impact.

test/testlib/nuodb_admin_utilities.go
```diff
@@ -142,6 +143,18 @@ func StartAdminTemplate(t *testing.T, options *helm.Options, replicaCount int, n
...
+		//With NuoDB v6.0.0 a license is mandatory to add
+		if os.Getenv("NUODB_LICENSE_CONTENT") == "" {
+			t.Error("License required to start NuoDB tests.")
+		}
+		ApplyLicense(t, namespaceName, admin0, ENTERPRISE)
```
sivanov-nuodb, line 153: You have added another env variable that has the limited license `NUODB_LIMITED_LICENSE_CONTENT`. I think it makes more sense to install it for all the tests now. E.g.

```go
if os.Getenv("NUODB_LIMITED_LICENSE_CONTENT") == "" {
  t.Error("Limited license required to start NuoDB tests.")
}
ApplyLicense(t, namespaceName, adminNames[0], LIMITED)
```

test/testlib/nuodb_admin_utilities.go
```diff
@@ -193,10 +206,23 @@ func AwaitNrLoadBalancerPolicies(t *testing.T, namespace string, podName string,
 		return err == nil && len(policies) == expectedNumber
 	}, 30*time.Second)
 }
-
+/*
+	Deprecated
```
sivanov-nuodb, line 210: Nit: [Godoc](https://go.dev/blog/godoc) is looking for string literal `Deprecated:` on a new paragraph.

> Sometimes a struct field, function, type, or even a whole package becomes redundant or unnecessary, but must be kept for compatibility with existing programs. To signal that an identifier should not be used, add a paragraph to its doc comment that begins with “Deprecated:” followed by some information about the deprecation.

test/testlib/nuodb_database_utilities.go
```diff
@@ -636,6 +642,22 @@ func RunOnNuoDBVersion(t *testing.T, versionCheckFunc func(*semver.Version) bool
 	}
 }
 
+func RunOnNuoDBVersionFromOptions(t *testing.T, options *helm.Options, versionCheckFunc func(*semver.Version) bool, actionFunc func(*semver.Version)) {
```
sivanov-nuodb, line 645: Why not reuse the `RunOnNuoDBVersionFromOptions` in `RunOnNuoDBVersion`?

test/testlib/nuodb_database_utilities.go
```diff
@@ -593,6 +595,9 @@ func ServeFileViaHTTP(t *testing.T, namespaceName string, localFilePath string)
 }
 
 func GetNuoDBVersion(t *testing.T, namespaceName string, options *helm.Options) string {
+	if nuoDBVersion != "" {
+		return nuoDBVersion
```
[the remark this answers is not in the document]
sivanov-nuodb: What I'm saying here is that if someone calls this function with two different tags supplied in options, the result will be the same and it should be.

test/testlib/minikube_utilities.go
```diff
@@ -830,11 +831,35 @@ func VerifyLicenseFile(t *testing.T, namespace string, podName string, expectedL
...
+	}
+	output = strings.ToLower(output)
+	expectedString = strings.ToLower(expectedString)
+	require.True(t, strings.Contains(output, expectedString), output)
+}
+
+
+// Deprecated: The VerifyLicenseIsCommunity function has been deprecated. When testing with NuoDB v6.0.0, use the VerifyLicense function instead.
+
+
```
sivanov-nuodb, line 856: Nit: Please remove the extra empty lines.

test/testlib/nuodb_admin_utilities.go
```diff
@@ -142,6 +143,17 @@ func StartAdminTemplate(t *testing.T, options *helm.Options, replicaCount int, n
 		}
 	}
 
+	// License is mandatory for running test with NuoDB 6.0
+	RunOnNuoDBVersionFromOptionCondition(t, options, ">=6.0.0", func(version *semver.Version) {
+		if os.Getenv("NUODB_LIMITED_LICENSE_CONTENT") != "" {
+            ApplyLicense(t, namespaceName, adminNames[0], LIMITED)
+		} else if os.Getenv("NUODB_LICENSE_CONTENT") != "" {
+            ApplyLicense(t, namespaceName, adminNames[0], ENTERPRISE)
+        } else {
```
sivanov-nuodb, line 152: Nit: Could you please fix the code formatting?

test/testlib/nuodb_database_utilities.go
sivanov-nuodb: To be more precise you should have something like:

```go
func RunOnNuoDBVersion(t *testing.T, versionCheckFunc func(*semver.Version) bool, actionFunc func(*semver.Version)) {
	RunOnNuoDBVersionFromOptions(t, &helm.Options{}, versionCheckFunc, actionFunc)
}

func RunOnNuoDBVersionFromOptions(t *testing.T, options *helm.Options, versionCheckFunc func(*semver.Version) bool, actionFunc func(*semver.Version)) {
...
}
```

test/testlib/nuodb_admin_utilities.go
NexFlare: sure

test/testlib/nuodb_database_utilities.go
```diff
@@ -593,6 +597,10 @@ func ServeFileViaHTTP(t *testing.T, namespaceName string, localFilePath string)
 }
 
 func GetNuoDBVersion(t *testing.T, namespaceName string, options *helm.Options) string {
+	_, ok := nuoDBVersionMap[options.SetValues["nuodb.image.tag"]]
+	if(ok) {
+		return nuoDBVersionMap[options.SetValues["nuodb.image.tag"]]
+	}
```
sivanov-nuodb, line 603: Nit: You can avoid getting the value from the map twice. It is often written in Go like this:

```go
        if v, ok := nuoDBVersionMap[options.SetValues["nuodb.image.tag"]]; ok {
                 return v
        }
```

sivanov-nuodb (approved):
Looks good. 
Please check my comment about reusing the `RunOnNuoDBVersionFromOptions` in `RunOnNuoDBVersion`.

test/testlib/nuodb_database_utilities.go
NexFlare: If we are actually just calling RunOnNuoDBVersionFromOptions from RunOnNuoDBVersion, should I just not deprecate the RunOnNuoDBVersion ?

NexFlare: I have done something like this and extracted the common in both functions.
```
func RunOnNuoDBVersion(t *testing.T, versionCheckFunc func(*semver.Version) bool, actionFunc func(*semver.Version)) {
	versionString := GetNuoDBVersion(t, "default", &helm.Options{})
	// Select only the main NuoDB version (i.e 4.2.1) from the full version string
	checkVersion(t, versionString, versionCheckFunc, actionFunc)
}

func RunOnNuoDBVersionFromOptions(t *testing.T, options *helm.Options, versionCheckFunc func(*semver.Version) bool, actionFunc func(*semver.Version)) {
	versionString, ok := options.SetValues["nuodb.image.tag"]
	if !ok || !NuoDBVersionRegex.MatchString(versionString) {
		versionString = GetNuoDBVersion(t, "default", options)
	}
	checkVersion(t, versionString, versionCheckFunc, actionFunc)
}

func checkVersion(t *testing.T, versionString string, versionCheckFunc func(*semver.Version) bool, actionFunc func(*semver.Version)) {
	version, err := semver.NewVersion(NuoDBVersionRegex.ReplaceAllString(versionString, "${1}"))
	require.NoError(t, err, "Unable to parse NuoDB version", versionString)

	if versionCheckFunc(version) {
		actionFunc(version)
	} else {
		t.Logf("Conditional logic not executed for NuoDB version %s", versionString)
	}
}
```

test/testlib/nuodb_database_utilities.go
NexFlare: Thanks. I will make the necessary changes.

test/testlib/nuodb_database_utilities.go
sivanov-nuodb: > If we are actually just calling RunOnNuoDBVersionFromOptions from RunOnNuoDBVersion, should I just not deprecate the RunOnNuoDBVersion ?

There is no reason to deprecate `RunOnNuoDBVersion` function as it is just a wrapper of `RunOnNuoDBVersionFromOptions`. 

> I have done something like this and extracted the common in both functions.

You can replace the entire body of the existing `RunOnNuoDBVersion` by calling `RunOnNuoDBVersionFromOptions` with empty options as suggested above.

test/testlib/nuodb_database_utilities.go
```diff
@@ -39,9 +39,13 @@ type ExtractedOptions struct {
 	ClusterName           string
 	EntrypointClusterName string
 	DbPrimaryRelease      bool
-	DomainName            string
+	DomainName            string
```
adriansuarez, line 42: Remove the trailing space/tab.

test/minikube/minikube_rolling_upgrade_test.go
```diff
@@ -234,7 +234,7 @@ func TestKubernetesUpgradeFullDatabase(t *testing.T) {
 	expectedNewVersion := testlib.GetUpgradedReleaseVersion(t, &options)
 
 	// enable debug logging for KAA after the update
-	options.ValuesFiles = []string{"../files/admin-kaa-debug.yaml"}
+	// options.ValuesFiles = []string{"../files/admin-kaa-debug.yaml"}
```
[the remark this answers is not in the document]
NexFlare: Sure. I will make the necessary changes

adriansuarez:
@NexFlare, did you verify that the Bamboo tests pass without the change to remove the logging override?

Also, don't forget to remove this from your commit message:
> 3. Commented code from ../files/admin-kaa-debug.yaml, since it was overriding the license supplied from the nuodb repo.

adriansuarez (approved):
Looks good.

test/testlib/nuodb_admin_utilities.go
```diff
@@ -142,6 +143,24 @@ func StartAdminTemplate(t *testing.T, options *helm.Options, replicaCount int, n
 		}
 	}
 
+	
+	// LIMITED license takes priority over Enterprise license because
+	// we don't want to change the test flow for NuoDB version 6.0.
+	// Any test which requires Enterprise license would install it using
+	// ApplyLicense function
+	if os.Getenv("NUODB_LIMITED_LICENSE_CONTENT") != "" {
+		ApplyLicense(t, namespaceName, adminNames[0], LIMITED)
```
sivanov-nuodb, line 152: This is a noop for NuoDB prior to 6.0.0, right?

sivanov-nuodb (approved):
Looks good.

adriansuarez (approved):
Looks good.

PR #330, opened 2023-10-12
TLS rotation enhancements

sivanov-nuodb:
**Changes**

```
- Added support for supplying NuoAdmin keystore and truststore passwords via
  Secret. If both `password` and `passwordKey` values are defined, the first
  one takes precedence.
- Now NuoAdmin domain doesn't need to be restarted on keystore change. The
  volume mounts will automatically receive updates for the volume sources
  defined in the projected volume which simplifies the TLS keys rotation.
- Added support for TLS keys rotation via `helm upgrade`. Previously the
  database processes had to be manually restarted with `kubectl rollout restart`
  to receive new TLS certificates. Now the restart is performed
  automatically allowing the user to interact only with the Helm client to
  perform the TLS keys rotation.
```

**Testing**

- Added new integration tests (previously there weren't any for TLS)
- Modified the `TestKubernetesTLSRotation` E2E test to use `helm upgrade` strategy for TLS keys rotation


stable/admin/README.md
```diff
@@ -200,9 +200,11 @@ The following tables list the configurable parameters for the `admin` option of
 | `tlsKeyStore.secret` | TLS keystore secret name | `nil` |
 | `tlsKeyStore.key` | TLS keystore secret key | `nil` |
 | `tlsKeyStore.password` | TLS keystore secret password | `nil` |
+| `tlsKeyStore.passwordKey` | TLS keystore secret password key. One of `tlsKeyStore.password` or `tlsKeyStore.passwordKey` must be used | `nil` |
```
adriansuarez, line 203: This is misleading, because 0 can be used. I would change the wording to "At most one..."

adriansuarez: Why don't we just set a default of `password` for the password key and only use it if TLS is enabled and `tlsKeyStore.password` is `nil`?

adriansuarez: So it looks like actually the default value of `tlsKeyStore` is `nil`, and any sub-keys within it are what indicate that TLS is enabled. I guess we can inject the `password` in the place where the environment variable is defined.

stable/database/templates/_helpers.tpl
```diff
@@ -649,4 +649,92 @@ calculated and appended to the truncated string to make it unique.
...
+*/}}
+{{- define "database.tlsConfigAnnotations" -}}
+{{- if .Values.admin.tlsKeyStore }}
+  {{- $password := include "database.keystorePassword" (list . .Values.admin.tlsKeyStore) -}}
+  {{- if $password }}
+checksum/tls-passwords: {{ sha256sum $password }}
+  {{- end }}
+  {{- $keystore := include "database.keystore" (list . .Values.admin.tlsKeyStore) -}}
+  {{- if $keystore }}
+checksum/tls-keystore: {{ sha256sum $keystore }}
```
adriansuarez, line 737: The user still has to take manual action for the new keystore and truststore to be loaded. Before it was by running `kubectl rollout statefulset/deployment ...`, and now it is by running `helm upgrade ...` just to cause the checksum to be recalculated from the secret resources.

test/testlib/tls.go
```diff
@@ -167,6 +192,32 @@ func RotateTLSCertificates(t *testing.T, options *helm.Options, namespaceName st
...
+			if actual != adminPem {
+				notUpdated = append(notUpdated, admin)
+			}
+		}
+		if len(notUpdated) > 0 {
+			t.Logf("Server certificates not updated for APs %s", notUpdated)
+			return false
+		}
+		return true
+	}, timeout)
```
adriansuarez, line 218: So ConfigMaps/Secrets do get reloaded? I have not been able to find documentation that confirms this and was always under the impression that this was not the case. Is this a newer feature?

stable/admin/README.md
sivanov-nuodb: If TLS is disabled, then `admin.tlsKeyStore`, `admin.tlsTrustStore`, `admin.tlsClientPEM`, and `admin.tlsCACert` are `nil`.  This change adds another optional property `passwordKey` to `admin.tlsKeyStore` and `admin.tlsTrustStore` Helm variables so that the user can supply passwords via Secret. Are you proposing to set a default value for this new property (i.e. `password`) in [here](https://github.com/nuodb/nuodb-helm-charts/blob/sivanov/HELM-389/stable/admin/templates/statefulset.yaml#L124-L132)?

test/testlib/tls.go
sivanov-nuodb: This is done by _kubelet_ and I don't think is a new feature. The secrets mounted with `subPath` doesn't receive updates though. This was how TLS secrets were mounted before which was unfortunate.
For more information on the Secret updates, see [here](https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-files-from-a-pod).

stable/admin/README.md
adriansuarez: Yes, basically replacing [line 131](https://github.com/nuodb/nuodb-helm-charts/blob/sivanov/HELM-389/stable/admin/templates/statefulset.yaml#L131) with
```
              key: {{ default "password" .Values.admin.tlsKeyStore.passwordKey }}
```

But you would also have to propagate that to the places where the actual password is obtained like the `admin.keystorePassword`. But if that's too much of a pain, we can skip it.

adriansuarez (approved):
Looks good.

test/testlib/tls.go
adriansuarez: Thanks for the pointer. It's good that ConfigMaps/Secrets mounted as files are reloaded. I thought we were going to have to rely on K8s API access to fetch the data.

bkelly-ndb (approved):
The change looks good, should the user guide for rotation be updated with this: https://github.com/nuodb/nuodb-helm-charts/blob/master/docs/HowToCertificatesRotation.md

Any impact/additions that could go into: https://github.com/nuodb/nuodb-helm-charts/blob/master/docs/HowtoTLS.md

sivanov-nuodb:
> The change looks good, should the user guide for rotation be updated with this: https://github.com/nuodb/nuodb-helm-charts/blob/master/docs/HowToCertificatesRotation.md
> 
> Any impact/additions that could go into: https://github.com/nuodb/nuodb-helm-charts/blob/master/docs/HowtoTLS.md

So that you know, the changes here are optimizations only and are backward compatible with what has been documented.
The TLS rotation guide changes the TLS secret names and doesn't benefit from this optimization. I would live any possible updates to the guides once the new version is released.

adriansuarez:
@sivanov-nuodb Latest changes look good.

acabrele (approved):
Changes look good to me.  Great to have this added, thanks Stanimir!

PR #332, opened 2023-10-13
Add storage group resource label

sivanov-nuodb:
**Changes**

```
Added `storage-group` label to relevant resources if TPSG is enabled for the
Helm release. Environments that use TPSG should be redeployed because of the
additional label added to the PVC section of the StatefulSet. This should be
fine as TPSG is an experimental feature in NuoDB Helm Charts v3.6.0.
```

### DIFF
--- a/stable/database/templates/_helpers.tpl
+++ b/stable/database/templates/_helpers.tpl
@@ -513,10 +513,30 @@ release: {{ .Release.Name | quote }}
 {{- end -}}
 
 {{/*
+Renders the storage group labels. IMPORTANT: Adding new entries must be done
+with caution as these labels are rendered in the PVC spec which is immutable.
+The storage group name can not be reconfigured by definition.
+*/}}
+{{- define "database.storageGroupLabels" -}}
+{{- if .Values.database.sm.storageGroup -}}
+{{- if .Values.database.sm.storageGroup.enabled -}}
+storage-group: {{ include "database.storageGroup.name" . | quote }}
+{{- end -}}
+{{- end -}}
+{{- end -}}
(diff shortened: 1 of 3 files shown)


agoelzer (approved):

kontaras (approved):

stable/database/templates/_helpers.tpl
```diff
@@ -512,13 +512,33 @@ release: {{ .Release.Name | quote }}
...
 
+{{/*
+Renders the storage group labels. IMPORTANT: Adding new entries must be done
+with caution as these labels are rendered in the PVC spec which is immutable.
+The storage group name can not be reconfigured by definition.
+*/}}
+{{- define "database.storageGroupLabels" -}}
+{{- if .Values.database.sm.storageGroup -}}
+{{- if .Values.database.sm.storageGroup.enabled -}}
+storage-group: {{ include "database.storageGroup.name" . | quote }}
```
adriansuarez, line 523: We follow the convention of prefixing the label and annotation keys with `cp.nuodb.com/` in the Control Plane. Should we adopt something similar with the Helm charts?

adriansuarez (approved):
Looks good.

bkelly-ndb (approved):

stable/database/templates/_helpers.tpl
sivanov-nuodb: Actually, "nuodb.com/" prefix is used only for KAA annotations. All other labels/annotations rendered by NuoDB Helm charts don't follow this concept. I would leave it like this for now.

PR #333, opened 2023-10-26
Delete Jobs when suspending CronJobs

sivanov-nuodb:
Previously suspending CronJobs didn't correctly delete jobs that may have been already scheduled.


kontaras (approved):

adriansuarez (approved):
Looks good.

PR #334, opened 2023-10-26
Failure to record metadata for backups

sivanov-nuodb:
**Issue**

Once the backup index reaches `NUODB_MAX_BACKUP_HISTORY`, it will be reset to
0. This will cause problems with recording future backup metadata history as
the values in the KV store for these keys are expected to be empty.

**Changes**

- Set the expected value for the next backup by first fetching the existing
  value. This solves the issue at hand while preserving the guard for
  concurrent modifications of the backup metadata at that index.
- Make sure that backup indexes always start with `0`. Previously the initial
  value was `1` but after reaching `NUODB_MAX_BACKUP_HISTORY`, the index was
reset to `0`. Changed the `report-backups` command to account for that.
- Render `database.env` Helm values in CronJob allowing users to pass opaque
  environment variables.

**Testing**

- Fixed initial backup indexes in restore tests that explicitly assert them
- Added integration and E2E tests

### DIFF
--- a/test/minikube/minikube_base_restore_test.go
+++ b/test/minikube/minikube_base_restore_test.go
@@ -24,13 +24,16 @@ import (
 
 func verifyBackup(t *testing.T, namespaceName string, podName string, databaseName string, options *helm.Options) {
 	// verify that the backup has been documented by the Admin layer
-	backupset, err := k8s.RunKubectlAndGetOutputE(t, options.KubectlOptions,
+	output, err := k8s.RunKubectlAndGetOutputE(t, options.KubectlOptions,
 		"exec", podName, "--",
 		"nuodocker", "get", "current-backup", "--db-name", databaseName,
 	)
+	require.NoError(t, err, "Error running 'nuodocker get current-backup': %s", output)
+	require.True(t, output != "")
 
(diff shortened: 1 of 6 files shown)


kontaras (approved):

test/minikube/minikube_base_restore_test.go
```diff
@@ -24,13 +24,16 @@ import (
 
 func verifyBackup(t *testing.T, namespaceName string, podName string, databaseName string, options *helm.Options) {
 	// verify that the backup has been documented by the Admin layer
-	backupset, err := k8s.RunKubectlAndGetOutputE(t, options.KubectlOptions,
+	output, err := k8s.RunKubectlAndGetOutputE(t, options.KubectlOptions,
 		"exec", podName, "--",
 		"nuodocker", "get", "current-backup", "--db-name", databaseName,
 	)
+	require.NoError(t, err, "Error running 'nuodocker get current-backup': %s", output)
+	require.True(t, output != "")
```
adriansuarez, line 32: Better to use `require.NotEqual(t, "", output)` so that the output is printed.

adriansuarez (approved):
Looks good.

PR #337, opened 2023-11-20
Start a database from a volume and archive snapshot

kontaras:
To enable restore and clone use cases, we need a way to start a database from a volume snapshot.

An archive snapshot is specified as a [TypedObjectReference](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#typedobjectreference-v1-core) and is passed as the helm value `database.persistence.dataSourceRef`. If the snapshot is of an archive without a journal, the journal snapshot should be passed in the helm values `database.sm.hotCopy.journalPath.persistence.dataSourceRef` and `database.sm.noHotCopy.journalPath.persistence.dataSourceRef`.

If the name of the domain or database has changed, there is code to relocate the file and register the moved archive with the admin service.

Unfortunately, this logic cannot differentiate between a snapshot recovered without changing the database and domain names and an SM that was restarted. In order to trigger the registration logic in this case, when taking the snapshot, there needs to be a pre-snapshot script that will create a `backup.json` file in the archive (and journal, if it is on a separate volume). That file should specify a single use backup id like `'{"id": "NONCE"}'`. That same backup id will need to be set as the helm value `database.autoImport.backup_id`. At restore time, the file will be deleted to prevent the archive from being registered twice.

**Test**:
I have added template tests for the new helm values as well as end to end tests for the restore logic. In order to support these tests the k8s/api version needs to be bumped, causing a cascade of other dependency updates. These dependencies require a newer version of GoLang (requiring an update of the CircleCI machine image). In addition, minikube needs to be reconfigured to enable support for CSI and volume snapshotting.

https://app.circleci.com/pipelines/github/nuodb/nuodb-helm-charts?branch=ktarashchanskiy%2FHELM-407
Branch renamed from https://app.circleci.com/pipelines/github/nuodb/nuodb-helm-charts?branch=kontaras%2Frestore-snapshot


stable/database/files/nuosm
```diff
@@ -387,6 +387,69 @@ function resolveLatestSource() {
...
+    if [ -z "$archives" ] || [ $(echo "$archives" | wc -l) != 1 ]; then 
+      die 1 "Did not find exactly 1 archive: ${archives}"
+    fi
+
+    local ARCHIVE_PATH=$( dirname $archives )
+    log "Found snapshot at ${ARCHIVE_PATH}"
+
+    if [ -n "$BACKUP_ID" ]; then
+      if [ ! -e ${ARCHIVE_PATH}/backup.json ] || \
+          [ $BACKUP_ID != $(sed -e 's|.*"id":\s*"\?\([^\s"]\+\)"\?[,}].*|\1|' "$ARCHIVE_PATH/backup.json") ]; then
```
adriansuarez, line 407: If the `backup.json` file has an extra space before the colon as below, then this will not find the backup ID.
```json
{
  "id" : "theId"
}
```

I think it defeats the point of using a JSON format if it is brittle like this. My preference is to just use a file with its entire contents consisting of the backup ID, and we can migrate to a JSON format whenever we `nuosm` as a `nuocmd` subcommand in Python.

stable/database/files/nuosm
```diff
@@ -387,6 +387,69 @@ function resolveLatestSource() {
...
+      if [ ! -d $EXPECTED_JOURNAL ]; then
+        die 1 "Did not find a journal snapshot at '$EXPECTED_JOURNAL'"
+      fi
+      
+      if [ -n "$BACKUP_ID" ]; then
+        if [ ! -e ${EXPECTED_JOURNAL}/backup.json ] || \
+            [ $BACKUP_ID != $(sed -e 's|.*"id":\s*"\?\([^\s"]\+\)"\?[,}].*|\1|' "$EXPECTED_JOURNAL/backup.json") ]; then
+          die 1 "Incorrect backup id in journal"
+        fi
+      fi
```
adriansuarez, line 425: Can you make this into its own function so that this regex doesn't appear twice?

stable/database/files/nuosm
```diff
@@ -387,6 +387,69 @@ function resolveLatestSource() {
   return 0
 }
 
+#=======================================
+# function - Looks for a previous archive and journal and imports it
+#
+function loadFromSnapshot() {
+  local RECREATE_ARCHIVES="false"
+
+  if [ -z "$(ls -A ${DB_DIR})" ]; then
```
adriansuarez, line 396: This is just checking that the archive directory is not empty?

I think we should the same thing we do in `nuocmd start sm`, which is look for `$DB_DIR/1.atm`, which contains the master catalog. See the Python snippet below from `nuodocker.py`:

```python
    def _has_archive_data(self, directory_path):
        return os.path.exists(os.path.join(directory_path, "1.atm"))
```

@madscientist, is it still safe with LSA to look for `1.atm`?

stable/database/files/nuosm
```diff
@@ -387,6 +387,69 @@ function resolveLatestSource() {
...
+      fi
+      
+      if [ -n "$BACKUP_ID" ]; then
+        if [ ! -e ${EXPECTED_JOURNAL}/backup.json ] || \
+            [ $BACKUP_ID != $(sed -e 's|.*"id":\s*"\?\([^\s"]\+\)"\?[,}].*|\1|' "$EXPECTED_JOURNAL/backup.json") ]; then
+          die 1 "Incorrect backup id in journal"
+        fi
+      fi
+
+      mv ${EXPECTED_JOURNAL}/* ${JOURNAL_DIR}
```
adriansuarez, line 427: This moves all of the files individually. Why not delete `$JOURNAL_DIR`, which we would expect to be empty or useless anyway, and then move the directory itself, which should be atomic.

stable/database/files/nuosm
```diff
@@ -387,6 +387,69 @@ function resolveLatestSource() {
...
+
+    fi
+
+    # Don't move the archive until after we have done error checking.
+    # Once we move the archive, we are going to be in the `archive already exists` branch
+    # when kubernetes retries/restarts the pod.
+    mv ${ARCHIVE_PATH}/* ${DB_DIR}
```
adriansuarez, line 434: Same comment as above. It is better to rename the directory rather than moving the entire contents of it.

stable/database/files/nuosm
```diff
@@ -387,6 +387,69 @@ function resolveLatestSource() {
...
+
+    local RECREATE_ARCHIVES="true"
+  else
+    log "archive already exists, not looking for snapshot archives"
+
+    if [ -e ${DB_DIR}/backup.json ] && [ -n "$BACKUP_ID" ] && \
+          [ $BACKUP_ID == $(sed -e 's|.*"id":\s*"\?\([^\s"]\+\)"\?[,}].*|\1|' "$DB_DIR/backup.json") ]; then
+      local RECREATE_ARCHIVES="true"
+    fi
```
adriansuarez, line 443: We check the `backup.json` in the journal directory above as a sanity check. Should we do the same here?

Also, we abort on mismatching backup ID above, but here we seem to disregard a mismatching backup ID.

stable/database/files/nuosm
```diff
@@ -415,6 +478,12 @@ if [ "$SEPARATE_JOURNAL" == "true" ]; then
   log "journal-dir=${JOURNAL_DIR}"
 fi
 
+if [ "$SNAPSHOT_RESTORED" == "true" ]; then
+  loadFromSnapshot
+else
+  log "No snapshot"
```
adriansuarez, line 484: Is this for diagnostics? I think we should log something if we are relocating the snapshot archive/journal and be silent otherwise.

adriansuarez:
@kontaras Can you add a commented out example to `stable/database/values.yaml` and descriptions of the new Helm values to `stable/database/README.md`?

stable/database/templates/statefulset.yaml
```diff
@@ -173,6 +173,8 @@ spec:
         - { name: OVERWRITE_COPIES,    value: "{{ .Values.database.sm.logPersistence.overwriteBackoff.copies | default "3" }}" }
         - { name: OVERWRITE_WINDOW,    value: "{{ .Values.database.sm.logPersistence.overwriteBackoff.windowMinutes | default "120" }}" }
         - { name: SEPARATE_JOURNAL,    value: "{{- include "defaultfalse" .Values.database.sm.noHotCopy.journalPath.enabled}}" }
+        - { name: SNAPSHOT_RESTORED,   value: "{{ not (empty .Values.database.persistence.dataSourceRef) }}" }
+        - { name: BACKUP_ID,           value: "{{ .Values.database.autoImport.backup_id }}" }
```
adriansuarez, line 177: Why is it `backup_id` instead of `backupId`, which is more consistent with our current naming conventions?

stable/database/templates/statefulset.yaml
```diff
@@ -686,6 +698,10 @@ spec:
         matchLabels:
           database: {{ .Values.database.name }}
 {{- end }}
+      {{- if .Values.database.persistence.dataSourceRef }}
```
adriansuarez, line 701: I think we should conditionalize on the name being non-empty and have an entire `dataSourceRef` section in our `values.yaml` with the following default values:

```yaml
      dataSourceRef:
        name: ""
        kind: VolumeSnapshot
        apiGroup: snapshot.storage.k8s.io
```

This way someone only has to specify `database.persistence.dataSourceRef.name` to some value to refer to VolumeSnapshot in the same namespace.

test/minikube/minikube_base_restore_test.go
```diff
@@ -436,3 +435,91 @@ func TestKubernetesAutoRestore(t *testing.T) {
...
+			"database.sm.hotCopy.journalPath.enabled":                            "true",
+		},
+	})
+
+	output, err = k8s.RunKubectlAndGetOutputE(t, kubectlOptions, "exec", admin0, "-c", "admin", "-c", "admin", "--", "bash", "-c",
+		fmt.Sprintf("echo \"SELECT id FROM testtbl;\" | nuosql %s --user dba --password secret", restoredDb))
+
+	require.NoError(t, err, output)
+
+	require.True(t, strings.Contains(output, "123"))
```
adriansuarez, line 524: I'd like to see this test write more data to the database and then verify that it is present after a restart.
1. Write more data.
2. Delete the SM pod.
3. Wait for the database to become running again.
4. Check that the data written in 1 is present.

This verifies that we can actually diverge from the snapshot/dataSource and also exercises the "normal restart" code paths in `nuosm`.

test/minikube/minikube_long_restore_test.go
```diff
@@ -891,3 +891,220 @@ func TestKubernetesRestoreDatabaseSeparateJournal(t *testing.T) {
...
+		fmt.Sprintf("echo \"CREATE TABLE testtbl (id INT); INSERT INTO testtbl (id) values (123);\" | nuosql %s --user dba --password secret", sourceDb))
+
+	require.NoError(t, err, output)
+
+	defer testlib.Teardown(testlib.TEARDOWN_SNAPSHOT)
+
+	smPod := fmt.Sprintf("sm-%s-nuodb-cluster0-%s-hotcopy-0", sourceDatabaseChartName, sourceDb)
+	achiveVolumeName := "archive-volume-" + smPod
+
+	noJournalNoBidSnapshotName := "noj-nobid-snapshot"
```
adriansuarez, line 1003: Can you add comments in here to say what each of these snapshots is testing?

test/minikube/minikube_long_restore_test.go
```diff
@@ -891,3 +891,220 @@ func TestKubernetesRestoreDatabaseSeparateJournal(t *testing.T) {
...
+	output = testDb("noj-misbid", noJournalNoBidSnapshotName, "", "123abc", false)
+	require.Contains(t, output, "Incorrect backup id in archive")
+
+	output = testDb("noj-badbid", noJournalBadBidSnapshotName, "", "123abc", false)
+	require.Contains(t, output, "Incorrect backup id in archive")
+
+	output = testDb("noj-2arch", noJournalExtraArchiveSnapshotName, "", "123abc", false)
+	require.Contains(t, output, "Did not find exactly 1 archive:")
+
+	sourceDb = "src-journ"
```
adriansuarez, line 1042: This is the same test case as above but with a separate journal?

test/testlib/nuodb_database_utilities.go
```diff
@@ -747,3 +758,18 @@ func SuspendDatabaseBackupJobs(t *testing.T, namespaceName string, domain, dbNam
...
+func SnapshotVolume(t *testing.T, namespaceName string, pvcName string, snapName string) {
+	kubectlOptions := k8s.NewKubectlOptions("", "", namespaceName)
+	k8s.KubectlApplyFromString(t, kubectlOptions, fmt.Sprintf(SNAPSHOT_TEMPLATE, snapName, VOLUME_SNAPSHOT_CLASS, pvcName))
+
+	AddTeardown(TEARDOWN_SNAPSHOT, func() {
+		k8s.RunKubectl(t, kubectlOptions, "delete", "volumesnapshot", snapName)
+	})
+
+	Await(t, func() bool {
+		out, error := k8s.RunKubectlAndGetOutputE(t, kubectlOptions, "get", "volumesnapshot", snapName, "-o", "custom-columns=:.status.readyToUse", "--no-headers")
```
adriansuarez, line 772: Why don't you use `-o jsonpath='{.status.readyToStatus}'`?

test/minikube/minikube_base_restore_test.go
```diff
@@ -436,3 +435,91 @@ func TestKubernetesAutoRestore(t *testing.T) {
...
+	archiveSnapshotName := "archive-snapshot"
+	testlib.SnapshotVolume(t, namespaceName, achiveVolumeName, archiveSnapshotName)
+
+	journalVolumeName := "journal-volume-" + smPod
+	journalSnapshotName := "journal-snapshot"
+	testlib.SnapshotVolume(t, namespaceName, journalVolumeName, journalSnapshotName)
+
+	helm.DeleteE(t, &helm.Options{KubectlOptions: k8s.NewKubectlOptions("", "", namespaceName)}, sourceDatabaseChartName, true)
+
+	restoredDb := "db-clone"
```
adriansuarez, line 498: Is there a test where we exercise creation of a database with the same name as the snapshotted database?

stable/database/files/nuosm
kontaras: I really like the idea of having it be a JSON file, for future extension. I was a bit in denial about how brittle the regex would be.

You are right that it is probably too much complexity for no immediate benefit.

stable/database/files/nuosm
kontaras: Will do.

stable/database/files/nuosm
kontaras: Will do.

stable/database/files/nuosm
kontaras: Will do

stable/database/files/nuosm
kontaras: This is deliberate. We found a non-empty archive directory. It likely means that we this is just a regular SM restart and not a first boot.

If there is no backup id, that could mean that the archive was already recovered. We have, at least, deleted the info.json.

If there is a mismatched backup id, that could mean that we crashed while a snapshot was being taken.

stable/database/files/nuosm
kontaras: I like this as a debug log. I will remove it.

stable/database/templates/statefulset.yaml
kontaras: I chose the name to match the environment variable. You are right that it does not match the other helm values. Fixing.

stable/database/templates/statefulset.yaml
kontaras: I like this idea. Will do.

test/minikube/minikube_base_restore_test.go
kontaras: Sure thing.

test/minikube/minikube_long_restore_test.go
kontaras: Will do.

test/minikube/minikube_long_restore_test.go
kontaras: Kind of. These are tests for recovering journals and mis-configured journals.

test/testlib/nuodb_database_utilities.go
kontaras: I did not use it because kubectl needs better docs.

test/minikube/minikube_base_restore_test.go
kontaras: No. Do we have a good way to clear out left over PVCs?

kontaras:
> @kontaras Can you add a commented out example to `stable/database/values.yaml` and descriptions of the new Helm values to `stable/database/README.md`?

I thought that this was going to be an undocumented feature that only our code will use.

adriansuarez:
> @adriansuarez @sivanov-nuodb I thought that this was going to be an undocumented feature that only our code will use.

We should still document it so that we know how to use it. But either way, if it is exposed in the Helm charts then it is available for use. People like @dbutson-mdsol and @acabrele will also appreciate documentation on the values.

test/minikube/minikube_base_restore_test.go
adriansuarez: The PVC names should be predictable based on the archive names appearing in `volumeClaimTemplates`. You can just delete `"archive-volume-" + smPodName`.

sivanov-nuodb (commented):
@kontaras Can you please add the new Helm values in `values.yaml` and in README.md file.

test/minikube/minikube_long_restore_test.go
```diff
@@ -891,3 +891,220 @@ func TestKubernetesRestoreDatabaseSeparateJournal(t *testing.T) {
 		testlib.CheckRestoreRequests(t, namespaceName, admin0, opt.DbName, "", "")
 	})
 }
+
+// Extensive end to end tests for creating a database off of a VolumeSnapshot
+func TestCornerCaseKubernetesSnapshotRestore(t *testing.T) {
```
sivanov-nuodb, line 896: How much time this test takes?

stable/database/files/nuosm
kontaras: > if backup.json exist && $BACKUP_ID is not empty => verify the backup ID and fail on mismatch

That is a valid state of the SM, likely caused by a bug in our pre-snap logic.

For example:

1. A database is started from a backup with backup id `1`.
2. The SM starts up successfully and begins serving a database.
3. Some time later, a backup policy kicks in and triggers a snapshot.
4. `backup.json` is created with an id of `2`.
5. Due to a bug in the `fsfreeze` sidecar, the pod crashes.
6. Kubernetes sees a failed pod and restarts it.
7. `nuosm` starts with a Helm supplied $BACKUP_ID of `1` but `backup.json` says the id should be `2`.

The SM still has a valid archive and is registered with the Admin. It just needs to start and will be able to resume serving up the database.

go.mod
```diff
@@ -3,21 +3,15 @@ module github.com/nuodb/nuodb-helm-charts/v3
 go 1.14
```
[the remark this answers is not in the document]
kontaras: Will do.

stable/database/files/nuosm
```diff
@@ -387,6 +387,69 @@ function resolveLatestSource() {
   return 0
 }
 
+#=======================================
+# function - Looks for a previous archive and journal and imports it
+#
+function loadFromSnapshot() {
+  local RECREATE_ARCHIVES="false"
```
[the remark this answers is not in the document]
kontaras: I was not aware of that convention.

stable/database/files/nuosm
```diff
@@ -387,6 +387,69 @@ function resolveLatestSource() {
...
+    fi
+
+    # Don't move the archive until after we have done error checking.
+    # Once we move the archive, we are going to be in the `archive already exists` branch
+    # when kubernetes retries/restarts the pod.
+    mv ${ARCHIVE_PATH}/* ${DB_DIR}
+
+    local RECREATE_ARCHIVES="true"
+  else
+    log "archive already exists, not looking for snapshot archives"
```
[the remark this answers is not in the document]
kontaras: Sure

test/integration/template_database_test.go
```diff
@@ -2517,3 +2517,86 @@ func TestDatabaseTLSConfig(t *testing.T) {
...
+			"database.autoImport.backup_id":             "123abc",
+		}
+
+		setFields := func(specKey string, label string) {
+			for _, field := range []string{"apiGroup", "kind", "name", "namespace"} {
+				values[specKey+".dataSourceRef."+field] = label + field
+			}
+		}
+
+		check := func(specKey string, sourceDef *v1.TypedObjectReference) {
```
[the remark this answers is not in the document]
kontaras: Yes

test/minikube/minikube_base_restore_test.go
```diff
@@ -436,3 +435,91 @@ func TestKubernetesAutoRestore(t *testing.T) {
...
+			"database.persistence.storageClass":                        testlib.SNAPSHOTABLE_STORAGE_CLASS,
+			"database.sm.hotCopy.journalPath.persistence.storageClass": testlib.SNAPSHOTABLE_STORAGE_CLASS,
+			"database.sm.hotCopy.journalPath.enabled":                  "true",
+		},
+	})
+
+	output, err := k8s.RunKubectlAndGetOutputE(t, kubectlOptions, "exec", admin0, "-c", "admin", "--", "bash", "-c",
+		"echo \"CREATE TABLE testtbl (id INT); INSERT INTO testtbl (id) values (123);\" | nuosql demo --user dba --password secret")
+
+	require.NoError(t, err, output)
```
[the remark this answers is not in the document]
kontaras: Good to know, I will use it.

test/minikube/minikube_base_restore_test.go
```diff
@@ -436,3 +435,91 @@ func TestKubernetesAutoRestore(t *testing.T) {
...
+		"echo \"{\\\"id\\\": \\\"123abc\\\"}\" > /var/opt/nuodb/archive/nuodb/demo/backup.json")
+	require.NoError(t, err, output)
+
+	output, err = k8s.RunKubectlAndGetOutputE(t, kubectlOptions, "exec", smPod, "-c", "engine", "--", "bash", "-c",
+		"echo \"{\\\"id\\\": \\\"123abc\\\"}\" > /var/opt/nuodb/journal/nuodb/demo/backup.json")
+	require.NoError(t, err, output)
+
+	defer testlib.Teardown(testlib.TEARDOWN_SNAPSHOT)
+
+	achiveVolumeName := "archive-volume-" + smPod
```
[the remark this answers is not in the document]
kontaras: Right, renaming `volume` to `volumeClaim` would be more accurate.

test/minikube/minikube_base_restore_test.go
```diff
@@ -436,3 +435,91 @@ func TestKubernetesAutoRestore(t *testing.T) {
...
+	archiveSnapshotName := "archive-snapshot"
+	testlib.SnapshotVolume(t, namespaceName, achiveVolumeName, archiveSnapshotName)
```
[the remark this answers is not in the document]
kontaras: Once we have backup logic in place to drop in the backup id and run fsfreeze, the test will need to be modified to use it. I will note that in a comment.

test/minikube/minikube_long_restore_test.go
```diff
@@ -891,3 +891,220 @@ func TestKubernetesRestoreDatabaseSeparateJournal(t *testing.T) {
...
+
+			smPod := fmt.Sprintf("sm-%s-nuodb-cluster0-%s-hotcopy-0", databaseChartName, dbName)
+			var pod *corev1.Pod
+			testlib.Await(t, func() bool {
+				var err error
+				pod, err = k8s.GetPodE(t, kubectlOptions, smPod)
+				return err == nil && len(pod.Status.ContainerStatuses) > 0 &&
+					pod.Status.ContainerStatuses[0].State.Terminated != nil &&
+					pod.Status.ContainerStatuses[0].State.Terminated.Reason == "Error"
+			}, 60*time.Second)
```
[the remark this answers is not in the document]
kontaras: Does the restart count get incremented when the pod crashes or after k8s restarts it? I want to wait for the pod to be in a crashed state and grab the logs from before the crash.

Also, doing it this way initializes the `pod` variable correctly.

test/minikube/minikube_long_restore_test.go
kontaras: Around 10 minutes on CircleCI. Is there some scenarios you would recommend cutting?

adriansuarez: Can you try using `t.Run("testSpecificScenario", ...)` so that test failures are easier to locate? The Golang testing framework doesn't have setup and teardown hooks, but you can define test cases in a hierarchical fashion in that way. Not sure if you would have to do something special in each test case to leave the environment in a state that the subsequent sub-tests can run despite a failure.

test/minikube/minikube_base_restore_test.go
```diff
@@ -436,3 +435,91 @@ func TestKubernetesAutoRestore(t *testing.T) {
...
+	defer testlib.Teardown(testlib.TEARDOWN_DATABASE)
+
+	sourceDatabaseChartName := testlib.StartDatabase(t, namespaceName, admin0, &helm.Options{
+		SetValues: map[string]string{
+			"database.sm.resources.requests.cpu":                       testlib.MINIMAL_VIABLE_ENGINE_CPU,
+			"database.sm.resources.requests.memory":                    testlib.MINIMAL_VIABLE_ENGINE_MEMORY,
+			"database.te.resources.requests.cpu":                       testlib.MINIMAL_VIABLE_ENGINE_CPU,
+			"database.te.resources.requests.memory":                    testlib.MINIMAL_VIABLE_ENGINE_MEMORY,
+			"database.persistence.storageClass":                        testlib.SNAPSHOTABLE_STORAGE_CLASS,
+			"database.sm.hotCopy.journalPath.persistence.storageClass": testlib.SNAPSHOTABLE_STORAGE_CLASS,
```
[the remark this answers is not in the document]
kontaras: I was trying to minimize the amount of Helm values I was passing in to only those strictly necessary for the test to run. If I move away from that, it is better to just use `noHotCopy` SMs.

test/minikube/minikube_long_restore_test.go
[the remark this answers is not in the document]
adriansuarez: Are you saying that there's a specific time window when you can fetch the logs? Have you observed this as an issue?

The `GetPodLogs()` function supplies `--previous` to get logging from the last invocation, so I think it is fine to wait until it gets restarted.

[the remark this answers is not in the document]
kontaras: I have not observed this issue. I suspect that their is a (generous) time window, but I have never hit it.

terratest does not support the `--previous` flag. Other places in the code, we call the k8s go bindings directly.

stable/database/files/nuosm
sivanov-nuodb: I feel like if the user is providing a `backup_id`, then we should verify it and fail on mismatch. If the user wants to disregard the validation, they can skip the `backup_id` altogether.

kontaras: So how do we differentiate the SM starting for the first time vs. the SM being restarted (and the SM being restarted while a snapshot is in progress)?

stable/database/files/nuosm
kontaras: I prefer checking for `1.cat` instead of `info.json` because it help in a very weird race condition: if the SM is being restored from a snapshot, the archive is moved, and the `info.json` is deleted, but nuosm crashes before it can register the archive with the admin and write a new `info.json`

If we look for `info.json`, the SM will be left in a state where it is trying to find an archive to copy, cannot find another one in the volume, and crashes.

If we look for `1.cat`, the SM will start properly, though there is a small risk that the archive ends up registered twice with the admin.

Honestly, it is a 1 in a million chance that this specific timing will ever actually happen, but I am one to take defensive programming a bit too far.

stable/database/files/nuosm
sivanov-nuodb: We've discussed this offline and the issue that Kon is trying to solve here is that the `info.json` _shouldn't be removed_ on container restart in case the post-backup hook didn't remove the `backup.json` for some reason.
I think in case of a backup ID mismatch (which is an abnormal scenario), we should at least _log_.

test/minikube/minikube_long_restore_test.go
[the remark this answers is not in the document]
sivanov-nuodb: > Does the restart count get incremented when the pod crashes or after k8s restarts it

Yes, once the container crash the restart count is incremented. You could use our own helpers to assert that a message appears in the SM logs:

```golang
testlib.AwaitPodRestartCountGreaterThan(t, namespaceName, smPod, 1, 60*time.Second)
require.GreaterOrEqual(t, testlib.GetStringOccurrenceInLog(t, namespaceName, smPod,
				expectedMsg, &corev1.PodLogOptions{Previous: true}), 1)
```

stable/database/files/nuosm
kontaras: @sivanov-nuodb Non-LSA archives don't always have an `1.cat`. Talking to Paul S, he suggested looking for `state.dat`.

adriansuarez: Agreed. We should also update the code snippet I posted. I created a JIRA for the `nuodocker start sm` change.

sivanov-nuodb: @agoelzer recently changed [nuosm](https://github.com/nuodb/nuodb-helm-charts/blob/master/stable/database/files/nuosm#L497) to check for `1.cat`. It seems that this needs to be changed as well.

test/minikube/minikube_long_restore_test.go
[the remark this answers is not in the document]
kontaras: Interestingly, I am seeing the opposite race condition: when I try to check the logs, the newly crashed container may not yet be the "previous".

stable/database/values.yaml
```diff
@@ -390,6 +393,9 @@ database:
           accessModes:
             - ReadWriteOnce
           # storageClass: "-"
+          dataSourceRef:
```
adriansuarez, line 396: Can you a comment with a link to the Kubernetes documentation so that we know what the schema is what this is used for, e.g.
```
# The data source for the archive PVC
# see https://kubernetes.io/docs/concepts/storage/persistent-volumes/#data-source-references
```

We should also include these in the README. We want to be comprehensive in our descriptions in the README so that people know how to supply these values.

stable/database/values.yaml
```diff
@@ -431,6 +439,10 @@ database:
           accessModes:
             - ReadWriteOnce
           # storageClass: "-"
+
+          # The data source for the archive PVC
```
adriansuarez, line 443: This is the journal PVC.

stable/database/values.yaml
```diff
@@ -393,6 +397,10 @@ database:
...
```
adriansuarez, line 401: This is the journal PVC for hot copy SMs.

kontaras: Oops. A good reason to not make one more change after a log day of debugging.

stable/database/values.yaml
sivanov-nuodb, line 1: Can you please add the `database.autoImport.backupId` to the _values.yaml_ with a default value and description?

kontaras: Yes, documentation still needs updating. I am getting the code/tests right first since those take longer to iterate.

stable/database/files/nuosm
```diff
@@ -387,6 +387,82 @@ function resolveLatestSource() {
...
+    rm -r ${DB_DIR}
+    mv ${archive_path} ${DB_DIR}
+
+    local recreate_archives="true"
+  else
+    trace "archive already exists, not looking for snapshot archives"
+
+    local backup_file=${DB_DIR}/backup.txt
+    if [ -n "$BACKUP_ID" ] && [ -f  $backup_file ] ; then
+      if [ "$(cat $backup_file | tr -d [:space:])" == "$BACKUP_ID" ] ; then
```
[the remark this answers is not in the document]
kontaras: So we can have the `else` branch which logs that we have both a `BACKUP_ID` and a `backup_file` but they do not match. `checkBackupId` does not differentiate if the `backup_file` is missing or present and has the wrong value.

stable/database/files/nuosm
```diff
@@ -387,6 +387,82 @@ function resolveLatestSource() {
...
+        die 1 "Did not find a journal snapshot at '$expected_journal'"
+      fi
+      
+      if [ -n "$BACKUP_ID" ]; then
+        if ! checkBackupId ${expected_journal}/backup.txt ; then
+          die 1 "Incorrect backup id in journal"
+        fi
+      fi
+
+      rm -r ${JOURNAL_DIR}
```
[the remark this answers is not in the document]
kontaras: `rm -f` always feels unsafe to me, but I agree that it is warranted here.

sivanov-nuodb (approved):
Looks good.
I left some more minor comments.

stable/database/README.md
```diff
@@ -254,6 +256,7 @@ The following tables list the configurable parameters of the `database` chart an
 | `sm.hotCopy.persistence.size` | size of the hotcopy storage PV | `20Gi` |
 | `sm.hotCopy.persistence.accessModes` | access modes for the hotcopy storage PV | `[ ReadWriteOnce ]` |
 | `sm.hotCopy.persistence.size` | size of the hotcopy storage PV | `20Gi` |
+| `sm.hotCopy.journalPath.persistence.dataSourceRef` | Data Source to initialize the journal volume. See below.  |  |
```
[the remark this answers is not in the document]
kontaras: I omitted it because none of the other `sm.noHotCopy.persistence.*` values are documented. I will add this one.

stable/database/README.md
```diff
@@ -335,6 +338,21 @@ The purpose of this section is to allow customisation of the names of the cluste
...
+
+Define a VolumeSnapshot, PersistentVolumeClaim, or other dataSource from which to initialize a database volume.
+
+See [Persistent Volume Doumentation](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#data-source-references) and [Resource Schema](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#typedobjectreference-v1-core)
+
+When taking a Volume Snapshot, you can drop a file called `backup.txt` in the directory root containing an arbitrary value. That value should be passed in via `database.autoImport.backupId`. Required if restoring a database with the same domain and database names, optional otherwise.
+
+| Key | Description | Default |
+| ----- | ----------- | ------ |
+| `name` | Backup or volume name. dataSource will be omitted if `nil` | `nil` |
```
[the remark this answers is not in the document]
kontaras: Good idea

stable/database/files/nuosm
```diff
@@ -387,6 +387,82 @@ function resolveLatestSource() {
...
+      else
+        log "Archive present but with a wrong backupId. Did the SM crash during a snapshot?"
+      fi
+    fi
+  fi
+
+  if [ $recreate_archives == "true" ]; then
+    log "Removing outdated archive metadata"
+    rm ${DB_DIR}/info.json ${DB_DIR}/backup.txt
+    [ -e ${JOURNAL_DIR}/backup.json ] && rm ${JOURNAL_DIR}/backup.txt
```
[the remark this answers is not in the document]
kontaras: Thank you for catching that

stable/database/files/nuosm
```diff
@@ -387,6 +387,82 @@ function resolveLatestSource() {
...
+        local recreate_archives="true"
+      else
+        log "Archive present but with a wrong backupId. Did the SM crash during a snapshot?"
+      fi
+    fi
+  fi
+
+  if [ $recreate_archives == "true" ]; then
+    log "Removing outdated archive metadata"
+    rm ${DB_DIR}/info.json ${DB_DIR}/backup.txt
```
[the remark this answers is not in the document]
kontaras: What `backup_file` variable? What would I use if for? That variable is only set if the expected archive exists, not if we have to move the archive.

stable/database/templates/statefulset.yaml
```diff
@@ -315,6 +317,10 @@ spec:
         matchLabels:
           database: {{ .Values.database.name }}
 {{- end }}
+      {{- if .Values.database.persistence.dataSourceRef.name }}
+      dataSourceRef:
+        {{- toYaml .Values.database.persistence.dataSourceRef | nindent 8}}
+      {{- end }}
```
adriansuarez, line 323: I think we may want to allow the `dataSourceRef` to be optional as well, to be compatible with versions of `values.yaml`. That can be implemented by wrapping a second conditional around this as follows:

```
      {{- if .Values.database.persistence.dataSourceRef }}
      {{- if .Values.database.persistence.dataSourceRef.name }}
      dataSourceRef:
        {{- toYaml .Values.database.persistence.dataSourceRef | nindent 8}}
      {{- end }}
      {{- end }}
```

test/minikube/minikube_base_restore_test.go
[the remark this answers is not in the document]
adriansuarez: What was the resolution here? This is still HCSM and cronjobs are still enabled.

It makes sense to disable cronjobs whenever they are not needed because they may unexpectedly generate pods that cause the "are all NuoDB pods gone?" checks to fail.

test/minikube/minikube_base_restore_test.go
adriansuarez: Is this implemented in a different test case?

test/minikube/minikube_base_restore_test.go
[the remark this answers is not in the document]
kontaras: I changed the long tests to use non-HC SMs. I did not do the same for the short test. I will do so.

adriansuarez (approved):
Looks good.

stable/database/templates/statefulset.yaml
kontaras: Good idea, I will fix the logic and add a template test for the scenario.

PR #339, opened 2023-12-08
[HELM-416] ignore completed pods in scheduled pod list

agoelzer:
test case is failing because it is expecting one pod but finds two - one is in Running state (the valid one) while the other one is in Completed state and has a similar name (contains an additional (*crd-job-1*) in the pod name)

### DIFF
--- a/test/testlib/minikube_utilities.go
+++ b/test/testlib/minikube_utilities.go
@@ -451,6 +451,11 @@ func AwaitNrReplicasScheduled(t *testing.T, namespace string, expectedName strin
 		var podNames string
 		for _, pod := range FindAllPodsInSchema(t, namespace) {
 			if strings.Contains(pod.Name, expectedName) {
+				//ignore all completed pods
+				if pod.Status.Phase == corev1.PodSucceeded {
+					continue;
+				}
+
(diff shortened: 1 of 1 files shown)


test/testlib/minikube_utilities.go
```diff
@@ -451,6 +451,11 @@ func AwaitNrReplicasScheduled(t *testing.T, namespace string, expectedName strin
 		var podNames string
 		for _, pod := range FindAllPodsInSchema(t, namespace) {
 			if strings.Contains(pod.Name, expectedName) {
+				//ignore all completed pods
+				if pod.Status.Phase == corev1.PodSucceeded {
+					continue;
```
adriansuarez, line 456: Perhaps we should log the name of the completed pod, since this check is looking for continuously running pods in a workload and it is strange to find a completed job.

This is due to HAProxy introducing some kind of job that checks something having to do with their CRDs: https://github.com/haproxytech/helm-charts/commit/9baeddf7659ac9ea05cbc312db78ec3a199e40d7

adriansuarez (approved):
@agoelzer: Thanks for jumping on this and figuring out what the issue is so quickly.

PR #341, opened 2023-12-12
Don't start CronJobs when testing upgrade from 3.3 or 3.4

kontaras:
In order to be able to create volumes from volume snapshots, we need to update out version of kubernetes. Unfortunately, newer versions have changed `CronJob` `apiVersion` from `v1beta1` to `v1`. Helm charts versions 3.5.0 and newer handle it properly (using Capabilities) but 3.3 and 3.4 do not.

In order to allow upgrade tests to run on current versions of kubernetes, do not create backup CronJobs when testing old versions.


sivanov-nuodb (approved):
Looks good.

adriansuarez (approved):
Looks good.

PR #346, opened 2023-12-29
Add backup hooks sidecar and validate data sources

adriansuarez:
This change adds a sidecar that exposes HTTP endpoints for pre- and post-backup hooks to be invoked before and after snapshotting the archive and journal volumes. This is implemented using the new backup_hooks.py script, which is mounted into an official Python 3 image as a configmap.

The backup_hooks.py script controls the lifecycle of a backup by creating the backup.txt file containing the supplied backup ID when the `POST /pre-backup/{backupId}` endpoint is invoked. This is used by the `nuosm` entrypoint script to detect that an archive and journal are being started for the first time from snapshots.

If the journal volume is separate from the archive volume, then coordination has to be performed to make sure that the snapshots of the archive and journal volume are consistent with each other. By default, this is done by invoking `fsfreeze --freeze` on the archive volume. To enable testing and to support environments where privileged containers are not allowed, it is also possible to obtain consistent snapshots by suspending the nuodb process while snapshots are being taken, which is enabled by the `database.backupHooks.useSuspend` Helm value.

The `POST /post-backup/{backupId}` endpoint unfreezes write to the archive or resumes the nuodb process, and deletes the backup.txt file. It is also possible to directly invoke the backup hooks by using `kubectl exec` instead of issuing HTTP requests.

This commit also makes a couple of changes to the way data sources are specified to reduce the potential for misconfiguration. The backup ID is now nested within `database.snapshotRestore`, which also includes `snapshotNameTemplate`, which allows the user to define a naming convention that is used to resolve the snapshots to use as data sources.

The archive and journal data sources can still be specified explicitly to override the data sources resolved from
`database.snapshotRestore.snapshotNameTemplate`, but it is no longer possible to supply the hotcopy and non-hotcopy SMs with different data sources, since that makes no sense (both SM statefulsets use the same archive data source, so if the journal data sources are different, at least one of them has to be invalid).

Finally, validation is now performed by default on the resolved data sources when installing or upgrading a Helm release, to make sure that the data sources exist. This is necessary because Kubernetes silently ignores non-existent data sources and allows the volumes to be initialized with no data, which will lead to confusing behavior.


test/integration/template_database_test.go
```diff
@@ -2531,85 +2671,169 @@ func TestDatabaseStatefulSetVolumeSnapshot(t *testing.T) {
 			for _, template := range obj.Spec.VolumeClaimTemplates {
 				assert.Nil(t, template.Spec.DataSourceRef)
 			}
-			assert.False(t, testlib.EnvContains(obj.Spec.Template.Spec.Containers[0].Env, "SNAPSHOT_RESTORED", "true"))
+			assert.True(t, testlib.EnvContains(obj.Spec.Template.Spec.Containers[0].Env, "SNAPSHOT_RESTORED", "false"))
```
sivanov-nuodb, line 2674: Was that a bug?

test/integration/template_database_test.go
```diff
@@ -2531,85 +2671,169 @@ func TestDatabaseStatefulSetVolumeSnapshot(t *testing.T) {
...
+				"database.persistence.journalDataSource.name":      "journal-volume-otherdb-0",
+				"database.persistence.journalDataSource.namespace": "otherns",
+				"database.persistence.journalDataSource.apiGroup":  "null", // Unset the default
+				"database.sm.hotCopy.journalPath.enabled":          "true",
+			},
+		}
+
+		_, err := helm.RenderTemplateE(t, options, helmChartPath, "release-name", []string{"templates/statefulset.yaml"})
+		assert.NotNil(t, err)
+		assert.Contains(t, err.Error(), "Invalid data source: PersistentVolumeClaim/journal-volume-otherdb-0 not found in namespace otherns")
```
sivanov-nuodb, line 2826: I assume that we can't test the opposite scenario in the _template_ tests as we don't have a real K8s API server?

stable/database/templates/configmap.yaml
```diff
@@ -37,6 +37,15 @@ data:
 ---
 apiVersion: v1
 kind: ConfigMap
+metadata:
+  labels:
+    {{- include "database.resourceLabels" . | nindent 4 }}
+  name: {{ template "database.fullname" . }}-backup-hooks
+data:
+{{ (.Files.Glob "files/backup_hooks.py").AsConfig | indent 2 }}
```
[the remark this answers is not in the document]
adriansuarez: Good idea.

stable/database/templates/statefulset.yaml
```diff
@@ -260,8 +257,49 @@ spec:
...
+        env:
+          - name: USE_SUSPEND
+            value: {{ include "defaultfalse" .Values.database.backupHooks.useSuspend | quote }}
+          - name: NUODB_ARCHIVE_DIR
+            value: {{ printf "/mnt/archive/%s/%s" .Values.admin.domain .Values.database.name }}
+          {{- if eq (include "defaultfalse" .Values.database.sm.noHotCopy.journalPath.enabled) "true" }}
+          - name: NUODB_JOURNAL_DIR
+            value: {{ printf "/mnt/journal/%s/%s" .Values.admin.domain .Values.database.name }}
+          {{- end }}
+        resources: {}
```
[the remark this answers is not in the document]
adriansuarez: Good point. That was an oversight.

test/integration/template_database_test.go
adriansuarez: Before it was checking for the absence of the env variable. Now it is explicitly set to false.

test/integration/template_database_test.go
adriansuarez: Correct, there is no K8s cluster to test against.

stable/database/files/backup_hooks.py
```diff
@@ -0,0 +1,387 @@
...
+            action = "--unfreeze"
+        else:
+            LOGGER.info("Freezing writes to archive volume")
+            action = "--freeze"
+        subprocess.check_output(
+            ["fsfreeze", action, ARCHIVE_DIR], stderr=subprocess.STDOUT
+        )
+
+
+def pre_backup(backup_id, payload):
```
kontaras, line 153: Could we have a background thread that automatically calls unfreeze after a long timeout in case the snapshot code crashes?

test/minikube/minikube_long_restore_test.go
```diff
@@ -1010,8 +1000,6 @@ func TestCornerCaseKubernetesSnapshotRestore(t *testing.T) {
 
 	// Test restoring an archive with a journal embedded
 	t.Run("testArchive", func(t *testing.T) {
-		// TODO: Once it exists, use proper database snapshot function to freeze database, set backup id, and create volume snapshots
-
```
kontaras, line 1014: You removed the TODO, but did not actually implement it. Snapshotting a running SM is not likely to break the test, but it is safer to do it right. Also, it would be nice to reduce how many places in the code hard code the location and format of `backup.txt`.

stable/database/files/backup_hooks.py
```diff
@@ -0,0 +1,387 @@
...
+        for pid in pids:
+            if unfreeze:
+                LOGGER.info("Resuming nuodb process with pid %s", pid)
+                os.kill(int(pid), signal.SIGCONT)
+            else:
+                LOGGER.info("Suspending nuodb process with pid %s", pid)
+                os.kill(int(pid), signal.SIGSTOP)
+                # Make sure all threads are suspended
+                for pid in pids:
+                    await_suspended(pid)
```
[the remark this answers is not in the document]
adriansuarez: Good catch.

test/minikube/minikube_long_restore_test.go
adriansuarez: The tests is exercising negative cases for `nuosm` where the archive is in an unexpected state. The backup hooks will not leave the archive in a state that can exercise these cases, so I don't think the TODO is relevant here.

stable/database/files/backup_hooks.py
adriansuarez: Makes sense.

stable/database/values.yaml
```diff
@@ -315,6 +324,35 @@ database:
...
+
+    # If the non-hotcopy SM is configured with separate archive and journal
+    # volumes, writes are frozen to the archive volume when the pre-backup hook
+    # is invoked and unfrozen when the post-backup hook is invoked. If
+    # useSuspend=true, then the SM process is suspended completely rather than
+    # only writes to the archive being frozen. This is to enable usage in
+    # environments where fsfreeze cannot be invoked on the archive filesystem,
+    # e.g. because it is using hostpath-based storage where other processes
+    # would be affected.
+    useSuspend: false
```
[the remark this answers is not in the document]
adriansuarez: I'll add a note in the README.

stable/database/files/backup_hooks.py
adriansuarez: This is a good suggestion, but I don't want to make it part of this change. I'll create a follow-on JIRA for this improvement.

stable/database/README.md
```diff
@@ -339,21 +342,6 @@ The purpose of this section is to allow customisation of the names of the cluste
...
-
-When taking a Volume Snapshot, you can drop a file called `backup.txt` in the directory root containing an arbitrary value. That value should be passed in via `database.autoImport.backupId`. Required if restoring a database with the same domain and database names, optional otherwise.
-
-| Key | Description | Default |
-| ----- | ----------- | ------ |
-| `name` | Backup or volume name. The entire dataSource will be omitted if this value is empty | `nil` |
-| `namespace` | Namespace containing the source | `nil` |
-| `kind` | Data source kind | `VolumeSnapshot` |
-| `apiGroup` | APIGroup is the group for the resource. If APIGroup is not specified, the specified Kind must be in the core API group. | `snapshot.storage.k8s.io` |
-
```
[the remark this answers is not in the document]
adriansuarez: I restored and updated the content.

adriansuarez:
@kontaras, @sivanov-nuodb: I updated the PR based on your comments. Have another look.

test/minikube/minikube_long_restore_test.go
kontaras: Out of the 8 snapshots, 4 are ones that `SnapshotSm` can handle (correct and incorrect `backupid` and with and without a journal). With the other 4, you need to muck around with the SM after the `pre-backup` script but before taking the snapshot: for 2 of them, you delete `backup.txt`, one with the archive directory moved, and one with a duplicate journal.

test/testlib/nuodb_database_utilities.go
```diff
@@ -786,3 +789,77 @@ func SnapshotVolume(t *testing.T, namespaceName string, pvcName string, snapName
...
+	Message string `json:"message"`
+}
+
+func GetBackupHookResponse(t *testing.T, namespaceName, smPod, resourcePath string) *BackupHookResponse {
+	// Issue request to backup hook server. The nuodb image has curl, so use
+	// it in the SM container to issue HTTP request against the backup hook
+	// server running in a sidecar container on the same pod.
+	kubectlOptions := k8s.NewKubectlOptions("", "", namespaceName)
+	output, err := k8s.RunKubectlAndGetOutputE(t, kubectlOptions, "exec", smPod, "-c", "engine", "--",
+		"curl", "-s", "-X", "POST", "http://0.0.0.0:8000/"+resourcePath)
```
sivanov-nuodb, line 822: In the real scenario, I presume that we need to use `http://<POD_IP>:8080` URL to invoke the backup hooks?

stable/database/files/backup_hooks.py
```diff
@@ -0,0 +1,386 @@
...
+    # archive and journal directories are frozen
+    if os.path.exists(ARCHIVE_BACKUP_ID_FILE):
+        raise UserError(
+            "Backup ID file {} exists. Execute post-backup hook to complete current backup.".format(
+                ARCHIVE_BACKUP_ID_FILE
+            )
+        )
+
+    # Write backup ID to archive directory
+    write_file(ARCHIVE_BACKUP_ID_FILE, backup_id)
```
sivanov-nuodb, line 164: Should we validate that the supplied {{backup_id}} is valid (i.e not `None`)?

test/minikube/minikube_fsfreeze_test.go
```diff
@@ -0,0 +1,228 @@
...
+	}, 50*time.Millisecond)
+
+	// Check that backup hook sidecar logged expected messages
+	output, err := k8s.RunKubectlAndGetOutputE(t, kubectlOptions, "logs", smPod, "-c", "backup-hooks")
+	require.NoError(t, err, output)
+	require.Contains(t, output, "Freezing writes to archive volume")
+	require.Contains(t, output, "Unfreezing writes to archive volume")
+
+	t.Run("negativeTests", func(t *testing.T) {
+		// Negative test: use a bogus resource path
```
sivanov-nuodb, line 201: Can you also add a negative test with multiple path components (for example `/pre-backup/backupId/bogus`)?

sivanov-nuodb (approved):
Looks great!
Thanks for figuring out how to test the fsfreeze with hostpath CSI driver.

stable/database/templates/statefulset.yaml
```diff
@@ -260,8 +257,50 @@ spec:
           # the SM becomes unready if it does not start within 15 minutes = 30s + 15s*58
           successThreshold: 2
           timeoutSeconds: {{ default 5 .Values.database.sm.readinessTimeoutSeconds }}
-      {{- include "nuodb.sidecar" (list . .Values.database.sm) | nindent 6 }}
-{{- include "nuodb.imagePullSecrets" . | indent 6 }}
+      {{- if eq (include "defaultfalse" .Values.database.backupHooks.enabled) "true" }}
+      - name: backup-hooks
```
sivanov-nuodb, line 261: Should we add `healthz` handler and a readiness probe for the container?

stable/database/files/backup_hooks.py
```diff
@@ -0,0 +1,386 @@
...
+            prefix = "State:"
+            for line in f:
+                if line.startswith(prefix):
+                    state = line[len(prefix) :].strip()
+                    if state != "T (stopped)":
+                        LOGGER.info("Thread %s not stopped: %s", tid, state)
+                        return False
+                    return True
+    except FileNotFoundError as e:
+        LOGGER.warning("Unable to get state of thread %s: %s", tid, e)
```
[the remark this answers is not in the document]
adriansuarez: The caller doesn't use this, but `None` indicates that the thread status is unknown.

stable/database/files/backup_hooks.py
```diff
@@ -0,0 +1,386 @@
...
+            action = "--freeze"
+        subprocess.check_output(
+            ["fsfreeze", action, ARCHIVE_DIR], stderr=subprocess.STDOUT
+        )
+
+
+def pre_backup(backup_id, payload):
+    # Make sure we do not attempt to execute pre-backup hook if a backup is
+    # already in progress, which may block indefinitely if writes to the
+    # archive and journal directories are frozen
```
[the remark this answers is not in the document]
adriansuarez: I think the main loop directly invokes the request handler, so everything should execute in the main thread.

kontaras (approved):

test/testlib/nuodb_database_utilities.go
adriansuarez: Yes, it should work with `http://<pod IP>:8000` assuming that you have access to the internal network. I didn't exercise it that way because I don't think the pod IPs are routable in the test.

stable/database/files/backup_hooks.py
adriansuarez: Good idea. The request dispatcher checks that the correct number of path components are specified, but the user can still supply an empty backup ID which we should prevent.

PR #348, opened 2024-01-05
Update DNS configuration before multi-cluster tests

adriansuarez:
The IPs for the elastic load balancer endpoints that are used to expose each DNS server are not permanent, so the CoreDNS configuration has to be updated to include the current IPs. This change programmatically resolves the IPs of each DNS load balancer and adds them as upstream resolvers for the other.

The "Corefile" configmap entry is assumed to contain the marker string "# BEGIN BASE CONFIG", which delimits the non-changing portion of the config and enables updates to the generated portion.

### DIFF
--- a/test/testlib/multicluster_utilities.go
+++ b/test/testlib/multicluster_utilities.go
@@ -3,9 +3,12 @@ package testlib
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"math"
+	"net"
+	"sort"
 	"strings"
 	"testing"
 	"time"
@@ -13,6 +16,7 @@ import (
 	"github.com/gruntwork-io/terratest/modules/helm"
 	"github.com/gruntwork-io/terratest/modules/k8s"
 	"github.com/stretchr/testify/require"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const CONTEXT_CLUSTER_KEY = CONTEXT_KEY("cluster")
@@ -150,6 +154,14 @@ func ChangeCluster(t *testing.T, cluster K8sCluster) (func(), func()) {
 	return changeFunc, changeBackFunc
 }
 
+func GetClusterDeploymentContext(t *testing.T, context context.Context) *ClusterDeploymentContext {
+	value := context.Value(CONTEXT_CLUSTER_KEY)
+	require.NotNil(t, value, "Unable to retrieve cluster deployment context")
+	deploymentContext, ok := value.(*ClusterDeploymentContext)
+	require.True(t, ok, "Unexpected type for cluster deployment context")
+	return deploymentContext
+}
+
 /**
  * Deploy NuoDB Helm charts using deployment context
  *
@@ -174,9 +186,7 @@ func ChangeCluster(t *testing.T, cluster K8sCluster) (func(), func()) {
  */
 func DeployWithContext(t *testing.T, context context.Context, deployFunc func(context *ClusterDeploymentContext, options *helm.Options)) {
 	// Retrieve cluster deployment information from the context
-	value := context.Value(CONTEXT_CLUSTER_KEY)
-	require.NotNil(t, value, "Unable to retrieve cluster deployment context")
-	deploymentContext := value.(*ClusterDeploymentContext)
+	deploymentContext := GetClusterDeploymentContext(t, context)
 	thisCluster := deploymentContext.ThisCluster
 	entrypointCluster := deploymentContext.EntrypointCluster
 	thisCluster = InjectClusters(t, thisCluster)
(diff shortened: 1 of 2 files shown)


test/testlib/multicluster_utilities.go
```diff
@@ -243,3 +253,97 @@ func AdjustPodTimeout(podName string, timeout time.Duration) time.Duration {
...
+	t.Logf("Adding DNS config snippet %s", dnsConfigSnippet)
+	cm.Data[COREFILE_KEY] = updatedConfig
+	_, err = clientset.CoreV1().ConfigMaps(COREDNS_NS).Update(ctx, cm, v1.UpdateOptions{})
+	require.NoError(t, err)
+}
+
+// UpdateDnsConfig adds the DNS server from one cluster as an upstream resolver
+// for the other, replacing any stale IP addresses that may currently be appear
+// in the CoreDNS configuration.
+func UpdateDnsConfig(t *testing.T, fromCtx context.Context, toCtx context.Context) {
```
kontaras, line 339: Nit: It is a bit confusing to have two functions with names that only differ by the capitalization of the first letter.

kontaras (approved):

test/testlib/multicluster_utilities.go
adriansuarez: How is this any more confusing than method/function overloading in languages like Java and C++, where functions can be named identically as long as they have different signatures?

If Go supported function overloading, I would use it.

test/testlib/multicluster_utilities.go
```diff
@@ -243,3 +253,97 @@ func AdjustPodTimeout(podName string, timeout time.Duration) time.Duration {
...
+	DNS_CONFIG_MARKER = "# BEGIN BASE CONFIG"
+	COREDNS_NS        = "kube-system"
+	COREDNS_CM        = "coredns"
+	COREFILE_KEY      = "Corefile"
+	DNS_SERVICE       = "dns-service"
+)
+
+func getClusterDnsServer(t *testing.T, clusterContext context.Context) string {
+	var ret string
+	DeployWithContext(t,
```
sivanov-nuodb, line 267: Nit: It's better to use `WithClusterDeployment` as this is just executing `kubectl`.

test/testlib/multicluster_utilities.go
```diff
@@ -243,3 +253,97 @@ func AdjustPodTimeout(podName string, timeout time.Duration) time.Duration {
...
+	require.NoError(t, err)
+}
+
+// UpdateDnsConfig adds the DNS server from one cluster as an upstream resolver
+// for the other, replacing any stale IP addresses that may currently be appear
+// in the CoreDNS configuration.
+func UpdateDnsConfig(t *testing.T, fromCtx context.Context, toCtx context.Context) {
+	dnsConfigSnippet := getDnsConfigSnippet(t, fromCtx)
+	if dnsConfigSnippet != "" {
+		DeployWithContext(t,
```
sivanov-nuodb, line 342: Nit: Same comment here for using `WithClusterDeployment` instead.

test/testlib/multicluster_utilities.go
```diff
@@ -243,3 +253,97 @@ func AdjustPodTimeout(podName string, timeout time.Duration) time.Duration {
...
+
+	// Update configmap if generated config changed
+	updatedConfig := dnsConfigSnippet + "\n" + config[idx:]
+	if updatedConfig == config {
+		t.Logf("DNS configuration up to date")
+		return
+	}
+	t.Logf("Adding DNS config snippet %s", dnsConfigSnippet)
+	cm.Data[COREFILE_KEY] = updatedConfig
+	_, err = clientset.CoreV1().ConfigMaps(COREDNS_NS).Update(ctx, cm, v1.UpdateOptions{})
```
sivanov-nuodb, line 332: Does this require a CoreDNS service restart? I remember that I needed to restart the pods after changing the `ttl` value in the same configuration.

sivanov-nuodb (approved):
Looks good.

It's fine to do that programatically every time for testing. What would be a good solution in production? Can we assign elastic IPs for CoreDNS? Maybe @acabrele knows more.

test/testlib/multicluster_utilities.go
adriansuarez: The difference is in the callback used, which does not take a `helm.Options`?

I do use the `helm.Options` to get the `k8s.KubectlOptions` that I invoke `kubectl`, but maybe there is another way to do it. Either way, I'd rather not go through another round of manual testing, so I think I'll leave it the way it is.

test/testlib/multicluster_utilities.go
adriansuarez: I see that the CoreDNS containers in the EKS have not undergone any restarts and they seem to dynamically pick up changes to the config.

```console
$ kubectl -n kube-system get pods
NAME                                 READY   STATUS    RESTARTS         AGE
...
coredns-7454b7b4d5-s5w9v             1/1     Running   0                21d
coredns-7454b7b4d5-xcfxn             1/1     Running   0                21d
...
```

It looks like CoreDNS reloads the config when it detects changes to the hash of the config.

```console
$ kubectl -n kube-system logs coredns-7454b7b4d5-xcfxn
...
[INFO] plugin/reload: Running configuration SHA512 = 5433e9ae056a43e2582ccea14671ad5dccbab0a40f9f731144af4f8a6b33066dd10f78d2d8926ebfadebd4c2b0800a583581386c3f97c21aa032d58c3e029507
[INFO] Reloading complete
```

sivanov-nuodb: Thanks for checking!

adriansuarez:
> Looks good.
> 
> It's fine to do that programatically every time for testing. What would be a good solution in production? Can we assign elastic IPs for CoreDNS? Maybe @acabrele knows more.

It would have to be an IP assigned to the load balancer, yes. I defer to @acabrele on whether that can work, but it sounds like it should.

PR #350, opened 2024-01-25
Bump versions of Helm charts

adriansuarez:


stable/admin/Chart.yaml
```diff
@@ -12,5 +12,5 @@ maintainers:
   email: support@nuodb.com
 engine: gotpl
 icon: https://raw.githubusercontent.com/nuodb/nuodb-helm-charts/master/images/nuodb.svg
-appVersion: 5.0.3
+appVersion: 5.0.4
```
sivanov-nuodb, line 15: I guess that we need to bump the pinned `tag` in the values.yaml so that to is accurate.

sivanov-nuodb (approved):
Thanks for doing that!

stable/admin/Chart.yaml
adriansuarez: Good point.

PR #352, opened 2024-02-29
add the archiveType to the readme

agoelzer:


adriansuarez (approved):
@agoelzer Looks good. Thanks for adding this so quickly.

sivanov-nuodb (approved):
Looks good.

Adding a validation should be easy as well. I assume that you are deferring it for another PR. Something like:

```
{{- define "database.archiveType" -}}
{{- $valid := list "lsa" "" nil }}
{{- if not (has .Values.database.archiveType $valid) }}
{{- fail (printf "Invalid archive type value: %s" .Values.database.archiveType) }}
{{- end }}
{{- end -}}
```

adriansuarez:
> Looks good.
> 
> Adding a validation should be easy as well. I assume that you are deferring it for another PR. Something like:
> 
> ```
> {{- define "database.archiveType" -}}
> {{- $valid := list "lsa" "" nil }}
> {{- if not (has .Values.database.archiveType $valid) }}
> {{- fail (printf "Invalid archive type value: %s" .Values.database.archiveType) }}
> {{- end }}
> {{- end -}}
> ```

Sounds good, but I don't think we need it for 3.8.0.

bkelly-ndb (approved):

PR #354, opened 2024-03-12
Make sure files are durable that should be in snapshot

adriansuarez:


sivanov-nuodb (approved):
Thanks for the quick turnaround!

kontaras (approved):

PR #358, opened 2024-03-18
Require root user for fsfreeze

adriansuarez:
This change sets `runAsUser=0` and `runAsGroup=0` on the backup-hooks sidecar container and suppresses `runAsNonRoot=true` on the pod security context if fsfreeze is required.

### DIFF
--- a/stable/database/templates/_helpers.tpl
+++ b/stable/database/templates/_helpers.tpl
@@ -172,15 +172,35 @@ Get security context runAsUser and runAsGroup
 {{- if eq (include "defaultfalse" .Values.database.securityContext.enabled) "true" }}
 runAsUser: {{ default 1000 .Values.database.securityContext.runAsUser }}
 runAsGroup: 0
-  {{- if and (or (eq (include "defaulttrue" .Values.database.initContainers.runInitDisk) "false") (eq (include "defaulttrue" .Values.database.initContainers.runInitDiskAsRoot) "false")) (ne (toString (default 1000 .Values.database.securityContext.runAsUser)) "0") }}
-runAsNonRoot: true
-  {{- end }}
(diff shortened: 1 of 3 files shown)


stable/database/templates/_helpers.tpl
```diff
@@ -172,15 +172,35 @@ Get security context runAsUser and runAsGroup
...
+{{- $runAsNonRoot := true -}}
+{{- if and
+       (eq (include "defaulttrue" .Values.database.initContainers.runInitDisk) "true")
+       (eq (include "defaulttrue" .Values.database.initContainers.runInitDiskAsRoot) "true") }}
+  {{- $runAsNonRoot = false -}}
+{{- else if .Values.database.backupHooks -}}
+  {{- if and
+         (eq (include "defaultfalse" .Values.database.backupHooks.enabled) "true")
+         (eq (include "defaultfalse" .Values.database.backupHooks.useSuspend) "false")
+         (eq (include "defaultfalse" .Values.database.sm.noHotCopy.journalPath.enabled) "true") -}}
```
sivanov-nuodb, line 198: Technically this is not needed for the TE deployments as they don't have backup hooks container but I guess it will be too much of a hassle to conditionalize the helper with _context_.

sivanov-nuodb (approved):
Looks good.

stable/database/templates/_helpers.tpl
adriansuarez: True. I did not think the extra complexity was warranted just to decouple the security contexts from each other. If there's a policy that requires `runAsNonRoot=true`, we cannot be in compliance with it anyway if fsfreeze is required, so I'm not sure what the value is making the TE Deployment compliant by itself.

PR #359, opened 2024-03-18
Add create-repo.sh script to create in-cluster Helm repo

adriansuarez:
This change adds a script that creates a Helm repo that is usable within the local Kubernetes cluster that contains all published versions along with the development version packaged from the current commit.


adriansuarez:
@sivanov-nuodb This is just copying what is in `operator/config/repo` in the nuodb-control-plane repo and running it as a script instead of a make target.

scripts/repo/charts/index.html
```diff
@@ -0,0 +1,23 @@
+<!DOCTYPE html>
```
kontaras, line 1: Isn't this file getting overriden in `create-repo.sh`?
`curl -s https://nuodb.github.io/nuodb-helm-charts/index.yaml > repo/charts/index.yaml`

So, it should be in `.gitignore`.

kontaras (approved):

scripts/repo/charts/index.html
adriansuarez: I'm going to adjust the readiness probe so that this file is not required.

sivanov-nuodb (approved):
Thanks for adding this into NuoDB Helm charts repo!

PR #360, opened 2024-03-20
[HELM-422] validate helm `archiveType` value

agoelzer:
prevent deployment if invalid archiveType value is specified

### DIFF
--- a/stable/database/values.schema.json
+++ b/stable/database/values.schema.json
@@ -1,0 +1,21 @@
+{
+    "$schema": "http://json-schema.org/schema#",
+    "type": "object",
+    "required": [
+        "database"
+    ],
+    "properties": {
+        "database": {
+            "type": "object",
+            "required": [
+                "archiveType"
+            ],
+            "properties": {
+                "archiveType": {
+                    "type": "string",
+                    "pattern": "^(|lsa)$"
+                }
+            }
+        }
+    }
+}
--- a/stable/database/values.schema.json
+++ b/stable/database/values.schema.json
@@ -13,7 +13,10 @@
             "properties": {
                 "archiveType": {
                     "type": "string",
-                    "pattern": "^(|lsa)$"
+                    "enum": [
+                        "",
+                        "lsa"
+                    ]
                 }
             }
         }
--- a/test/integration/template_database_test.go
+++ b/test/integration/template_database_test.go
@@ -251,6 +251,60 @@ func TestDatabaseNodePortServiceRenders(t *testing.T) {
 	}
 }
 
+func TestDatabaseInvalidArchiveType(t *testing.T) {
+	// Path to the helm chart we will test
+	helmChartPath := testlib.DATABASE_HELM_CHART_PATH
+
+	options := &helm.Options{
+		SetValues: map[string]string{
+			"cloud.provider":                        "amazon",
+			"database.te.externalAccess.enabled":    "true",
+			"database.te.externalAccess.type":       "NodePort",
+			"database.te.externalAccess.internalIP": "true",
+			"database.archiveType": "invalid",
+		},
+	}
+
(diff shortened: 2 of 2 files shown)


stable/database/values.schema.json
```diff
@@ -0,0 +1,21 @@
...
+    "properties": {
+        "database": {
+            "type": "object",
+            "required": [
+                "archiveType"
+            ],
+            "properties": {
+                "archiveType": {
+                    "type": "string",
+                    "pattern": "^(|lsa)$"
```
adriansuarez, line 16: Is it possible to use [`enum`](https://json-schema.org/understanding-json-schema/reference/enum) instead?

That might generate a better error message and seems more consistent with how we are handling the property.

stable/database/values.schema.json
adriansuarez, line 1: I didn't know about this Helm feature. Do you know what version introduced it?

adriansuarez (approved):
Looks good. Good find with the JSON Schema support in Helm.

stable/database/values.schema.json
agoelzer: it's supported since Helm 3.x (not supported in 2.x)

test/integration/template_database_test.go
```diff
@@ -251,6 +251,60 @@ func TestDatabaseNodePortServiceRenders(t *testing.T) {
...
+			"cloud.provider":                        "amazon",
+			"database.te.externalAccess.enabled":    "true",
+			"database.te.externalAccess.type":       "NodePort",
+			"database.te.externalAccess.internalIP": "true",
+			"database.archiveType": "invalid",
+		},
+	}
+
+	_, err := helm.RenderTemplateE(t, options, helmChartPath, "release-name", []string{"templates/service.yaml"})
+	assert.Contains(t, err.Error(), "database.archiveType: Does not match pattern")
```
sivanov-nuodb, line 269: Can you post the full error? Does Helm print the pattern etc.?

sivanov-nuodb (approved):
Looks good.

stable/database/values.schema.json
agoelzer: good point - tested and it works! Adjusted code.

stable/database/values.schema.json
agoelzer: yes, the default "values.yaml" file is used and when overwriting with "--set-string" or similar, those get validated as well.

test/integration/template_database_test.go
```diff
@@ -251,6 +251,60 @@ func TestDatabaseNodePortServiceRenders(t *testing.T) {
...
+func TestDatabaseInvalidArchiveType(t *testing.T) {
+	// Path to the helm chart we will test
+	helmChartPath := testlib.DATABASE_HELM_CHART_PATH
+
+	options := &helm.Options{
+		SetValues: map[string]string{
+			"cloud.provider":                        "amazon",
+			"database.te.externalAccess.enabled":    "true",
+			"database.te.externalAccess.type":       "NodePort",
+			"database.te.externalAccess.internalIP": "true",
```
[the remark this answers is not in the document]
agoelzer: not sure why they are needed, but when I remove them the test case fails with `Error: could not find template templates/service.yaml in chart`

test/integration/template_database_test.go
agoelzer: with the code change to an enum, the new error check is `database.archiveType must be one of the following:` - I didn't include the allowed values in the check to prevent a test failure if new values are added in the future.

test/integration/template_database_test.go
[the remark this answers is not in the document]
adriansuarez: That's because you're resolving `templates/service.yaml` in the next line. You're basically just trying to render the templates and actually have no interest in the service. It would more sense to supply `templates/statefulset.yaml`, since that will always be present and is where the `archiveType` value is used (even though you're only checking that an error occurred and are not interested in the content).

PR #361, opened 2024-03-21
Account for gap between snapshot preparation and archive creation

adriansuarez:
For SMs created from snapshots, we currently we rely on the absence of the info.json file and presence of the backup.txt file to detect that the archive has to be prepared from a snapshot. But there is a window of time until the info.json file for the newly-prepared archive is created in which any failure would leave the `nuosm` script unable to detect if it should perform a normal restart.

This change adds a restore.txt file which is created at the same time the old info.json and backup.txt file are deleted and is used to detect that the SM already performed snapshot preparation and should continue with normal startup.

### DIFF
--- a/stable/database/files/nuosm
+++ b/stable/database/files/nuosm
@@ -402,7 +402,7 @@ function checkBackupId() {
 function loadFromSnapshot() {
   local recreate_archives="false"
 
-  if [ ! -f "$DB_DIR/info.json" ]; then
+  if [ ! -f "$DB_DIR/info.json" ] && [ ! -f "$DB_DIR/restored.txt" ]; then
     local archives="$(find /var/opt/nuodb/archive -name info.json)"
     if [ -z "$archives" ] || [ "$(echo "$archives" | wc -l)" != 1 ]; then
       # Relax check for archive snapshot for SMs other than ordinal 0. It is
@@ -467,9 +467,15 @@ function loadFromSnapshot() {
   fi
 
   if [ "$recreate_archives" == "true" ]; then
+    # Create restored.txt to signal that snapshot preparation is complete. This
+    # is needed in the absence of info.json, which is not created for the
+    # restored archive object until later.
+    echo "$BACKUP_ID" > "${DB_DIR}/restored.txt"
+
     log "Removing metadata from snapshot archive"
-    rm "${DB_DIR}/info.json" "${DB_DIR}/backup.txt"
-    [ -e "${JOURNAL_DIR}/backup.txt" ] && rm "${JOURNAL_DIR}/backup.txt"
+    rm -f "${DB_DIR}/info.json"
+    rm -f "${DB_DIR}/backup.txt"
+    rm -f "${JOURNAL_DIR}/backup.txt"
   fi
 }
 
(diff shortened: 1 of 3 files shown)


stable/database/files/nuosm
```diff
@@ -467,9 +467,15 @@ function loadFromSnapshot() {
   fi
 
   if [ "$recreate_archives" == "true" ]; then
+    # Create restored.txt to signal that snapshot preparation is complete. This
+    # is needed in the absence of info.json, which is not created for the
+    # restored archive object until later.
+    echo "$BACKUP_ID" > "${DB_DIR}/restored.txt"
```
kontaras, line 473: Micro-optimization: why not `mv "${DB_DIR}/info.json" "${DB_DIR}/restored.txt"`? It closes the (extremely unlikely) window if the SM crashes between those two commands and we are left with an Archive that is not registered with the admin and does not want to try.

kontaras (commented):
Don't forget to delete `restored.txt` at the end. It will break an SM being restored, snapshot, and restored.

adriansuarez:
> Don't forget to delete `restored.txt` at the end. It will break an SM being restored, snapshot, and restored.

Great point. The `info.json` file is created by `nuodocker start sm`, so we can delete `restored.txt` after `info.json` is created, because we have handed control of the startup lifecycle to `nuodocker start sm`. Since `nuodocker start sm` is implemented in the container image, that would create a dependency on a particular image version for backup to work.

I will put the deletion inside the `backup_hooks.py` script in the `pre-backup` hook, before the new `backup.txt` is created.

stable/database/files/nuosm
adriansuarez: What is the window that you're trying to close? The presence of `backup.txt` should not inhibit creation of the new archive object.

kontaras: Never mind, I missed a `else` branch.

sivanov-nuodb:
> > Don't forget to delete `restored.txt` at the end. It will break an SM being restored, snapshot, and restored.
> 
> Great point. The `info.json` file is created by `nuodocker start sm`, so the `nuosm` script cannot delete `restored.txt` after `info.json` is created, because we have handed control of the startup lifecycle to `nuodocker start sm` by that point. Since `nuodocker start sm` is implemented in the container image, putting that behavior there would create a dependency on a particular image version for backup to work.
> 
> I will put the deletion inside the `backup_hooks.py` script in the `pre-backup` hook, before the new `backup.txt` is created.

What is the issue by not deleting the `restore.txt` file? I see that only the absence of the file is checked in combination with the absence of `info.json`.

adriansuarez:
> What is the issue by not deleting the `restore.txt` file? I see that only the absence of the file is checked in combination with the absence of `info.json`.

After thinking a little bit more about it, I actually do not think it is an issue. A stale `restore.txt` can only be found in the case where there is a stale `info.json`, i.e. restore with same domain and database. The presence of the `backup.txt` would still prevent us from skipping the deletion of these files.

Either way, I think it makes sense to remove the stale `restore.txt`, since it is irrelevant which backup the backup came from.

kontaras (approved):

PR #363, opened 2024-03-22
Tighten semver parsing for nuosql version

adriansuarez:
This change tightens the parsing of the nuosql version to account for any non-numerical suffix that is not delimited by a dash (`-`). For example, development versions on release branches have the form `<major>.<minor>.<patch>.<comment>-<build num>-<SHA>-<arch>`. The `<comment>` part of the version was not being stripped away and screwed up the semver parsing.

### DIFF
--- a/test/minikube/minikube_external_access_test.go
+++ b/test/minikube/minikube_external_access_test.go
@@ -77,7 +77,7 @@ func getNuoSQLVersion(t *testing.T) *semver.Version {
 	require.NotNil(t, match, string(out))
 	versionStr := normalizeNuoSQLVersion(match[1])
 	version, err := semver.NewVersion(versionStr)
-	require.NoError(t, err, version)
+	require.NoError(t, err, versionStr)
 	return version
 }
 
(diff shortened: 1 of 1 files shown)


test/minikube/minikube_external_access_test.go
```diff
@@ -74,14 +74,31 @@ func getNuoSQLVersion(t *testing.T) *semver.Version {
 		require.NoError(t, err)
 	}
 	match := regexp.MustCompile("NuoDB Client build (.*)").FindStringSubmatch(string(out))
-	require.NotNil(t, match, out)
-	// strip comment from semantic version
-	versionStr := strings.Split(match[1], "-")[0]
+	require.NotNil(t, match, string(out))
```
sivanov-nuodb, line 77: Nit: The check could be moved before the `FindStringSubmatch()` call.

test/minikube/minikube_external_access_test.go
```diff
@@ -74,14 +74,31 @@ func getNuoSQLVersion(t *testing.T) *semver.Version {
...
+	versionStr := normalizeNuoSQLVersion(match[1])
 	version, err := semver.NewVersion(versionStr)
-	require.NoError(t, err)
+	require.NoError(t, err, version)
```
sivanov-nuodb, line 80: The `version` variable will probably be `nil`. The `versionStr` should be printed instead.

sivanov-nuodb (approved):
Looks good.

PR #365, opened 2024-06-18
Support for HotSnap in backup hooks

sivanov-nuodb:
**Changes**

```
This change allows using `hotsnap` (available in NuoDB 6.0.2 and above) to
pause/resume SM archiving in the backup hooks container. A new Helm value
`database.backupHooks.freezeMode` is added so that the user can select the mode
in which the backup hook service operates.

Added `database.backupHooks.timeout` (30s by default) which defines the timeout
in seconds after which the backup is automatically canceled if the post-backup
hook is not invoked. This is generally useful since it guards against
situations where the post-backup hook is not called to unfreeze the archive
volume. Automatic backup canceling is mandatory when "hotsnap" mode is
activated. The SM automatically resumes the archiving and the backup hook
server must remove the backup metadata file.

When "hotsnap" mode is used, an in-progress backup is invalidated on SM
restart. The backup hook server stores the SM start ID and the backup ID when
the pre-backup hook is called. This allows it to detect SM restart and remove
the backup metadata in such cases.

Deprecated the `database.backupHooks.useSuspend` Helm value which is replaced
by `database.backupHooks.freezeMode=suspend`.
```


sivanov-nuodb:
The `TestKubernetesIngress` test failure is fixed by https://github.com/nuodb/nuodb-helm-charts/pull/366.

stable/database/files/backup_hooks.py
```diff
@@ -153,22 +182,40 @@ def freeze_archive(unfreeze=False):
         subprocess.check_output(
             ["fsfreeze", action, ARCHIVE_DIR], stderr=subprocess.STDOUT
         )
-
+    else:
+        raise RuntimeError("not supported quiesce mode '{}'".format(FREEZE_MODE))
```
adriansuarez, line 186: Nit: Can you change this to "Unsupported quiesce mode..."?

stable/database/files/backup_hooks.py
```diff
@@ -142,7 +149,29 @@ def freeze_archive(unfreeze=False):
                 os.kill(int(pid), signal.SIGSTOP)
                 # Make sure all threads are suspended
                 await_suspended(pid)
-    else:
+    elif FREEZE_MODE == "hotsnap":
+        # Freeze or unfreeze the archive using hotsnap
+        sid = processes[0]["sid"]
+        extra_args = []
+        if unfreeze:
+            LOGGER.info("Resuming archiving for nuodb process with startId %s", sid)
```
adriansuarez, line 157: Nit: Can you write the start ID as "start ID"? There is another error or log message where it appears as "start ID".

stable/database/files/backup_hooks.py
```diff
@@ -153,22 +182,40 @@ def freeze_archive(unfreeze=False):
...
     # already in progress, which may block indefinitely if writes to the
     # archive and journal directories are frozen
     if os.path.exists(ARCHIVE_BACKUP_ID_FILE):
-        raise UserError(
-            "Backup ID file {} exists. Execute post-backup hook to complete current backup.".format(
-                ARCHIVE_BACKUP_ID_FILE
+        # Check if the SM process was restarted which will invalidate the
+        # previous pre-backup operation using HotSnap.
+        stored_start_id = get_start_id()
+        if FREEZE_MODE == "hotsnap" and stored_start_id and processes[0]["sid"] != stored_start_id:
```
adriansuarez, line 205: What is special about freeze mode `hotsnap` when it comes to handling the presence of the backup file?

stable/database/files/backup_hooks.py
```diff
@@ -376,6 +473,15 @@ def start_server(port):
         LOGGER.info("Starting backup hooks server on port %s", port)
         httpd.serve_forever()
 
+def verify_prerequisites():
+    if FREEZE_MODE == "hotsnap":
+        if which("nuocmd") is None:
+            raise RuntimeError("'nuocmd' command not found")
+        elif subprocess.call(["nuocmd", "pause", "archiving", "-h"], 
+                           stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL) == 2:
```
adriansuarez, line 481: What if it returns 1? I think we should only check that the exit code equals 2 before saying that the command is not supported, otherwise we should propagate whatever error appeared on stdout/stderr.

stable/database/templates/_helpers.tpl
```diff
@@ -84,6 +87,22 @@ Also, we can't use a single if because lazy evaluation is not an option
...
+
+{{/*
+Renders the freeze mode for backup hooks.
+*/}}
+{{- define "backupHooks.freezeMode" -}}
+{{- if eq .Values.database.backupHooks.freezeMode "" -}}
+  {{- if (eq (include "defaultfalse" .Values.database.backupHooks.useSuspend) "true") -}}
+suspend
+  {{- else -}}
+hotsnap
```
adriansuarez, line 100: How is this in terms of backwards compatibility? If someone upgrades from 3.8.0 to 3.9.0 who was using fsfreeze with some nuodb version that does not support hotsnap, what happens? Do they have to explicitly configure `freezeMode=fsfreeze`?

stable/database/files/backup_hooks.py
sivanov-nuodb: Good catch!

stable/database/files/backup_hooks.py
sivanov-nuodb: Will do.

stable/database/templates/_helpers.tpl
sivanov-nuodb: > Do they have to explicitly configure freezeMode=fsfreeze?
Yes, they do. Otherwise, the backup hook container will fail saying that `hotsnap` is not supported. 

It is a change in behaviour indeed but I can't think of a good way to enable `hotsnap` by default without unreliably parsing the image tag to determine if hotsnap is actually supported by the selected image.

stable/database/files/backup_hooks.py
sivanov-nuodb: I have an E2E test for that [here](https://github.com/nuodb/nuodb-helm-charts/pull/365/files#diff-2a9c2e48a29de00e39dd2e1784e4ababaaec6e085b7f0c91a79a93c10a009966R338-R370) but it doesn't fully exercise your scenario since it doesn't enable _restore from snapshot_.

I think in your scenario we just log. See https://github.com/nuodb/nuodb-helm-charts/blob/master/stable/database/files/nuosm#L460.

kontaras: But why are we cleaning up the backup files and trying the backup for `hotsnap` while erroring out for other backup modes?

sivanov-nuodb: SM restart invalidates the backup ID when using `hotsnap`, which is not the case when using `fsfreeze`.

test/minikube/minikube_fsfreeze_test.go
```diff
@@ -241,3 +249,123 @@ func TestFsFreezeBackupHook(t *testing.T) {
...
+			"database.sm.noHotCopy.journalPath.persistence.storageClass": testlib.SNAPSHOTABLE_STORAGE_CLASS,
+			"database.sm.noHotCopy.journalPath.enabled":                  "true",
+			"database.sm.noHotCopy.replicas":                             "1",
+			"database.sm.hotCopy.replicas":                               "0",
+			"database.backupHooks.enabled":                               "true",
+			"database.backupHooks.freezeMode":                            "hotsnap",
+		},
+	})
+
+	// Invoke backup hooks on SM pod
```
kontaras, line 283: Nit: Should this comment be further down in the file?

test/minikube/minikube_fsfreeze_test.go
```diff
@@ -241,3 +249,123 @@ func TestFsFreezeBackupHook(t *testing.T) {
...
+		testlib.InvokeBackupHook(t, namespaceName, smPod, "pre-backup/"+backupId, "timeout", timeout)
+
+		// Wait for the automatic backup canceling and metadata files removal
+		testlib.Await(t, func() bool {
+			return testlib.GetStringOccurrenceInLog(t, namespaceName, smPod,
+				fmt.Sprintf("Canceling backup with ID %s. Timeout after %ss", backupId, timeout),
+				&corev1.PodLogOptions{Container: "backup-hooks"}) > 0
+		}, 10*time.Second)
+		k8s.RunKubectl(t, kubectlOptions, "exec", smPod, "-c", "engine", "--",
+			"sh", "-c", `while [ -f "/var/opt/nuodb/archive/backup.txt" ]; do sleep 1; done`)
```
kontaras, line 331: Nit: Why not add a timeout to this loop so, in case of failure, the test can exit cleanly instead of getting killed?

test/testlib/nuodb_database_utilities.go
```diff
@@ -820,13 +820,27 @@ type BackupHookResponse struct {
...
+func GetBackupHookResponse(t *testing.T, namespaceName, smPod, resourcePath string, payload ...string) *BackupHookResponse {
+	cmdArgs := []string{
+		"exec", smPod, "-c", "engine", "--",
+		"curl", "-s", "-X", "POST", "http://0.0.0.0:8000/" + resourcePath,
+	}
+	// Prepare the payload as map
+	if len(payload) > 0 {
+		require.Equal(t, 0, len(payload)%2, "Backup hook payload must contain even number of entries")
+		payloadMap := make(map[string]string, len(payload)/2)
+		for i := 0; i < len(payload)-1; i++ {
```
kontaras, line 832: Shouldn't `i` get incremented by 2?

kontaras (approved):

stable/database/files/backup_hooks.py
adriansuarez: So basically, we want to avoid deleting the backup files if freeze mode is `fsfreeze`, since that would block things completely. I think we can safely apply this logic for the `suspend` freeze mode.

test/minikube/minikube_fsfreeze_test.go
```diff
@@ -241,3 +249,123 @@ func TestFsFreezeBackupHook(t *testing.T) {
...
+	helmChartReleaseName, namespaceName := testlib.StartAdmin(t, &helm.Options{}, 1, "")
+
+	// Create database release with CSI driver storage class and backup hooks enabled
+	admin := fmt.Sprintf("%s-nuodb-cluster0", helmChartReleaseName)
+	admin0 := fmt.Sprintf("%s-0", admin)
+	defer testlib.Teardown(testlib.TEARDOWN_DATABASE)
+	sourceDatabaseChartName := testlib.StartDatabase(t, namespaceName, admin0, &helm.Options{
+		SetValues: map[string]string{
+			// HotSnap is available in NuoDB version 6.0.2 and above
+			"nuodb.image.tag":                                            "6.0.2",
```
adriansuarez, line 268: How does this work with the image inject in Bamboo? I guess master and rel/6.0-dev and up will have supported product versions, but not rel/5.1-dev, etc.

test/testlib/nuodb_database_utilities.go
```diff
@@ -820,13 +820,27 @@ type BackupHookResponse struct {
 	Message string `json:"message"`
 }
 
-func GetBackupHookResponse(t *testing.T, namespaceName, smPod, resourcePath string) *BackupHookResponse {
+func GetBackupHookResponse(t *testing.T, namespaceName, smPod, resourcePath string, payload ...string) *BackupHookResponse {
```
adriansuarez, line 823: Why not pass a map that gets serialized to JSON?

adriansuarez (approved):
Looks good. I left a couple more comments.

test/testlib/nuodb_database_utilities.go
sivanov-nuodb: I didn't want to change all the callers of this function. This is why I used a variadic argument.

test/minikube/minikube_fsfreeze_test.go
sivanov-nuodb: Good point! I forgot about Bamboo. I will skip the test based on version constraints.

stable/database/files/backup_hooks.py
```diff
@@ -126,14 +135,12 @@ def await_suspended(pid, interval=0.25, retries=16):
     )
 
 
-def freeze_archive(unfreeze=False):
-    if USE_SUSPEND:
+def freeze_archive(backup_id, processes, unfreeze=False, timeout=None):
+    if FREEZE_MODE == "suspend":
```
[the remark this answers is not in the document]
sivanov-nuodb: Will do.

test/testlib/nuodb_database_utilities.go
sivanov-nuodb: If this will make it read better, I can change it.

test/integration/template_database_test.go
```diff
@@ -2681,11 +2684,80 @@ func TestDatabaseStatefulSetBackupHooksSidecar(t *testing.T) {
...
+			volumes[i] = v.Name
+		}
+		assert.Contains(t, volumes, "archive-volume")
+		assert.Contains(t, volumes, "backup-hooks")
+		assert.NotContains(t, volumes, "journal-volume")
+		assert.Contains(t, volumes, "eph-volume")
+		assert.Contains(t, volumes, "tls")
+
+		// Check that nuodb/nuodb-ce container image is used
+		assert.Contains(t, sidecar.Image, "docker.io/nuodb/nuodb-ce")
```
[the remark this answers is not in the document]
sivanov-nuodb: Will do.

stable/database/files/backup_hooks.py
sivanov-nuodb: Right. I can make the following change:

```diff
@@ -202,7 +206,7 @@ def pre_backup(backup_id, payload):
         # Check if the SM process was restarted which will invalidate the
         # previous pre-backup operation using HotSnap.
         stored_start_id = get_start_id()
-        if FREEZE_MODE == "hotsnap" and stored_start_id and processes[0]["sid"] != stored_start_id:
+        if FREEZE_MODE != MODE_FSFREEZE and stored_start_id and processes[0]["sid"] != stored_start_id:
             # Remote the backup files from the previous pre-backup operation
             LOGGER.warning("Unexpected start ID: current=%s, stored=%s. " +
                             "SM process restarted while executing backup ID %s",
```

test/testlib/nuodb_database_utilities.go
adriansuarez: Okay, thanks for the explanation.

test/minikube/minikube_fsfreeze_test.go
sivanov-nuodb: Will do.

PR #366, opened 2024-06-19
Fix HAProxy deployment readiness check

sivanov-nuodb:
**Changes**

```
HAProxy Helm chart deploys Kubernetes job with the same as the controller pod
which messes up our readiness check. Filter out all pods started from
Kubernetes jobs when searching for HAProxy pod.
```


adriansuarez (approved):
Looks good.

PR #367, opened 2024-07-24
Enable and test NuoDocker topology discovery

kontaras:
I am adding new logic to NuoDocker to add admin labels, process labels, and admin affinity based on topology labels on the kubernetes node. This functionality requires new permissions for the Service Account (gated behind `nuodb.addClusterRoleBinding` chart value). Also, this functionality cannot be easily tested in existing NuoDocker test suite, so add an integration test to this repo.

### DIFF
--- a/stable/admin/README.md
+++ b/stable/admin/README.md
@@ -108,6 +108,7 @@ The following tables list the configurable parameters for the `nuodb` option:
 | `serviceAccount` | The name of the service account used by NuoDB Pods | `nuodb` |
 | `addServiceAccount` | Whether to create a new service account for NuoDB containers | `true` |
 | `addRoleBinding` | Whether to add role and role-binding giving `serviceAccount` access to Kubernetes APIs (Pods, PersistentVolumes, PersistentVolumeClaims, StatefulSets) | `true` |
+| `addClusterRoleBinding` | If `addRoleBinding` is enabled, add addition cluster wide role and role-binding giving `serviceAccount` access to Cluster wide Kubernetes APIs (read access to Nodes) | `true` |
 
 The `registry` option can be used to connect to private image repositories, such as Artifactory.
 
--- a/stable/admin/README.md
+++ b/stable/admin/README.md
@@ -108,7 +108,7 @@ The following tables list the configurable parameters for the `nuodb` option:
 | `serviceAccount` | The name of the service account used by NuoDB Pods | `nuodb` |
 | `addServiceAccount` | Whether to create a new service account for NuoDB containers | `true` |
 | `addRoleBinding` | Whether to add role and role-binding giving `serviceAccount` access to Kubernetes APIs (Pods, PersistentVolumes, PersistentVolumeClaims, StatefulSets) | `true` |
-| `addClusterRoleBinding` | If `addRoleBinding` is enabled, add addition cluster wide role and role-binding giving `serviceAccount` access to Cluster wide Kubernetes APIs (read access to Nodes) | `true` |
+| `addClusterRoleBinding` | If `addRoleBinding` is enabled, add addition cluster role and cluster role-binding giving `serviceAccount` access to Cluster wide Kubernetes APIs (read access to Nodes) | `true` |
 
 The `registry` option can be used to connect to private image repositories, such as Artifactory.
 
--- a/stable/admin/README.md
+++ b/stable/admin/README.md
@@ -108,7 +108,7 @@ The following tables list the configurable parameters for the `nuodb` option:
 | `serviceAccount` | The name of the service account used by NuoDB Pods | `nuodb` |
 | `addServiceAccount` | Whether to create a new service account for NuoDB containers | `true` |
 | `addRoleBinding` | Whether to add role and role-binding giving `serviceAccount` access to Kubernetes APIs (Pods, PersistentVolumes, PersistentVolumeClaims, StatefulSets) | `true` |
-| `addClusterRoleBinding` | If `addRoleBinding` is enabled, add addition cluster role and cluster role-binding giving `serviceAccount` access to Cluster wide Kubernetes APIs (read access to Nodes) | `true` |
+| `addClusterRoleBinding` | If `addRoleBinding` is enabled, also add cluster role and cluster role-binding giving `serviceAccount` access to Cluster wide Kubernetes APIs (read access to Nodes) | `true` |
 
 The `registry` option can be used to connect to private image repositories, such as Artifactory.
 
--- a/test/minikube/minikube_kaa_additions_test.go
+++ b/test/minikube/minikube_kaa_additions_test.go
@@ -129,7 +129,6 @@ func TestKubernetesTopologyDiscover(t *testing.T) {
 
 	currentRegions := testlib.LabelNodesIfMissing(t, "topology.kubernetes.io/region", "test-region")
 	currentZones := testlib.LabelNodesIfMissing(t, "topology.kubernetes.io/zone", "test-zone")
-	currentNodes := testlib.GetNodeNames(t)
 
 	defer testlib.Teardown(testlib.TEARDOWN_ADMIN)
 
@@ -140,11 +139,13 @@ func TestKubernetesTopologyDiscover(t *testing.T) {
 	admin := fmt.Sprintf("%s-nuodb-cluster0", helmChartReleaseName)
 	admin0 := fmt.Sprintf("%s-0", admin)
 
-	testlib.VerifyLabels(t, namespaceName, admin0,
-		map[string][]string{
-			"node":   currentNodes,
-			"zone":   currentZones,
-			"region": currentRegions,
+	adminPod := testlib.GetPod(t, namespaceName, admin0)
+	adminNode := adminPod.Spec.NodeName
+	testlib.VerifyAdminLabels(t, namespaceName, admin0,
+		map[string]string{
+			"node":   adminNode,
+			"zone":   currentZones[adminNode],
+			"region": currentRegions[adminNode],
 		})
 
 	dbName := "db"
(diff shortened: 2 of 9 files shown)


test/minikube/minikube_kaa_additions_test.go
```diff
@@ -118,4 +120,52 @@ func TestKaaRolebindingDisabled(t *testing.T) {
 	require.True(t, len(config.StatefulSets) == 0)
 	require.True(t, len(config.Volumes) == 0)
 	require.True(t, len(config.Pods) == 0)
-}
\ No newline at end of file
+}
+
+func TestKubernetesTopologyDiscover(t *testing.T) {
+	testlib.SkipTestOnNuoDBVersionCondition(t, "< 6.0.3")
```
kontaras, line 126: I plan to cross-port the nuodocker changes to 5.1.3, 6.0.3, and 7.0.

stable/admin/README.md
```diff
@@ -108,6 +108,7 @@ The following tables list the configurable parameters for the `nuodb` option:
 | `serviceAccount` | The name of the service account used by NuoDB Pods | `nuodb` |
 | `addServiceAccount` | Whether to create a new service account for NuoDB containers | `true` |
 | `addRoleBinding` | Whether to add role and role-binding giving `serviceAccount` access to Kubernetes APIs (Pods, PersistentVolumes, PersistentVolumeClaims, StatefulSets) | `true` |
+| `addClusterRoleBinding` | If `addRoleBinding` is enabled, add addition cluster wide role and role-binding giving `serviceAccount` access to Cluster wide Kubernetes APIs (read access to Nodes) | `true` |
```
adriansuarez, line 111: Can you rephrase this as "add cluster role and cluster role-binding"? ClusterRole/ClusterRoleBinding are different resource kinds from Role/RoleBinding, but the way it is worded is confusing.

test/testlib/nuodb_admin_utilities.go
```diff
@@ -303,3 +303,22 @@ func AwaitServerState(t *testing.T, namespace string, adminPod string,
 		return false
 	}, timeout)
 }
+
+// VerifyLables checks if adminPod in namespace has label and its value is one of the expectedValues.
+func VerifyLables(t *testing.T, namespace string, adminPod string, label string, expectedValues []string) {
```
adriansuarez, line 308: NIt: VerifyLables -> VerifyLabels

test/testlib/nuodb_admin_utilities.go
```diff
@@ -303,3 +303,22 @@ func AwaitServerState(t *testing.T, namespace string, adminPod string,
...
+	fieldName := "labels." + label
+	output, err := k8s.RunKubectlAndGetOutputE(t, k8s.NewKubectlOptions("", "", namespace), "exec", adminPod, "-c", "admin", "--",
+		"nuocmd", "--show-json-fields", fieldName, "get", "servers")
+	require.NoError(t, err, output)
+
+	lines := strings.Split(output, "\n")
+	require.Equal(t, 3, len(lines))
```
adriansuarez, line 315: Can you add a comment that shows what the expected output looks like?

test/minikube/minikube_kaa_additions_test.go
```diff
@@ -118,4 +120,52 @@ func TestKaaRolebindingDisabled(t *testing.T) {
...
+	options := helm.Options{}
+	helmChartReleaseName, namespaceName := testlib.StartAdmin(t, &options, 1, "")
+	kubectlOptions := k8s.NewKubectlOptions("", "", namespaceName)
+	options.KubectlOptions = kubectlOptions
+	admin := fmt.Sprintf("%s-nuodb-cluster0", helmChartReleaseName)
+	admin0 := fmt.Sprintf("%s-0", admin)
+
+	testlib.VerifyLables(t, namespaceName, admin0, "hostname", currentNodes)
+	testlib.VerifyLables(t, namespaceName, admin0, "zone", currentZones)
+	testlib.VerifyLables(t, namespaceName, admin0, "region", currentRegions)
```
adriansuarez, line 145: We don't need three invocations of this. You can change the function to take a map of expected values and perform JSON deserialization of the `nuocmd get servers` output.

test/testlib/nuodb_admin_utilities.go
adriansuarez, line 308: We always know exactly what value the label is supposed to have, so why do we take an array?

test/testlib/nuodb_admin_utilities.go
```diff
@@ -303,3 +303,22 @@ func AwaitServerState(t *testing.T, namespace string, adminPod string,
...
+
+	var actualField string
+	var actualValue string
+	read, err := fmt.Sscan(lines[1], &actualField, &actualValue)
```
adriansuarez, line 319: This would be easier to read if we do JSON deserialization. There is already a function that deserializes the entire server object:
https://github.com/nuodb/nuodb-helm-charts/blob/master/test/testlib/nuodb_admin_utilities.go#L259

If you add a `Labels` field to the [`NuoDBServer` struct](https://github.com/nuodb/nuodb-helm-charts/blob/master/test/testlib/NuoDBServer.go#L29), then it would be easy to check for the presence of several labels.

test/testlib/nuodb_admin_utilities.go
kontaras: We will have multiple values for node (and possibly other labels) if the test is run on a multi-node cluster.

adriansuarez: Even if we ran these tests on a multi-host cluster like Kind, at the point where this is invoked in the test, the AP has already been scheduled on a node, so we would know exactly what labels it's supposed to have since we can get them from the node.

adriansuarez:
Should these changes also include a way to specify `--admin-affinity-label-keys`?
We would not want to enable that by default, since it has an image version requirement, but the user should be able to supply an explicit value.

kontaras:
> Should these changes also include a way to specify `--admin-affinity-label-keys`? We would not want to enable that by default, since it has an image version requirement, but the user should be able to supply an explicit value.

Yes, but if we do, we should also have a way to specify admin labels. I will do that as a separate commit to this PR.

test/testlib/nuodb_admin_utilities.go
kontaras: So we would need to get the node name from the AP and then look up the labels for that Node. It is more work, but I suppose that it does check that the AP is not picking the labels from the wrong node...

test/testlib/minikube_utilities.go
```diff
@@ -1287,6 +1287,30 @@ func LabelNodes(t *testing.T, namespaceName string, labelName string, labelValue
 	}
 }
 
+func LabelNodesIfMissing(t *testing.T, labelName string, labelValue string) []string {
```
sivanov-nuodb, line 1290: Could you please include a docstring for this function? I'm confused how the return value is consumed. It looks like the current values for different nodes are retuned, but how does the consumer know which one is from which node? Maybe you should return a `map` name to the current label value?

test/testlib/nuodb_admin_utilities.go
```diff
@@ -303,3 +303,15 @@ func AwaitServerState(t *testing.T, namespace string, adminPod string,
 		return false
 	}, timeout)
 }
+
+// VerifyLabels checks if adminPod in namespace has all the label values in expectedLabelValues
+func VerifyLabels(t *testing.T, namespace string, adminPod string, expectedLabelValues map[string][]string) {
+	admins, err := GetDomainServersE(t, namespace, adminPod)
+	require.NoError(t, err)
+	require.Contains(t, admins, adminPod)
```
sivanov-nuodb, line 311: Are we comparing different types here?

test/minikube/minikube_kaa_additions_test.go
```diff
@@ -118,4 +120,55 @@ func TestKaaRolebindingDisabled(t *testing.T) {
...
+	kubectlOptions := k8s.NewKubectlOptions("", "", namespaceName)
+	options.KubectlOptions = kubectlOptions
+	admin := fmt.Sprintf("%s-nuodb-cluster0", helmChartReleaseName)
+	admin0 := fmt.Sprintf("%s-0", admin)
+
+	testlib.VerifyLabels(t, namespaceName, admin0,
+		map[string][]string{
+			"node":   currentNodes,
+			"zone":   currentZones,
+			"region": currentRegions,
```
sivanov-nuodb, line 147: I'm not sure why these are _plural_ because one AP can be running in a single region, zone, and node. I guess this is working because in _minikube_ there is only one node.

sivanov-nuodb (approved):
Looks good.
I added some comments on the tests.

stable/admin/README.md
```diff
@@ -108,6 +108,7 @@ The following tables list the configurable parameters for the `nuodb` option:
 | `serviceAccount` | The name of the service account used by NuoDB Pods | `nuodb` |
 | `addServiceAccount` | Whether to create a new service account for NuoDB containers | `true` |
 | `addRoleBinding` | Whether to add role and role-binding giving `serviceAccount` access to Kubernetes APIs (Pods, PersistentVolumes, PersistentVolumeClaims, StatefulSets) | `true` |
+| `addClusterRoleBinding` | If `addRoleBinding` is enabled, add addition cluster role and cluster role-binding giving `serviceAccount` access to Cluster wide Kubernetes APIs (read access to Nodes) | `true` |
```
adriansuarez, line 111: Nit: "add addition" -> "add"

adriansuarez:
> Yes, but if we do, we should also have a way to specify admin labels. I will do that as a separate commit to this PR.

I don't think those two features need to be coupled together, though it would enable end-to-end testing of the affinity logic in a test that has multiple APs.

I think it would make sense for that to be a separate PR.

agoelzer (approved):

test/testlib/nuodb_admin_utilities.go
kontaras: I am not sure what you mean.

test/testlib/minikube_utilities.go
kontaras: It returns the list of all the current node values. You are right that returning a mapping of node name to current value would be better, because it makes some other changes I want to make easier.

test/minikube/minikube_kaa_additions_test.go
kontaras: There will be multiple values in the off-chance that someone tries to run the test in a multi-region cluster.

test/testlib/nuodb_admin_utilities.go
sivanov-nuodb: `admins` is `map[string]NuoDBServer` and `adminPod` is `string`. Does `require.Contains` magically search the expected string in the map keys?

test/minikube/minikube_kaa_additions_test.go
sivanov-nuodb: You could add a test (in another PR) that runs against the AWS multi-cluster and exercises the admin <-> engine affinity and labels on a multi-region/multi-node cluster.

test/testlib/nuodb_admin_utilities.go
kontaras: It does. From the [docs](https://pkg.go.dev/github.com/stretchr/testify@v1.8.4/require#Contains):
`assert.Contains(t, {"Hello": "World"}, "Hello")`

kontaras: It is iterating over all map keys: https://github.com/stretchr/testify/blob/b074924938f86d417f1c9a845c7e8b0784d7f937/assert/assertions.go#L904

PR #369, opened 2024-07-29
Chart values to set admin labels and admin affinity labels

kontaras:
- Add a chart value for setting Admin Lables
- Add a chart value for setting `admin-affinity-label-keys` for database processes

### DIFF
--- a/stable/admin/README.md
+++ b/stable/admin/README.md
@@ -230,6 +230,7 @@ The following tables list the configurable parameters for the `admin` option of 
 | `tde.storagePasswordsDir` | Transparent Data Encryption storage passwords mount path | `/etc/nuodb/tde` |
 | `evicted.servers` | A list of evicted servers excluded from RAFT consensus. Used during disaster recovery. | `[]` |
 | `resourceLabels` | Custom labels attached to the Kubernetes resources installed by this Helm chart. The labels are immutable and can't be changed with Helm upgrade | `{}` |
+| `labels` | Admin labels to apply to each Admin Process. Supported starting from NuoDB image version 6.0.3 | `{}`|
 
 For example, when using GlusterFS storage class, you would supply the following parameter:
 
--- a/stable/admin/README.md
+++ b/stable/admin/README.md
@@ -230,7 +230,7 @@ The following tables list the configurable parameters for the `admin` option of 
 | `tde.storagePasswordsDir` | Transparent Data Encryption storage passwords mount path | `/etc/nuodb/tde` |
 | `evicted.servers` | A list of evicted servers excluded from RAFT consensus. Used during disaster recovery. | `[]` |
 | `resourceLabels` | Custom labels attached to the Kubernetes resources installed by this Helm chart. The labels are immutable and can't be changed with Helm upgrade | `{}` |
-| `labels` | Admin labels to apply to each Admin Process. Supported starting from NuoDB image version 6.0.3 | `{}`|
+| `adminLabels` | Admin labels to apply to each Admin Process. Supported starting from NuoDB image version 6.0.3 | `{}`|
 
 For example, when using GlusterFS storage class, you would supply the following parameter:
 
--- a/test/integration/template_database_test.go
+++ b/test/integration/template_database_test.go
@@ -3221,3 +3221,52 @@ func TestDatabaseStatefulSetVolumeSnapshot(t *testing.T) {
 		}
 	})
 }
+func TestDatabaseAdminAffinityLabels(t *testing.T) {
+	// Path to the helm chart we will test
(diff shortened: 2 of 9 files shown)


test/minikube/minikube_long_admin_test.go
```diff
@@ -56,3 +59,52 @@ func TestKubernetesBasicAdminThreeReplicas(t *testing.T) {
...
+
+	defer testlib.Teardown(testlib.TEARDOWN_DATABASE)
+
+	testlib.StartDatabase(t, namespaceName, admin0, &options)
+
+	processes, err := testlib.GetDatabaseProcessesE(t, namespaceName, admin0, dbName)
+	require.NoError(t, err)
+	for _, process := range processes {
+		require.Equal(t, 1, testlib.GetStringOccurrenceInLog(t, namespaceName, process.Hostname,
+			"Looking for admin with labels matching: zone host", &v12.PodLogOptions{}))
```
adriansuarez, line 108: This is showing that the command arg plumbing is correct, but I think we can adjust the logging to display more useful information.

Instead of this logging, which would always get displayed, I think we should display the label that matched if any. E.g. if `zone` matched:
> Using AP <server ID> with matching zone <zone>

The presence or absence of the logging would show that we applied the policy correctly and are not just connecting to the correct admin by accident or because we have no other option.

adriansuarez (approved):
Looks good, but see my suggestion about testing and logging.

test/minikube/minikube_long_admin_test.go
kontaras: I like that suggestion. Will do.

stable/admin/README.md
```diff
@@ -230,6 +230,7 @@ The following tables list the configurable parameters for the `admin` option of
 | `tde.storagePasswordsDir` | Transparent Data Encryption storage passwords mount path | `/etc/nuodb/tde` |
 | `evicted.servers` | A list of evicted servers excluded from RAFT consensus. Used during disaster recovery. | `[]` |
 | `resourceLabels` | Custom labels attached to the Kubernetes resources installed by this Helm chart. The labels are immutable and can't be changed with Helm upgrade | `{}` |
+| `labels` | Admin labels to apply to each Admin Process. Supported starting from NuoDB image version 6.0.3 | `{}`|
```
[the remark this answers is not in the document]
kontaras: Yeah, redundancy in the path or an ambiguous variable name. I don't have a preference, so I will make the change.

test/integration/template_database_test.go
```diff
@@ -3221,3 +3221,52 @@ func TestDatabaseStatefulSetVolumeSnapshot(t *testing.T) {
...
+
+		found := false
+		for i, arg := range container.Args {
+			if arg == "--admin-affinity-label-keys" {
+				assert.Equal(t, "zone host", container.Args[i+1])
+				found = true
+			}
+		}
+
+		assert.True(t, found)
```
sivanov-nuodb, line 3252: Nit: Can be replaced with:
```
assert.True(t, testlib.ArgContains(args, "--admin-affinity-label-keys"))
assert.True(t, testlib.ArgContains(args, "zone host"))
```

test/integration/template_database_test.go
```diff
@@ -3221,3 +3221,52 @@ func TestDatabaseStatefulSetVolumeSnapshot(t *testing.T) {
...
+	}
+
+	output = helm.RenderTemplate(t, options, helmChartPath, "release-name", []string{"templates/deployment.yaml"})
```
sivanov-nuodb, line 3255: Where is `helmChartPath` variable set to `testlib.ADMIN_HELM_CHART_PATH`?

sivanov-nuodb (approved):
Looks good.
I've added some minor suggestions.

test/integration/template_database_test.go
kontaras: I want to validate that the args are in the right order and don't have any other flags in between (for example, I did not mess up the order and put `zone host --admin-affinity-label-keys` or I inserted another arg later and ended up with `--admin-affinity-label-keys --some-other-flag zone host`).

test/integration/template_database_test.go
kontaras: Line 3226:
`helmChartPath := testlib.DATABASE_HELM_CHART_PATH`

This test is looking at the TE start arguments, so I am using the database helm chart, not the admin. Am I wrong?

adriansuarez:
@kontaras I pushed my fix for the upgrade and TLS tests. You need to merge that in to get a green build.

test/integration/template_database_test.go
sivanov-nuodb: Please disregard my comment.

PR #371, opened 2024-08-14
Generate unique names for ClusterRoles and CRBindings

kontaras:
To avoid conflicts between multiple admin deployments, generate unique names for ClusterRole and ClusterRoleBinding by prepending the namespace and helm release name.

### DIFF
--- a/stable/admin/templates/role.yaml
+++ b/stable/admin/templates/role.yaml
@@ -33,7 +33,8 @@ rules:
   - get
   - create
   - update
-{{- if eq (include "defaulttrue" .Values.nuodb.addClusterRoleBinding) "true" }}
+{{- if and (eq (include "defaulttrue" .Values.nuodb.addClusterRoleBinding) "true")
+      (not (lookup "rbac.authorization.k8s.io/v1" "ClusterRole" "" "nuodb-kube-inspector")) }}
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
--- a/stable/admin/templates/role.yaml
+++ b/stable/admin/templates/role.yaml
@@ -40,7 +40,7 @@ rules:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: {{ $namespace }}-{{ $adminName }}-nuodb-kube-inspector
+  name: {{ include "admin.fullclustername" . }}-kube-inspector
 rules:
 - apiGroups:
   - ""
--- a/stable/admin/templates/role.yaml
+++ b/stable/admin/templates/role.yaml
@@ -34,13 +34,13 @@ rules:
   - create
   - update
 {{- if eq (include "defaulttrue" .Values.nuodb.addClusterRoleBinding) "true" }}
-{{- $namespace := default .Release.Namespace .Values.admin.namespace -}}
-{{- $release := .Release.Name }}
+{{- $namespace := default .Release.Namespace .Values.admin.namespace  | trunc 50 | trimSuffix "-" -}}
+{{- $adminName := include "admin.fullname" . }}
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: {{ $namespace }}-{{ $release }}-nuodb-kube-inspector
+  name: {{ $namespace }}-{{ $adminName }}-nuodb-kube-inspector
 rules:
 - apiGroups:
   - ""
--- a/stable/admin/templates/role.yaml
+++ b/stable/admin/templates/role.yaml
@@ -33,13 +33,14 @@ rules:
   - get
   - create
   - update
-{{- if and (eq (include "defaulttrue" .Values.nuodb.addClusterRoleBinding) "true")
-      (not (lookup "rbac.authorization.k8s.io/v1" "ClusterRole" "" "nuodb-kube-inspector")) }}
+{{- if eq (include "defaulttrue" .Values.nuodb.addClusterRoleBinding) "true" }}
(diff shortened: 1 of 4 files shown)


stable/admin/templates/role.yaml
```diff
@@ -34,11 +34,13 @@ rules:
...
   - update
 {{- if eq (include "defaulttrue" .Values.nuodb.addClusterRoleBinding) "true" }}
+{{- $namespace := default .Release.Namespace .Values.admin.namespace -}}
+{{- $release := .Release.Name }}
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: nuodb-kube-inspector
+  name: {{ $namespace }}-{{ $release }}-nuodb-kube-inspector
```
adriansuarez, line 43: Use `{{ include "admin.fullname" . }}` instead of the release name directly.

stable/admin/templates/role.yaml
```diff
@@ -34,11 +34,13 @@ rules:
   - create
   - update
 {{- if eq (include "defaulttrue" .Values.nuodb.addClusterRoleBinding) "true" }}
+{{- $namespace := default .Release.Namespace .Values.admin.namespace -}}
```
adriansuarez, line 37: Trim the namespace so that when combined with the release name and `nuodb-kube-inspector` suffix, it is not too huge to be a legal resource name. E.g.

```
{{- $namespace := default .Release.Namespace .Values.admin.namespace | trunc 50 | trimSuffix "-" -}}
```

adriansuarez (approved):
Looks good. See my comments about tweaking the prefix.

stable/admin/templates/role.yaml
kontaras: I understand the value of using `nameOverride` or `fullNameOverride` instead of release name, but why also include the `domain` and `cluster`? It seems to be creating a very long and very complicated name for the resource when all we really want is to reduce the risk of name collisions.

stable/admin/templates/role.yaml
kontaras: Why trim just the namespace and not the release name? It seems better to just truncate the full resource name.

adriansuarez: The release name is already trimmed by `admin.fullname`.

stable/admin/templates/role.yaml
adriansuarez: This is the name that is used as a prefix in every resource associated with the release. All the ConfigMaps have it as a prefix, it is used as the StatefulSet name, etc. Why use something different?

stable/admin/templates/role.yaml
kontaras: 1. It doesn't seem that kubernetes cares about lengths here. I was able to create a ClusterRole and ClusterRoleBinding with a 286 character long name.
2. If we want to keep names to under 63 characters, we can't just append a 50 character namespace to a 50 character full name without trimming. Maybe feed all of it to `truncWithHash` so that very long names are still unique?

adriansuarez (approved):
Looks good.

stable/admin/templates/rolebinding.yaml
```diff
@@ -11,14 +11,16 @@ subjects:
...
   name: {{ default "nuodb" .Values.nuodb.serviceAccount }}
 {{- if eq (include "defaulttrue" .Values.nuodb.addClusterRoleBinding) "true" }}
+{{- $namespace := default .Release.Namespace .Values.admin.namespace | trunc 50 | trimSuffix "-" -}}
+{{- $adminName := include "admin.fullname" . }}
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: nuodb-kube-inspector
+  name: {{ $namespace }}-{{ $adminName }}-nuodb-kube-inspector
```
sivanov-nuodb, line 20: This will result in something like `nuodb-admin-nuodb-cluster0-nuodb-kube-inspector`. A lot of repeating `nuodb`. Does it make sense to define another template for cluster-scoped resources? For example:

```go
{{- define "admin.fullclustername" -}}
{{- $name := include "admin.fullname" . -}}
{{- $ns := default .Release.Namespace .Values.admin.namespace | trunc 50 | trimSuffix "-" -}}
{{- if contains $name $ns -}}
{{- printf "%s" $name -}}
{{- else -}}
{{- printf "%s-%s" $name $ns -}}
{{- end -}}
{{- end -}}
```

and then use it as `{{ include "admin.fullclustername" . }}-kube-inspector`.

sivanov-nuodb (approved):
Looks good.
I've added some suggestions about adding reusable template and more integration testing.

test/integration/template_admin_test.go
```diff
@@ -1560,12 +1561,19 @@ func TestClusterRole(t *testing.T) {
 	helmChartPath := testlib.ADMIN_HELM_CHART_PATH
 
 	t.Run("testEnabled", func(t *testing.T) {
-		output := helm.RenderTemplate(t, &helm.Options{}, helmChartPath,
+		output := helm.RenderTemplate(t, &helm.Options{
+			SetValues: map[string]string{
+				"admin.fullnameOverride": "full-name",
```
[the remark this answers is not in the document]
kontaras: Good point about testing `admin.namespace`.

I used `admin.fullnameOverride` as a short cut so I would not need to control the `cloud.cluster.name` and `admin.domain`. I am assuming that the `admin.fullname` template works correctly and not testing every combination of values that it could take, just the most convenient.

stable/admin/templates/rolebinding.yaml
kontaras: Good idea

PR #374, opened 2024-08-16
Fixed two issues in TestKubernetesUpgradeFullDatabase

kontaras:
`TestKubernetesUpgradeFullDatabase/verifyProtocolVersion` was failing on NuoDB 5.1, because there is no protocol upgrade between 5.0 and 5.1. Change the test to only run on 6.0 and newer.

While I was investigating the issue, I found a race condition in the test. The the protocol upgrade is scaling up the TE count to 2 and upgrading the database version in a single Helm update. It assumes that if there the database has 2 TEs and is healthy, the upgrade is done, but it can also occur part way through the upgrade if one of the new TEs becomes running before the admin sees the second new TE.

### DIFF
--- a/test/minikube/minikube_rolling_upgrade_test.go
+++ b/test/minikube/minikube_rolling_upgrade_test.go
@@ -242,7 +242,17 @@ func TestKubernetesUpgradeFullDatabase(t *testing.T) {
 			// start two TEs so that we can supply TE preference query
 			options.SetValues["database.te.replicas"] = "2"
 			helm.Upgrade(t, &options, testlib.DATABASE_HELM_CHART_PATH, databaseHelmChartReleaseName)
-			testlib.AwaitDatabaseUp(t, namespaceName, admin0, opt.DbName, 3)
+
+			testlib.Await(t, func() bool {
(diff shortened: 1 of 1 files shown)


test/minikube/minikube_rolling_upgrade_test.go
```diff
@@ -233,14 +233,24 @@ func TestKubernetesUpgradeFullDatabase(t *testing.T) {
 		verifyAllProcessesRunning(t, namespaceName, admin0, 2)
 	})
 
-	testlib.RunOnNuoDBVersionCondition(t, ">4.2.2", func(version *semver.Version) {
+	testlib.RunOnNuoDBVersionCondition(t, ">6.0.0", func(version *semver.Version) {
```
adriansuarez, line 236: This will need upgrading every time we bump the "upgrade from" version.

adriansuarez (approved):
Looks good. Thanks for jumping on this.

sivanov-nuodb (approved):
Looks good.

sivanov-nuodb:
@kontaras I see that `TestKubernetesUpgradeFullDatabase` test is failing but I'm not sure if it is related to your change or not.

adriansuarez (approved):

PR #375, opened 2024-08-19
Don't test protocol upgrade for NuoDB 5.x

kontaras:
Split from https://github.com/nuodb/nuodb-helm-charts/pull/374

We bumped the before version for upgrade tests to be 5.0, so there is now no protocol upgrade when upgrading to 5.0-dev or 5.1. Bumping the minimum database version for protocol upgrade tests to 6.0.

### DIFF
--- a/test/minikube/minikube_rolling_upgrade_test.go
+++ b/test/minikube/minikube_rolling_upgrade_test.go
@@ -233,7 +233,7 @@ func TestKubernetesUpgradeFullDatabase(t *testing.T) {
 		verifyAllProcessesRunning(t, namespaceName, admin0, 2)
 	})
 
-	testlib.RunOnNuoDBVersionCondition(t, ">4.2.2", func(version *semver.Version) {
+	testlib.RunOnNuoDBVersionCondition(t, ">6.0.0", func(version *semver.Version) {
 		// check that KAA will upgrade database protocol version and restart TE
 		// automatically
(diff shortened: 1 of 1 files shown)


test/minikube/minikube_rolling_upgrade_test.go
```diff
@@ -233,7 +233,7 @@ func TestKubernetesUpgradeFullDatabase(t *testing.T) {
 		verifyAllProcessesRunning(t, namespaceName, admin0, 2)
 	})
 
-	testlib.RunOnNuoDBVersionCondition(t, ">4.2.2", func(version *semver.Version) {
+	testlib.RunOnNuoDBVersionCondition(t, ">6.0.0", func(version *semver.Version) {
```
adriansuarez, line 236: Can you add a comment saying to keep this one protocol version greater than [`OLD_RELEASE`](https://github.com/nuodb/nuodb-helm-charts/blob/ccbb4afc627537d994d03b0e6a5b9b3bdee87970/test/minikube/minikube_rolling_upgrade_test.go#L27)?

The other option would be to build the version constraint so that it is `>N.0.0` where `N` is one greater than the major version appearing in `OLD_RELEASE`. I think just putting a comment here is fine.

adriansuarez (approved):

sivanov-nuodb (approved):
Looks good.

PR #377, opened 2024-10-07
Change readiness probes to accelerate process start-up

adriansuarez:
This change accelerates process start-up by making two improvements:

1. Lower the period for the readiness probe used by SMs from 15 seconds to 5 seconds and reduce the failure threshold from 2 to 1. There is no need to wait for two successes since the readiness probe checks the process state, which is verifying an explicit state transition as opposed to inferring readiness based on some transient indicator of health, e.g. issuing a basic client request. Once an SM or TE achieves Running state, it does not change until it shuts down. This is an argument for replacing the readiness probe to a startup probe, to avoid doing pointless readiness checks every 5 seconds for the life of the process, but that is left for future work.
2. Wait longer for AP readiness in nuosm. There is a basic readiness check that verifies that the APs have elected a Raft leader, and this was only performing 3 iterations before giving up, which would lead to the SMs entering CrashLoopBackOff, delaying database creation. This change replaces the polling loop with one that is based on a timeout and polls for 1 minute with an exponentially increasing polling interval to reduce the amount of logging noise.

### DIFF
--- a/stable/database/files/nuosm
+++ b/stable/database/files/nuosm
@@ -16,8 +16,6 @@
 : ${NUODB_DOMAIN:="nuodb"}
 : ${DB_NAME:="demo"}
 
-liveness_req="/nuodb/nuosm/liveness"
-
 DB_DIR=${NUODB_ARCHIVEDIR}
 DB_PARENTDIR=$(dirname ${DB_DIR})
 
@@ -277,7 +275,7 @@ function perform_restore() {
   if [ -n "$restore_requested" ] && [ "$myArchive" -ne "-1" ]; then
     # complete restore request for this archive in retry loop because there
     # could be concurrent requests due to all archives restored at the same time
-    retry 5 completeRestoreRequest "$myArchive"
+    retry 10 completeRestoreRequest "$myArchive"
     retval=$?
     if [ $retval -ne 0 ]; then
       $undo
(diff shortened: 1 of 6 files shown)


stable/database/values.yaml
```diff
@@ -529,8 +529,20 @@ database:
     # keystore: "/etc/nuodb/keys/nuoadmin.p12"
     otherOptions: {}
 
-    #Some clusters require longer readiness probe timeouts
-    readinessTimeoutSeconds: 5
+    # Configuration for readiness probe, which controls ready status and whether
+    # the pod has traffic from Services dispatched to it.
+    readinessProbe:
```
sivanov-nuodb, line 534: We might want to update the README.md with the new Helm values.

sivanov-nuodb (approved):
Looks good.
I left some minor comments.

stable/database/templates/statefulset.yaml
```diff
@@ -253,14 +253,13 @@ spec:
...
+          periodSeconds: {{ default 5 .Values.database.sm.readinessProbe.periodSeconds }}
           exec:
             command: [ "readinessprobe" ]
-          failureThreshold: 58
-          # the SM becomes unready if it does not start within 15 minutes = 30s + 15s*58
-          successThreshold: 2
-          timeoutSeconds: {{ default 5 .Values.database.sm.readinessTimeoutSeconds }}
+          failureThreshold: {{ default 3 .Values.database.sm.readinessProbe.failureThreshold }}
+          successThreshold: {{ default 1 .Values.database.sm.readinessProbe.successThreshold }}
+          timeoutSeconds: {{ default 5 .Values.database.sm.readinessProbe.timeoutSeconds }}
```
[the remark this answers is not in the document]
sivanov-nuodb: I remember we needed to tune the timeout on OCP deployments previously. It might be fine to stop supporting the `database.te.readinessTimeoutSeconds` Helm value but we need to call that out.

stable/database/files/nuosm
```diff
@@ -295,27 +293,49 @@ function perform_restore() {
...
+    log "'$*' failed with exit code $rc and output [$out] - retrying in $interval seconds..."
+    sleep "$interval"
+    interval="$((interval * 2))"
+
+    # Clamp interval to 10 seconds or seconds until the timeout expires
+    [ "$interval" -lt 10 ] || interval=10
+    local current="$(date -u '+%s')"
+    local remaining="$((deadline - current))"
+    [ "$interval" -lt "$remaining" ] || interval="$remaining"
+    [ "$interval" -ge 1 ] || interval=1
```
kontaras, line 337: Wouldn't it make sense to cap the interval to the remaining time just before before sleeping? Otherwise you are doing an execution and then  sleeping for remaining before execution.

stable/database/templates/deployment.yaml
```diff
@@ -181,14 +181,17 @@ spec:
...
             command: [ "readinessprobe" ]
-          failureThreshold: 54
-          # the TE becomes unready if it does not start within 5 minutes = 30s + 5s*54
-          successThreshold: 2
+          failureThreshold: {{ default 3 .Values.database.te.readinessProbe.failureThreshold }}
+          successThreshold: {{ default 1 .Values.database.te.readinessProbe.successThreshold }}
+          {{- if .Values.database.te.readinessProbe.timeoutSeconds }}
+          timeoutSeconds: {{ .Values.database.te.readinessProbe.timeoutSeconds }}
+          {{- else }}
           timeoutSeconds: {{ default 5 .Values.database.te.readinessTimeoutSeconds }}
```
kontaras, line 193: Add a comment that this is the old value?

kontaras (approved):

stable/database/files/nuosm
adriansuarez: Makes sense. I updated the polling loop to clamp the value before sleeping.

stable/database/files/nuosm
```diff
@@ -511,11 +531,10 @@ if [ "$SNAPSHOT_RESTORED" == "true" ]; then
...
 fi
 
-# Retry the first request against the admin layer to make sure that Raft
-# commands can be committed and fail fast on error
-retry 3 nuocmd get value --key "$liveness_req" 2>&1
-if [ $? -ne 0 ]; then
-  die 1 "NuoDB Admin layer liveness check failed"
+# Perform a consistent read of an arbitrary KV-store entry in a polling loop to
+# confirm Raft leadership before proceeding with SM start-up.
+if ! retry 60 nuocmd get value --key nuodb/nuosm/readiness; then
```
NikTJ777, line 536: Nit: In the ds-nuodb component, I am using nuocmd check servers:

status="$(nuocmd check servers --check-leader --timeout "$timeout" 2>&1)"
  if [ $? -ne 0 ]; then
    die 1 "Admin layer is inoperative - exiting: ${status}"
  fi

In general, I try to use the timeouts built into nuocmd/nuodocker rather than a polling loop in the script.

Since retry() is used in multiple places, I'm not sure it's worth changing it in this PR.

NikTJ777 (approved):
Looks good!

stable/database/files/nuosm
adriansuarez: Agreed, but I tend to be conservative with changes that can have a big impact. This code path is exercised every time an SM starts up, so I didn't want to introduce too many changes in behavior.

PR #378, opened 2024-11-04
Add test of nuodocker functionality to delete old archives

adriansuarez:
This change adds a test of `nuodocker start sm` functionality that detects and deletes the old archive object associated with a PVC. This functionality will be delivered in 7.0 and perhaps backported to 6.0.3.

If the NuoDB version does not support this functionality, then the test asserts on the expected current behavior, which is that database restart is blocked waiting for an SM to be started on the defunct archive.

### DIFF
--- a/test/minikube/minikube_base_database_test.go
+++ b/test/minikube/minikube_base_database_test.go
@@ -279,6 +279,93 @@ func TestKubernetesAccessWithinPods(t *testing.T) {
 	})
 }
 
+func TestArchivePvcRecreated(t *testing.T) {
+	testlib.AwaitTillerUp(t)
+	options := helm.Options{}
+	defer testlib.Teardown(testlib.TEARDOWN_ADMIN)
+	helmChartReleaseName, namespaceName := testlib.StartAdmin(t, &options, 1, "")
+	admin0 := fmt.Sprintf("%s-nuodb-cluster0-0", helmChartReleaseName)
+
+	// Start database and wait for it to become running
+	defer testlib.Teardown(testlib.TEARDOWN_DATABASE)
+	databaseOptions := helm.Options{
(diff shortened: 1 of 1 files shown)


agoelzer (approved):

test/minikube/minikube_base_database_test.go
```diff
@@ -279,6 +279,93 @@ func TestKubernetesAccessWithinPods(t *testing.T) {
...
+
+		// Look for expected SM logging
+		out, err := k8s.RunKubectlAndGetOutputE(t, kubectlOptions, "logs", smPodName, "-c", "engine")
+		require.NoError(t, err)
+		require.Contains(t, out, "Deleting archive ID 0 associated with PVC "+archivePvcName)
+	})
+	// Otherwise, database restart will be blocked in
+	// AWAITING_ARCHIVE_HISTORIES_MSG state waiting for the archive for the
+	// deleted PVC to be started.
+	t.Run("testOldArchiveNotDeleted", func(t *testing.T) {
```
kontaras, line 351: Why test that the old behavior is still broken?

kontaras (approved):

test/minikube/minikube_base_database_test.go
adriansuarez: I wanted to confirm that the test actually reproduces the issue and understand what the symptom is. I think the test adds value by showing what happens on current versions.

I guess I could have deleted the sub-test after confirming locally and then conditionalized the entire test on the image having the fix. That would save a little bit of time on the test runs until we upgrade to 7.0 or 6.0.3 in the CI runs.

sivanov-nuodb (approved):
Looks good.
Thanks for adding a test!

PR #379, opened 2024-11-06
Specify connect-timeout for curl to avoid waiting indefinitely

adriansuarez:


adriansuarez:
It looks like `curl` hangs indefinitely now when fetching "sftp://sftp.nuodb.com/incoming/restore.bak.tz", and this is the reason for the TestKubernetesImportDatabase/startDatabaseStatefulSet failure.

kontaras (approved):

sivanov-nuodb (approved):
Thanks for fixing this!

PR #382, opened 2024-11-13
Update dependencies and other maintenance

adriansuarez:
- Bump versions exercised in Helm upgrade test.
- Update dependencies and consolidate some dependencies. For example, we had two YAML parsers with similar APIs that were being imported.
- Remove usage of deprecated ioutil functions.
- Reorganize imports according to Golang conventions, and use consistent import aliases for packages.
- Remove code related to obsolete Tiller operator.


sivanov-nuodb (approved):
Thanks. for doing the housekeeping!

test/testlib/NuoDBRegistryEntry.go
```diff
@@ -15,10 +15,8 @@ type Registry struct {
 }
 
 // UnmarshalImageYAML is used to unmarshal into map[string]string
-func UnmarshalImageYAML(s string) (err error, registry Registry) {
-	registry = Registry{}
-
-	err = yaml.Unmarshal([]byte(s), &registry)
-
-	return
```
agoelzer, line 23: change makes sense here, wondering how this worked before without returning anything?

agoelzer (approved):

test/testlib/NuoDBRegistryEntry.go
adriansuarez: Previously the return types had the local variables `err` and `registry` associated with them, which were specified on the function signature. So simply setting these variables was enough to generate the correct return value. But this is a weird syntax that we don't use anymore.

kontaras (approved):
Many of the changes look like formatting fixes. Can we set up some automation so that future work conform to this style?

adriansuarez:
> Many of the changes look like formatting fixes. Can we set up some automation so that future work conform to this style?

Good idea.

PR #383, opened 2024-12-11
Relax and sharpen version string checks for NuoDB binaries

madscientist:
Rather than relying on the complete string match when extracting the release from the --version output, test only against the "build" part. Require versions to start with a number, or don't match. Releases must be a single word so stop matching at whitespace.


sivanov-nuodb (approved):
Looks good.
Thanks for relaxing the version regex!

PR #391, opened 2025-05-28
Crossport SD-Helm changes to Helm repo

kontaras:


stable/admin/templates/_helpers.tpl
```diff
@@ -369,16 +369,44 @@ networking.gke.io/load-balancer-type: "Internal"
 Renders the labels for all resources deployed by this Helm chart
 */}}
 {{- define "admin.resourceLabels" -}}
-app: {{ template "admin.fullname" . }}
+{{- include "admin.labels" (dict "Root" . "ExtraLabels" .Values.admin.resourceLabels ) -}}
+{{- end -}}
+
+{{- define "admin.labels" -}}
+{{- if not .Root}}{{fail "<nuodb.env> Root is required"}}{{end}}
```
adriansuarez, line 376: Where does this get defined? I don't see this in the Helm documented as a [built-in object](https://helm.sh/docs/chart_template_guide/builtin_objects/), and I don't see it being defined here unless I missed something.

Under what circumstances would this check fail?

test/integration/template_resources_test.go
```diff
@@ -158,8 +166,6 @@ func TestResourcesDatabaseOverridden(t *testing.T) {
 		assert.EqualValues(t, 8, (*containers)[0].Resources.Limits.Cpu().ScaledValue(0))
 		assert.EqualValues(t, 16*1024*1024*1024, (*containers)[0].Resources.Limits.Memory().ScaledValue(0))
 
-		assert.True(t, testlib.ArgContains((*containers)[0].Args, "mem 4Gi"))
```
adriansuarez, line 161: Why did this change?

adriansuarez (approved):
Looks good. I left some minor changes.

Does it also make sense to regression test using the NuoDB Control Plane tests?

PR #392, opened 2025-06-10
Add test case exercising network policy

adriansuarez:
This change adds a test case that verifies that NuoDB domains and databases can be created with network policies that enhance security by limiting external connectivity.

### DIFF
--- a/test/minikube/minikube_networkpolicy_test.go
+++ b/test/minikube/minikube_networkpolicy_test.go
@@ -1,0 +1,119 @@
+//go:build long
+// +build long
+
+package minikube
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/gruntwork-io/terratest/modules/helm"
+	"github.com/gruntwork-io/terratest/modules/k8s"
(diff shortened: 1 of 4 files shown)


test/minikube/minikube_networkpolicy_test.go
```diff
@@ -0,0 +1,119 @@
...
+	defer testlib.VerifyTeardown(t)
+	defer testlib.Teardown(testlib.TEARDOWN_ADMIN)
+	adminReleaseName, namespace := testlib.StartAdmin(t, &helm.Options{
+		SetValues: map[string]string{
+			"admin.replicas": "2",
+		},
+	}, 2, "")
+	admin0 := fmt.Sprintf("%s-nuodb-cluster0-0", adminReleaseName)
+
+	// Create network policy that limits connectivity to group=nuodb
```
sivanov-nuodb, line 37: Should the policy be created even before installing the admin?

sivanov-nuodb (approved):
Looks good.
Thanks for also fixing the `nuote` script!

kontaras (approved):

PR #394, opened 2025-06-18
Do not use terratest library to create K8s client

adriansuarez:
This change is to avoid the useless logging generated by the terratest library.


adriansuarez:
I just realize that every `kubectl` invocation done by terratest also generates the same logging, and we have many more invocations of those methods than we do of the ones being replaced by this test. I'll abandon this and just filter the useless output when collecting logs.

PR #397, opened 2025-07-23
Crossport latest changes from ds-nuodb

kontaras:
Includes hooks to add extra annotations to services and generate a password.


adriansuarez (approved):
Looks good.

PR #399, opened 2025-07-28
Add example of TE groups to the doc

paulc4:
See title.

### DIFF
--- a/stable/admin/templates/_helpers.tpl
+++ b/stable/admin/templates/_helpers.tpl
@@ -599,3 +599,12 @@ Resources requested and limited
 {{- define "admin.resources" -}}
 {{- toYaml .Values.admin.resources -}}
 {{- end -}}
+
+{{/*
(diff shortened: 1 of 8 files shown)


docs/HowToConnectExternally.md
```diff
@@ -28,6 +28,134 @@ This allows flexible configuration of each TE group including but not limited to
 SQL clients can be configured to target each TE group separately using NuoDB Admin load balancer rules.
 Specifying the type of the Helm `database` release is controlled by the  `database.primaryRelease` option (`true` by default).
 
+Here is an example.
```
sivanov-nuodb, line 31: There is an example already [in the examples section](https://github.com/nuodb/nuodb-helm-charts/blob/80978faeb39eb3605d25b88fb1c79610ed9d1623/docs/HowToConnectExternally.md#examples). Is something missing there?
If yes, would it be better to update the existing example instead?

PR #400, opened 2025-08-01
Support static NuoDB collector configuration

sivanov-nuodb:
**Changes**

This change enables the NuoDB collector to be configured with plugins
statically by mounting ConfigMap resources directly in the container. It is
useful for deployments where it is not needed to pull the Telegraf plugins
dynamically (i.e. all plugins are known at the Helm release installation time).
The immediate benefit of such a configuration is disabling the NuoDB collector
config watcher sidecar, to reduce the overall cost at scale and minimize
the load on the Kubernetes API server.

The plugin configuration changes are automatically propagated inside the
container by kubelet, however, Telegraf must be started with `--watch-config`
option. See https://docs.influxdata.com/telegraf/v1/commands/

**Testing**
- integration tests
- E2E tests

### DIFF
--- a/stable/admin/values.yaml
+++ b/stable/admin/values.yaml
@@ -440,6 +440,10 @@ nuocollector:
     tag: 2.0.0
     pullPolicy: IfNotPresent
   watcher:
+    # Whether to enable NuoDB Collector configuration watcher. If disabled,
+    # decired output plugins must be defined during the installation of the
+    # NuoDB Helm Charts release
+    enabled: true
     registry: ghcr.io
     repository: nuodb/nuodb-sidecar
     tag: latest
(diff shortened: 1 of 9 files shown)


kontaras (approved):

stable/admin/values.yaml
```diff
@@ -440,6 +440,10 @@ nuocollector:
     tag: 2.0.0
     pullPolicy: IfNotPresent
   watcher:
+    # Whether to enable NuoDB Collector configuration watcher. If disabled,
+    # decired output plugins must be defined during the installation of the
```
adriansuarez, line 444: decired -> desired

adriansuarez (approved):
Looks good.

PR #401, opened 2025-08-06
Misc changes to support BOaaS (external backup manager)

sivanov-nuodb:
**Changes**

This change adds support for database in-place restore done using snapshots.  A
special `:snapshot` restore source signals that the database data has been
restored from snapshots by an external backup manager (e.g. BOaaS). Expose the
backup hooks port as Helm value.

**Testing**
- regrestion testing

### DIFF
--- a/stable/database/files/nuosm
+++ b/stable/database/files/nuosm
@@ -148,46 +148,48 @@ function perform_restore() {
 
   log "Restoring $restore_source; existing archive directores: $( ls -l $DB_PARENTDIR )"
 
-  # work out available space
-  archSize="$(du -s $DB_DIR | grep -o '^ *[0-9]\+')"
-  archSpace="$(df --output=avail $DB_DIR | grep -o ' *[0-9]\+')"
-
-  if [ $(( archSpace > archSize * 2)) ]; then
-    saveName="${DB_DIR}-save-$( date +%Y%m%dT%H%M%S )"
-    undo="mv $saveName $DB_DIR"
-    mv $DB_DIR $saveName
-
-    retval=$?
-    if [ $retval -ne 0 ]; then
-      $undo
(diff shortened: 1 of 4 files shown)


stable/database/files/nuosm
sivanov-nuodb, line 1: Please review the file with "Hide whitespaces" checked.

stable/database/files/nuosm
```diff
@@ -576,12 +605,20 @@ if [ "$myArchive" -ne "-1" ]; then
       # restore chart doesn't have notion of stream|backupset
       # `nuodocker restore archive` works with both
       restore_type="backupset"
+      if [ "$restore_source" == ":snapshot" ]; then
+        # the archive content has been restored in-place from a snapshot by an
+        # external backup manager (not by NuoDB Helm charts or NuoDB operator);
+        # we will get here only if the Kubernetes Aware Admin (KAA) is disabled
+        # since otherwise the archive metadata will be removed by KAA when the
+        # PVC is re-created with a snapshot reference
+        restore_type="snapshot"
```
adriansuarez, line 614: If KAA is enabled, is this unnecessary? Or does KAA have to be disabled?

adriansuarez (approved):
Looks good.

stable/database/files/nuosm
sivanov-nuodb: KAA has to be disabled. Otherwise, the old archives (those for which restore is requested) are purged when BOaaS re-creates the PVCs.

kontaras (approved):

stable/database/files/nuosm
sivanov-nuodb: The database restore request is the KV store, so it should be straightforward to check it. I will make a change.

PR #404, opened 2025-09-18
Supply restricted `node-port` range to database processes

adriansuarez:
Only port 48006 is exposed on database pods, but database processes are supplied the default `node-port` value 48006, which actually defines a range of ports starting at 48006 with no upper bound (except the maximum port value allowed by the operating system).

This change specifies the `portRange` for the APs to be `48006,48006` (the `node-port` value is based on the `portRange` AP property), so that the specific port that is exposed on database pods is what the processes listen on. If for some reason, a process cannot listen on that port, then it should just crash and be rescheduled.

Previously, if a process was unable to listen on 48006, it would try 48007, and so on, which is unreachable and prevents any of its peers in the database from connecting to it.

### DIFF
--- a/stable/admin/templates/statefulset.yaml
+++ b/stable/admin/templates/statefulset.yaml
@@ -160,6 +160,7 @@ spec:
           {{- end }}
           {{- end }}
           - "--"
+          - "portRange=48006,48006"
           {{- if .Values.admin.tde }}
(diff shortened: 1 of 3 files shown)


agoelzer (approved):

sivanov-nuodb (approved):
Looks good.

PR #416, opened 2026-04-15
[HELM-471] Added backup hooks sidecar to the TE pod

kontaras:
Add a `nuodb-sidecar` container to the TE Helm chart.

Testing done:
https://app.circleci.com/pipelines/github/nuodb/nuodb-helm-charts?branch=ktarashchanskiy%2FHELM-471_TE-sidecar
https://app.circleci.com/pipelines/github/nuodb/nuodb-control-plane?branch=ktarashchanskiy%2FHELM-471_TE-sidecar


stable/database/templates/_helpers.tpl
```diff
@@ -1340,3 +1345,100 @@ Set tolerations for the database TE pods.
...
+  - name: journal-volume
+    mountPath: /mnt/journal
+  {{- end }}
+  {{- end }}
+  {{- if eq (include "backupHooks.freezeMode" $root ) "hotsnap" }}
+  {{- if eq (include "database.enableEphemeralVolume" (list $root $root.Values.database.sm)) "true" }}
+  - name: eph-volume
+    mountPath: /tmp
+    subPath: tmp-hooks
+  {{- end }}
```
adriansuarez: Is this conditionalized on SM vs TE? It looks like we are creating an ephemeral volume for TEs just because it is required for SMs.

stable/database/templates/_helpers.tpl
```diff
@@ -1340,3 +1345,100 @@ Set tolerations for the database TE pods.
...
+      valueFrom:
+        secretKeyRef:
+          name: {{ template "database.secretName" $root }}
+          key: database-name
+    - name: FREEZE_MODE
+      value: {{ include "backupHooks.freezeMode" $root | quote }}
+    {{- if $root.Values.database.backupHooks.timeout }}
+    - name: FREEZE_TIMEOUT
+      value: {{ $root.Values.database.backupHooks.timeout | quote }}
+    {{- end }}
```
adriansuarez: These aren't relevant for TEs.

stable/database/templates/_helpers.tpl
```diff
@@ -1340,3 +1345,100 @@ Set tolerations for the database TE pods.
...
+    - name: NUODB_ARCHIVE_DIR
+      value: {{ printf "/mnt/archive/%s/%s" ( include "admin.domainName" $root ) ( include "database.dbName" $root ) }}
+    {{- if eq (include "defaultfalse" $root.Values.database.sm.noHotCopy.journalPath.enabled) "true" }}
+    - name: NUODB_JOURNAL_DIR
+      value: {{ printf "/mnt/journal/%s/%s" ( include "admin.domainName" $root ) ( include "database.dbName" $root ) }}
+    {{- end }}
+    {{- end }}
+    {{- if eq (include "backupHooks.freezeMode" $root ) "hotsnap" }}
+    - { name: NUOCMD_API_SERVER,   value: "{{ template "admin.address" $root }}:8888" }
+    {{- end }}
```
adriansuarez: Currently we only need `nuocmd` in order to invoke `nuocmd pause/resume archiving`.

We may want to have access to `nuocmd` for diagnostics, but in that case this code would change for both SMs and TEs. I would defer adding `nuocmd` to whatever change also adds `nuocmd` unconditionally for diagnostic purposes.

stable/database/templates/_helpers.tpl
```diff
@@ -1340,3 +1345,100 @@ Set tolerations for the database TE pods.
...
+  resources:
+  {{- ( include "database.backupHooks.resources" $root ) | trim | nindent 4 }}
+  {{- if and
+      (eq (include "defaultfalse" $root.Values.database.sm.noHotCopy.journalPath.enabled) "true")
+      (eq (include "backupHooks.freezeMode" $root ) "fsfreeze") }}
+  securityContext:
+    privileged: true
+    runAsUser: 0
+    runAsGroup: 0
+  {{- else }}
```
adriansuarez: This is more code that is not relevant for TEs.

I think we should just render them separately. I don't see much code reuse if we have to conditionalize every other block based on engine type, and rendering them separately makes it easier to read.

adriansuarez:
This does not allow the sidecar to be controlled independently on SMs and TEs. Everything seems to be controlled by the `database.backupHooks` value, which has values that are irrelevant for TEs. The user should be able to register different custom handlers on SMs and TEs and enable them independently of each other.

This is a good opportunity to add a value like `database.te.operationsSidecar`, and we can also add `database.sm.operationsSidecar` that takes precedence over `database.backupHooks` or is merged with it. Perhaps we keep the backup-specific values in `database.backupHooks` such as `freezeMode`, and add `customHandlers` to `database.sm.operationsSidecar` which can be merged with `database.backupHooks.customHandlers`.

stable/database/templates/_helpers.tpl
kontaras: There are only 2 conditional blocks: SM specific environment variables and volume mounts. I just need to move the start and end of the `if` blocks a bit.

stable/database/templates/_helpers.tpl
kontaras: I will conditionalize it.

stable/database/templates/_helpers.tpl
kontaras: I will conditionalize it.

stable/database/templates/_helpers.tpl
adriansuarez: You're saying that there will only be two once the sections are conditionalized that are currently being rendered for TEs but are not relevant? There are already two, and based on the structure of the YAML, I suspect there will be more and it will become more difficult to read and maintain.

I think they have more differences than similarities, and it is just two containers. Having two separate easy-to-read sections is better than one difficult-to-read section. Also, once the sidecars are controlled by different values, it will be even more difficult to reuse anything.

kontaras:
> This does not allow the sidecar to be controlled independently on SMs and TEs. Everything seems to be controlled by the `database.backupHooks` value, which has values that are irrelevant for TEs. The user should be able to register different custom handlers on SMs and TEs and enable them independently of each other.
> 
> This is a good opportunity to add a value like `database.te.operationsSidecar`, and we can also add `database.sm.operationsSidecar` that takes precedence over `database.backupHooks` or is merged with it. Perhaps we keep the backup-specific values in `database.backupHooks` such as `freezeMode`, and add `customHandlers` to `database.sm.operationsSidecar` which can be merged with `database.backupHooks.customHandlers`.

You are right, it makes sense to rename everything under `database.backupHooks` other than `enabled` (which is needed for both), `useSuspend`, `freezeMode`, and `timeout` to `database.operationsSidecar`. To avoid breaking existing code, we can have an explicitly provided `backupHooks` value take precedent (or be appended, for values like `customHandlers`.

I don't see the value in having separate `te` and `sm` values other than for `enabled` and `customHandlers`.

adriansuarez:
> I don't see the value in having separate `te` and `sm` values other than for `enabled` and `customHandlers`.

We should preserve `database.backupHooks` for backwards-compatibility and for configuring the backup-specific aspects of the sidecar, which is basically `freezeMode` and `timeout`. The following sub-values also appear in `backupHooks` and should be configurable independently:

- `resources`
- `port`
- `customHandlers`

I don't know / remember what `volumeMounts` is for, so I'm not sure about that one. If it is to allow custom scripts that may be invoked by custom handlers, then it should also be configurable independently.

I think we should also allow name of the sidecar to be specified, so that we do not have to give the TE sidecar the name `backup-hooks`, but we can keep that name on the SM for backwards compatibility reasons.

My proposal is:

- `database.backupHooks` only controls the SM sidecar. We cannot have the TE silently having a new sidecar without any change in values, because that is potentially a breaking change. We saw that it has an impact due to resource usage, but it maybe also break scripting that expects the TE deployment to only have 1 container, or 2 containers including `nuocollector`.
- If `database.sm.operationsSidecar.enable=true`, then `database.sm.operationsSidecar.*` takes precedence over the corresponding `backupHooks.*` value for all the values they have in common.
  - Also name the sidecar container `database.sm.operationsSidecar.name`.
  - Instead of overriding `backupHooks.customHandlers`, consider merging it with `database.sm.operationsSidecar.customHandlers`.
- The TE sidecar is completely controlled by `database.te.operationsSidecar`. `database.backupHooks` has no impact on it.

kontaras:
Before we get caught up on what value names should be, lets agree on what values should exist. In my opinion, out of the existing `backupHooks` values:
- `useSuspend`, `freezeMode`, and `timeout` are backup specific and should not be renamed.
- `image` and `port` can have a single value for both SM and TE.
- `resources`, `env`, and `volumeMounts` can also be shared between the SM and TE. If a user wants to really optimize, they could set different values per pod type, but the gains would be minimal.
- I agree that there is value in having `enabled` and `customHandlers` be configurable per pod type but it is just as useful to configure a handler on all pods without copy/pasting. I would split them into `enabledSm`, `enabledTe`, `customHandlers`, `customHandlersSm`, and `customHandlersTe`.

For the sake of completes, I should also add an operations sidecar to hot copy SMs.

By workload type, it would be:

- No Hot Copy SM
  - Enabled by either `backupHooks.enabled` or `operationsSidecar.enabledSm`
  - Gets `backupHooks.customHandlers`, `customHandlers`, and `customHandlersSm`
  - Has SM built-in handlers, based on `backupHooks.enabled`
  - Sidecar container name is `backup-hooks`
- Hot Copy SM
  - Enabled by `operationsSidecar.enabledSm`
  - Gets `customHandlers`, and `customHandlersSm`
  - No SM built-in handlers
  - Sidecar container name is `operations-sidecar`
- TE
  - Enabled by `operationsSidecar.enabledTe`
  - Gets `customHandlers`, and `customHandlersTe`
  - No SM built-in handlers
  - Sidecar container name is `operations-sidecar`

adriansuarez:
I'm not getting caught up in what the names should be, I'm making a proposal that addresses the requirement that the TE sidecar be configurable independently of the SM sidecar while also maintaining backwards compatibility with the existing `backupHooks` value.

Your counter-proposal just changes the names of the values but with more coupling between the SM and TE configurations. I don't care about the names that much (even though `database.sm.x` and `database.te.x` are more established as conventions as opposed to `database.x.ySm` and `database.x.yTe`). What I care about is that if we want to specify something differently for TEs, we are able to do that.

stable/database/templates/_helpers.tpl
```diff
@@ -904,6 +925,11 @@ Set shareProcessNamespace=true if needed for a TE deployment.
     {{- $hasSidecars = true -}}
   {{- end -}}
 {{- end -}}
+{{- if .Values.database.backupHooks -}}
```
sivanov-nuodb: Why the "backupHooks" Helm value is inspected in `database.te.shareProcessNamespace`?

stable/database/templates/_helpers.tpl
```diff
@@ -884,10 +907,8 @@ Set shareProcessNamespace=true if needed for an SM statefulset.
     {{- $hasSidecars = true -}}
   {{- end -}}
 {{- end -}}
-{{- if .Values.database.backupHooks -}}
-  {{- if eq (include "defaultfalse" .Values.database.backupHooks.enabled) "true" -}}
-    {{- $hasSidecars = true -}}
-  {{- end -}}
+{{- if eq (include "database.nuodb.sm.operationsSidecar.enabled" . ) "true" -}}
```
sivanov-nuodb: Should we enable the SM sidecar if `.Values.database.backupHooks.enabled=true` also?

stable/database/templates/_helpers.tpl
```diff
@@ -1340,3 +1385,13 @@ Set tolerations for the database TE pods.
...
 {{- end }}
+
+{{/*
+Is there an operations sidecar enabled for the SM?
+*/}}
+{{- define "database.nuodb.sm.operationsSidecar.enabled" -}}
+{{- or
+  ( eq (include "defaultfalse" .Values.database.backupHooks.enabled) "true" )
+  ( eq (include "defaultfalse" .Values.database.sm.operationsSidecar.enabled) "true" )
+-}}
```
sivanov-nuodb, line 1398: I'm not familiar with this ternary statement. Does it return `true|false` (string)?

stable/database/templates/deployment.yaml
```diff
@@ -29,10 +29,8 @@ spec:
   template:
     metadata:
       annotations:
-      {{- if .Values.database.configFiles }}
-        {{- with .Values.database.configFiles }}
-        checksum/config: {{ toYaml . | sha256sum }}
-        {{- end }}
+      {{- if (or .Values.database.configFiles .Values.database.backupHooks.enabled) }}
+        checksum/config: {{ dict "configFiles" .Values.database.configFiles "customHandlers" .Values.database.backupHooks.customHandlers | toYaml | sha256sum }}
```
sivanov-nuodb: The checksum should be calculated against `.Values.database.te.operationsSidecar.customHandlers` value.

stable/database/templates/deployment.yaml
```diff
@@ -29,10 +29,8 @@ spec:
   template:
     metadata:
       annotations:
-      {{- if .Values.database.configFiles }}
-        {{- with .Values.database.configFiles }}
-        checksum/config: {{ toYaml . | sha256sum }}
-        {{- end }}
+      {{- if (or .Values.database.configFiles .Values.database.backupHooks.enabled) }}
```
sivanov-nuodb: This should check the `.Values.database.te.operationsSidecar.enabled` instead of `.Values.database.backupHooks.enabled`.

stable/database/values.schema.json
```diff
@@ -74,6 +74,54 @@
                         }
                     }
                 },
+                "sm": {
+                  "type": "object",
+                  "properties": {
+                    "operationsSidecar": {
```
sivanov-nuodb: Nit: Does it make sense to reference and reuse the `operationsSidecar` definition using https://json-schema.org/understanding-json-schema/structuring#defs?

stable/database/values.yaml
```diff
@@ -693,6 +640,64 @@ database:
       # characters are allowed. By default the Helm release name is used.
       name: ""
 
+    operationsSidecar:
```
sivanov-nuodb, line 643: Can you please update the `README.md` with the new values?

test/integration/template_database_test.go
```diff
@@ -3040,9 +3040,37 @@ func TestBackupHooksCustomHandlersNegative(t *testing.T) {
 		require.Contains(t, err.Error(), "Error: values don't meet the specifications of the schema(s) in the following chart(s):")
 		require.Contains(t, err.Error(), `- database.backupHooks.customHandlers.0.statusMappings.1: Must be less than or equal to 599`)
 	})
+
+	t.Run("testOperations", func(t *testing.T) {
```
sivanov-nuodb: Nit: Can be renamed to `testNegativeOperations`?

test/integration/template_database_test.go
```diff
@@ -3333,6 +3368,168 @@ func TestDatabaseStatefulSetBackupHooksSidecar(t *testing.T) {
...
+					sidecar = &container
+					// Make sure shareProcessNamespace=true for pod
+					assert.NotNil(t, obj.Spec.Template.Spec.ShareProcessNamespace)
+					assert.True(t, *obj.Spec.Template.Spec.ShareProcessNamespace)
+				}
+			}
+		}
+		// Make sure securityContext appears and does not have
+		// privileged=true, which is only required when the journal
+		// volume is separate and fsfreeze mode is enabled
```
sivanov-nuodb: Nit: There is no journal volume for TEs.

stable/database/templates/_helpers.tpl
kontaras: Yep, that is a copy/paste error. I will take out the `if` block.

stable/database/templates/_helpers.tpl
kontaras: We are. Here `database.nuodb.sm.operationsSidecar.enabled` is the template defined in `_helpers.tpl` that is `backupHooks.enabled or sm.operationsSidecar.enabled`

stable/database/templates/_helpers.tpl
kontaras: If true, I am not sure if it returns a boolean `true` or the first truthy argument. 

On false, it returns a `false` boolean, but Helm templates always return a string, so the return value is cast to the string `false` which is truthy.

stable/database/templates/deployment.yaml
kontaras: Yep. I forgot to update it.

stable/database/templates/deployment.yaml
kontaras: Yep. I forgot to update it.

test/integration/template_database_test.go
kontaras: Yeah, that fits the naming convention of the other test cases better.

test/integration/template_database_test.go
kontaras: Done

stable/database/values.schema.json
kontaras: Good idea. I will try it.

stable/database/values.yaml
kontaras: Thank you for reminding me.

stable/database/templates/_helpers.tpl
sivanov-nuodb: I see.

sivanov-nuodb (approved):
Looks good.

stable/database/templates/configmap.yaml
```diff
@@ -45,7 +45,24 @@ metadata:
 data:
   handlers.json: |
     {
-      "handlers": {{ .Values.database.backupHooks.customHandlers | toJson }}
+      "handlers": {{ concat 
+        ( default (list) .Values.database.backupHooks.customHandlers )
+        .Values.database.sm.operationsSidecar.customHandlers
```
adriansuarez: Why not `( default (list) .Values.database.sm.operationsSidecar.customHandlers )` also?

stable/database/templates/statefulset.yaml
```diff
@@ -279,22 +279,23 @@ spec:
           {{- else }}
           timeoutSeconds: {{ default 5 .Values.database.sm.readinessTimeoutSeconds }}
           {{- end }}
-      {{- if eq (include "defaultfalse" .Values.database.backupHooks.enabled) "true" }}
+      {{- if eq (include "database.nuodb.sm.operationsSidecar.enabled" . ) "true" }}
       - name: backup-hooks
-        image: {{ include "backupHooks.image" . }}
-        imagePullPolicy: {{ default "" .Values.database.backupHooks.image.pullPolicy }}
+        image: {{ include "backupHooks.image" .}}
+        imagePullPolicy: {{ .Values.AsMap | dig "database" "backupHooks" "image" "pullPolicy" .Values.database.sm.operationsSidecar.image.pullPolicy | default "" }}
```
adriansuarez: Please use `| default "IfNotPresent"` as a default.

stable/database/templates/configmap.yaml
kontaras: In `values.yaml`, we do not define `database.backupHooks.customHandlers` but set `database.sm.operationsSidecar.customHandlers` to an empty list.

stable/database/templates/configmap.yaml
```diff
@@ -45,7 +45,24 @@ metadata:
 data:
   handlers.json: |
     {
-      "handlers": {{ .Values.database.backupHooks.customHandlers | toJson }}
+      "handlers": {{ concat 
+        ( default (list) .Values.database.backupHooks.customHandlers )
```
adriansuarez: We should only include the `backupHooks.customHandlers` is `backupHooks.enabled=true`.

stable/database/templates/deployment.yaml
```diff
@@ -200,6 +198,34 @@ spec:
           {{- else }}
           timeoutSeconds: {{ default 5 .Values.database.te.readinessTimeoutSeconds }}
           {{- end }}
+      {{- if eq (include "defaultfalse" .Values.database.te.operationsSidecar.enabled) "true" }}
+      - name: operations
+        image: {{ include "te.operationsSidecar.image" .}}
+        imagePullPolicy: {{ .Values.database.te.operationsSidecar.image.pullPolicy | default "" }}
```
adriansuarez: Use `IfNotPresent` as the default pull policy.

stable/database/templates/_helpers.tpl
```diff
@@ -66,9 +66,32 @@ Also, we can't use a single if because lazy evaluation is not an option
 Return the backup hooks sidecar image
 */}}
 {{- define "backupHooks.image" -}}
-{{- $registryName := .Values.database.backupHooks.image.registry -}}
-{{- $repositoryName := .Values.database.backupHooks.image.repository -}}
-{{- $tag := .Values.database.backupHooks.image.tag | toString -}}
+{{- $registryName := .Values.AsMap | dig "database" "backupHooks" "image" "registry" .Values.database.sm.operationsSidecar.image.registry -}}
+{{- $repositoryName := .Values.AsMap | dig "database" "backupHooks" "image" "repository" .Values.database.sm.operationsSidecar.image.repository -}}
+{{- $tag := .Values.AsMap | dig "database" "backupHooks" "image" "tag" .Values.database.sm.operationsSidecar.image.tag | toString -}}
```
adriansuarez: Even though the include is named `backupHooks.image`, this is used to render the SM operations sidecar regardless of whether or not `backupHooks.enabled=true`.

We should disregard the `backupHooks` values if `backupHooks.enabled=false`.

stable/database/templates/statefulset.yaml
```diff
@@ -279,22 +279,23 @@ spec:
...
-        - containerPort: {{ .Values.database.backupHooks.port }}
+        - containerPort: {{ .Values.AsMap | dig "database" "backupHooks" "port" .Values.database.sm.operationsSidecar.port }}
           protocol: TCP
         env:
           - name: DB_NAME
             valueFrom:
               secretKeyRef:
                 name: {{ template "database.secretName" . }}
                 key: database-name
+          {{- if eq (include "defaultfalse" .Values.database.backupHooks.enabled) "true" }}
```
adriansuarez: I see no reason to suppress the ability to execute pre- and post-backup hooks just because `backupHooks.enabled=false`.

stable/database/templates/configmap.yaml
adriansuarez: Got it.

stable/database/templates/statefulset.yaml
kontaras: Will do

stable/database/templates/configmap.yaml
kontaras: How about `backupHooks.enabled=true` enables `backupHooks.customHandlers`, `operationsSidecar.enabled=true` enables `operationsSidecar.customHandlers`?

stable/database/templates/deployment.yaml
kontaras: Sure

stable/database/templates/_helpers.tpl
kontaras: The only reason to look at `backupHooks.image` is because users (including `ds-nuodb-helm`) set that value already and we don't want to break their logic when they upgrade. If a user has already modified their values for this helm charts release (by setting `backupHooks.enabled=false`, setting `operationsSidecar.enabled=true`, and providing a new image in `operationsSidecar.image`) they probably also cleared the old value of `backupHooks.image`.

If not, should we do the same thing every other place where we check both `backupHooks` and `operationsSidecar` for a value?

stable/database/templates/statefulset.yaml
kontaras: It seems a bit inconsistent that `backupHooks.enabled=false` disables `backupHooks.customHandlers` but not the backup specific built in handlers. If anything, the opposite makes more sense since the pre-hooks require extra mounts and environment variables in the container.

stable/database/templates/_helpers.tpl
adriansuarez: How would my suggestion break upgrade? They have to explicitly set `database.sm.operationsSidecar.enabled=true` to change the resolution of the image that they already specified in their Helm values. I think what I'm suggesting preserves rendering behavior for older Helm values.

What I am trying to avoid is having confusing rendering behavior for those are trying to migrate to the new `database.sm.operationsSidecar` configuration but forgot to remove some defunct values.

Why don't we just write down what the precedence should be and implement that? I propose the following:

| `backupHooks.enabled` | `sm.operationsSidecar.enabled` | value to use |
|-|-|-|
| `false` | `false` | |
| `false` | `true` | `sm.operationsSidecar.*` |
| `true` | `false` | `backupHooks.*` |
| `true` | `true` | `sm.operationsSidecar.*` |

stable/database/templates/statefulset.yaml
adriansuarez: There is no overhead to mount the volumes or define environment variables. The PVCs already have to be attached to the pod. By doing this, we are just forcing people to specify `backupHooks.enabled=true` and `sm.operationsSidecar.enabled=true` just so that they get the backup hooks that everyone enabling the sidecar would want.

adriansuarez: Also, allowing SMs to not have the environment variables eliminates the ability to use the presence of the `NUODB_ARCHIVE_DIR` environment variable as a way to detect the engine type within the container. Maybe there are other ways to discover that, but why eliminate useful invariants just to enable a useless configuration?

stable/database/templates/_helpers.tpl
kontaras: I meant that discarding `backupHooks` values would break upgrade.

You are right that ignoring those values if `backupHooks.enabled=false` would be more correct.

I will implement what you are suggesting.

stable/database/templates/statefulset.yaml
kontaras: Will do

stable/database/templates/_helpers.tpl
kontaras: @adriansuarez Actually, no, I don't like it. Let's assume that a user has set some value for `backupHooks.*`. They upgrade to the latest helm chart and set one value: `sm.operationsSidecar.enabled=true` (maybe they are following a guide on using the operations functionality and did not realize they already had it enabled). When they deploy the helm chart they will start getting weird errors because all of their `backupHooks.*` values will be reset to default values.

So, the correct behavior would be that if a user sets both `sm.operationsSidecar.enabled=true` and `backupHooks.enabled=true`, check both places in the values for user settings. In order for this to work, we need to have a way to tell that a user explicitly set a value instead of a default coming from `values.yaml`. We would need to move all defaults into the chart templates and leave comments in `values.yaml`. Having a template be the source of truth for default values contradicts standard practices for Helm and risks us updating a default in the code and forgetting to update the value comments and `README`.

With what you proposed, if a user sets `sm.operationsSidecar.enabled=true`, they would also need to copy all of their `backupHooks.*` values into `sm.operationsSidecar.*` and if we tell a user to copy/paste values around in their values file, they can just as easily cut and paste.

adriansuarez: I made a proposal to give precedence to the current values over the deprecated ones if both are enabled. It seems weird to me to give precedence to deprecated values that are not even documented over the current ones that are documented.

If you have a different proposal, just make sure to document it in the README.

stable/database/templates/_helpers.tpl
```diff
@@ -1387,9 +1391,50 @@ Set tolerations for the database TE pods.
...
   ( eq (include "defaultfalse" .Values.database.backupHooks.enabled) "true" )
   ( eq (include "defaultfalse" .Values.database.sm.operationsSidecar.enabled) "true" )
 -}}
 {{- end }}
+
+{{/*
+Drill down into a series of nested dicts by following a list of keys.
+Reimplements the built in dig template function, accepting key path as a list.
+Arguments:
+ 0: A list of keys to look for
```
adriansuarez, line 1405: The only difference with this include compared to the builtin `dig` is that it accepts a list instead of a sequence of values?

kontaras: Yes. I don't know how to implement `database.sm.operationsSidecar.value` otherwise.

adriansuarez (approved):
Looks good.

PR #420, opened 2026-05-05
[DBAAS-869] Mount the log volume on the operations sidecar so it can access cores

kontaras:
The database helm chart puts both the `NUODB_CRASHDIR` and `crashbackupdir` under `NUODB_LOGDIR`. In order to expose available cores, the operations sidecar needs access to this directory.

There are no tests that actually exercise dumping a core and accessing it via the operation sidecar because the published sidecar image does not include this capability yet.

### DIFF
--- a/stable/database/templates/deployment.yaml
+++ b/stable/database/templates/deployment.yaml
@@ -214,6 +214,8 @@ spec:
               secretKeyRef:
                 name: {{ template "database.secretName" . }}
                 key: database-name
+          - name: NUODB_CORES_DIR
+            value: "/mnt/log"
           {{- include "database.te.operationsSidecar.env" . | indent 10 }}
         resources:
         {{- ( include "database.te.operationsSidecar.resources" . ) | trim | nindent 10 }}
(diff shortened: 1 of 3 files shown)


stable/database/templates/deployment.yaml
```diff
@@ -214,6 +214,8 @@ spec:
               secretKeyRef:
                 name: {{ template "database.secretName" . }}
                 key: database-name
+          - name: NUODB_CORES_DIR
```
sivanov-nuodb: Why not `NUODB_LOGDIR`?

sivanov-nuodb (approved):
Looks good.

stable/database/templates/deployment.yaml
kontaras: I like that variable name on the `nuodb-sidecar` side.

adriansuarez: +1 for `NUODB_LOGDIR`. That is the standard name.

adriansuarez (approved):
Looks good.

agoelzer-nuodb (approved):
looks good

PR #421, opened 2026-06-24
Load nuocmd plugins from ConfigMaps in admin and database

kontaras:
Add support for loading `nuocmd` plugins stored in ConfigMaps.

Review note:
The changes under `stable/database/` are copied from the `admin` chart, feel free to skim them.

### DIFF
--- a/stable/admin/README.md
+++ b/stable/admin/README.md
@@ -109,6 +109,7 @@ The following tables list the configurable parameters for the `nuodb` option:
 | `addServiceAccount` | Whether to create a new service account for NuoDB containers | `true` |
 | `addRoleBinding` | Whether to add role and role-binding giving `serviceAccount` access to Kubernetes APIs (Pods, PersistentVolumes, PersistentVolumeClaims, StatefulSets) | `true` |
 | `addClusterRoleBinding` | If `addRoleBinding` is enabled, also add cluster role and cluster role-binding giving `serviceAccount` access to Cluster wide Kubernetes APIs (read access to Nodes) | `true` |
+| `cmd.plugins` | Any `nuocmd` plugins to install in the container, as a map of filename to ConfigMap | `{}` |
 
 The `registry` option can be used to connect to private image repositories, such as Artifactory.
 
--- a/stable/admin/README.md
+++ b/stable/admin/README.md
@@ -109,7 +109,6 @@ The following tables list the configurable parameters for the `nuodb` option:
 | `addServiceAccount` | Whether to create a new service account for NuoDB containers | `true` |
 | `addRoleBinding` | Whether to add role and role-binding giving `serviceAccount` access to Kubernetes APIs (Pods, PersistentVolumes, PersistentVolumeClaims, StatefulSets) | `true` |
 | `addClusterRoleBinding` | If `addRoleBinding` is enabled, also add cluster role and cluster role-binding giving `serviceAccount` access to Cluster wide Kubernetes APIs (read access to Nodes) | `true` |
-| `cmd.plugins` | Any `nuocmd` plugins to install in the container, as a map of filename to ConfigMap | `{}` |
 
 The `registry` option can be used to connect to private image repositories, such as Artifactory.
 
@@ -233,6 +232,7 @@ The following tables list the configurable parameters for the `admin` option of 
 | `resourceLabels` | Custom labels attached to the Kubernetes resources installed by this Helm chart. The labels are immutable and can't be changed with Helm upgrade | `{}` |
 | `adminLabels` | Admin labels to apply to each Admin Process. Supported starting from NuoDB image version 6.0.3 | `{}`|
 | `env` | Additional environment variables to pass through to the Admin container | `[]`|
+| `nuocmdPlugins` | Any `nuocmd` plugins to install in the container, as a map with ConfigMap names as keys | `{}` |
 
 For example, when using GlusterFS storage class, you would supply the following parameter:
 
--- a/stable/admin/templates/statefulset.yaml
+++ b/stable/admin/templates/statefulset.yaml
@@ -28,10 +28,9 @@ spec:
     metadata:
       name: {{ template "admin.fullname" . }}
       annotations:
-      {{- if .Values.admin.configFiles }}
-        {{- with .Values.admin.configFiles }}
-        checksum/config: {{ toYaml . | sha256sum }}
-        {{- end }}
+      {{- if (or .Values.admin.configFiles .Values.admin.nuocmdPlugins) }}
+        checksum/config: {{ dict "configFiles" .Values.admin.configFiles
+          "nuocmdPlugins" (include "admin.cmd.joinPluginNames" (list . .Values.admin.nuocmdPlugins) | quote) | toYaml | sha256sum }}
       {{- else }}
         checksum/config: "0"
       {{- end }}
@@ -152,9 +151,9 @@ spec:
         - { name: NUODOCKER_CONF_DIR, value: "/tmp"}
         {{- end }}
         {{- end }}
-        {{- if .Values.nuodb.cmd.plugins }}
+        {{- if .Values.admin.nuocmdPlugins }}
         - name: NUOCMD_PLUGINS
-          value: {{include "nuodb.cmd.joinPluginNames" . | quote}}
+          value: {{include "admin.cmd.joinPluginNames" (list . .Values.admin.nuocmdPlugins) | quote}}
         {{- end }}
         args:
           - "nuoadmin"
@@ -250,10 +249,9 @@ spec:
         {{- end }}
         {{- end }}
         {{- end }}
-        {{- range $name, $cm := .Values.nuodb.cmd.plugins }}
-        - name: {{ include "nuodb.cmd.pluginNameToKey" $name }}
-          mountPath: {{ include "nuodb.cmd.pluginNameToPath" $name }}
-          subPath: {{ $name }}
+        {{- if .Values.admin.nuocmdPlugins  }}
+        - name: nuocmd-plugins
+          mountPath: {{ include "admin.cmd.pluginNameToPath" "" }}
         {{- end }}
       {{- include "admin.nuodb.sidecar.collector" . | nindent 6 }}
       {{- include "nuodb.imagePullSecrets" . | indent 6 }}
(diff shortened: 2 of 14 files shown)


stable/admin/README.md
```diff
@@ -109,6 +109,7 @@ The following tables list the configurable parameters for the `nuodb` option:
 | `addServiceAccount` | Whether to create a new service account for NuoDB containers | `true` |
 | `addRoleBinding` | Whether to add role and role-binding giving `serviceAccount` access to Kubernetes APIs (Pods, PersistentVolumes, PersistentVolumeClaims, StatefulSets) | `true` |
 | `addClusterRoleBinding` | If `addRoleBinding` is enabled, also add cluster role and cluster role-binding giving `serviceAccount` access to Cluster wide Kubernetes APIs (read access to Nodes) | `true` |
+| `cmd.plugins` | Any `nuocmd` plugins to install in the container, as a map of filename to ConfigMap | `{}` |
```
adriansuarez: Filename in what directory?

stable/admin/values.yaml
```diff
@@ -72,6 +72,12 @@ nuodb:
   # StatefulSets is added to nuodb.serviceAccount
   addRoleBinding: true
 
+  # nuocmd plugins to install
+  cmd:
+    # map of ConfigMap key to ConfigMap name
+    plugins: {}
+      # nuocmd_plugin.py: myplugin-cm
```
adriansuarez: `nuodb.cmd.plugins` is not very descriptive and I don't think this is the right place for this, since it is impossible to specify different plugins for APs, SMs, and TEs.

Can you change this to `admin.nuocmdPlugins`, `database.sm.nuocmdPlugins`, `database.te.nuocmdPlugins`?

adriansuarez: This interface is weird. You have to specify a mapping of key to ConfigMap name that happens to contain that key. If the same ConfigMap contains multiple plugins, then an entry must appear for every ConfigMap entry. This creates a lot of opportunities for misconfiguration and a lot of friction. If someone wants to change the name of a plugin, they have to change this value in addition to changing the key in the ConfigMap.

Why not just make this an array, where a ConfigMap can be used to specify several plugins by just appearing once. Having an array is more flexible because you can specify subproperties that implement different behaviors, instead of hardwiring everything on a key-value mapping. For example,

```
  nuocmdPlugins:
  - configMap: cm-with-many-plugins-all-implicitly-loaded
  - configMap: cm-with-many-plugins-some-not-loaded
    ignore:
    - key-to-ignore-1
    - key-to-ignore-2
  - configMap: cm-with-many-plugins-names-overridden
    overrides:
      cm-key: filename
```

The second form omits some entries from the ConfigMap, and the third renames some of the keys. But the default behavior is to load everything.

I am not suggesting to implement the second and third forms in this PR, but proposing to use the array format because it allows us to extend it to support these forms later, while also being simpler than the current key-value mapping.

stable/admin/README.md
adriansuarez: I guess it doesn't matter what the filename is as long as there are no collisions. I think the issue with my suggestion in the other comment was that `NUOCMD_PLUGINS` takes a list of files, and does not accept a directory with many files. But that can be addressed by doing `lookup` on the ConfigMap.

Unfortunately, that would not allow updating of ConfigMap to update the plugins, and would require also doing a `helm upgrade` after. But the current interface requires that also.

stable/admin/templates/statefulset.yaml
```diff
@@ -304,6 +313,11 @@ spec:
           name: {{ template "admin.fullname" . }}-livenessprobe
           defaultMode: 0755
       {{- end }}
+      {{- range $name, $cm := .Values.nuodb.cmd.plugins }}
+      - name: {{ include "nuodb.cmd.pluginNameToKey" $name }}
```
sivanov-nuodb: Mounting them as a `projected` volume will be better because the plugins will be refreshed by kubelet automatically without the need for `helm upgrade`.
Actually, even with `helm upgrade`, the pods won't get restarted because there is no plugins checksum or anything on the pod template.

stable/database/templates/_helpers.tpl
```diff
@@ -1438,3 +1438,33 @@ backupHooks values are used if there is an explicit value and database.backupHoo
...
+cmd-plugin-{{ . | replace "." "-" | replace "_" "-" }}
+{{- end -}}
+
+{{/*
+Get the file system path where a nuocmd plugin should be mounted.
+
+Argument: Plugin file name
+*/}}
+{{- define "nuodb.cmd.pluginNameToPath" -}}
+/opt/nuodb/etc/nuocmd-plugins/{{ . }}
```
sivanov-nuodb: How are file name collisions handled?

stable/admin/README.md
kontaras: > Filename in what directory?

It would be clearer if I called it the CM key.

> But that can be addressed by doing lookup on the ConfigMap.

Why did I not use `lookup`? Because I forgot it was an option.

stable/admin/values.yaml
kontaras: > Can you change this to admin.nuocmdPlugins, database.sm.nuocmdPlugins, database.te.nuocmdPlugins?

Sure

stable/database/templates/_helpers.tpl
kontaras: It is not a possible in the current design because the filename is key in the `nuodb.cmd.plugins` dictionary. If I decide to switch to importing all of the keys from each ConfigMap, I will come up with something else (like inserting the CM name in the file name).

stable/admin/values.yaml
kontaras: > Why not just make this an array, where a ConfigMap can be used to specify several plugins by just appearing once.

Using an array of CM names is definitely the cleaner syntax for a helm charts user, but it could cause annoyance in the Control Plane, since only one Helm Feature would be able to install plugins by setting `nuocmdPlugins`.

My initial idea is to tweak your suggestion into:
```
nuocmdPlugins:
  cm-with-many-plugins-all-implicitly-loaded: {}
  cm-with-many-plugins-some-not-loaded:
    ignore:
    - key-to-ignore-1
    - key-to-ignore-2
  cm-with-many-plugins-names-overridden:
    overrides:
      cm-key: filename
```

stable/admin/templates/statefulset.yaml
kontaras: Assuming the projected volume can propagate file contents changes into the container, we should only need to add filenames into the checksum annotation.

sivanov-nuodb: Do we really want to restart the engine just for this environment variable?
Let's make sure that the plugin's content is updated by kubelet as part of this PR. I would suggest that we introduce a way for `nuocmd` to load plugins from a directory (i.e. `NUOCMD_PLUGINS_DIR`).

test/integration/template_admin_test.go
```diff
@@ -1781,3 +1781,34 @@ func TestAdminTolerationsAsString(t *testing.T) {
 		}
 	})
 }
+
+func TestAdminNuoCmdPlugins(t *testing.T) {
```
kontaras, line 1785: The integration test is not testing much because `lookup` does not work. I will create a minikube test.

stable/admin/templates/statefulset.yaml
```diff
@@ -28,10 +28,9 @@ spec:
     metadata:
       name: {{ template "admin.fullname" . }}
       annotations:
-      {{- if .Values.admin.configFiles }}
-        {{- with .Values.admin.configFiles }}
-        checksum/config: {{ toYaml . | sha256sum }}
-        {{- end }}
+      {{- if (or .Values.admin.configFiles .Values.admin.nuocmdPlugins) }}
+        checksum/config: {{ dict "configFiles" .Values.admin.configFiles
+          "nuocmdPlugins" (include "admin.cmd.joinPluginNames" (list . .Values.admin.nuocmdPlugins) | quote) | toYaml | sha256sum }}
```
sivanov-nuodb, line 33: Is this to account for changes in the plugin content? The `admin` container will be restarted since the`NUOCMD_PLUGINS` will change on the new plug-ins.

The projected volume will update the plugin content automatically, so there is no need to restart the container. I would want to avoid restarting all databases because some plugin delivered by diagnose service has changed.

stable/admin/values.yaml
```diff
@@ -436,6 +436,12 @@ admin:
     loadBalancerJob:
       enabled: false
 
+  # nuocmd plugins to install
+  # The format is a dictionary with ConfigMap names as keys. The value should be an
+  # empty dictionary and may be used in the future.
+  nuocmdPlugins: {}
+  #  myplugin-cm: {}
```
sivanov-nuodb, line 444: Can you please add a full example of the `myplugin-cm` object?
It seems an empty object is equivalent to "include all keys".

stable/database/templates/deployment.yaml
```diff
@@ -271,6 +281,20 @@ spec:
...
         {{- end }}
+        {{- if .Values.database.te.nuocmdPlugins }}
+        - name: nuocmd-plugins
+          projected:
+            sources:
+            {{- range $cmName, $keys := include "database.getCmdPlugins" (list . .Values.database.te.nuocmdPlugins) | fromYaml }}
+            - configMap:
+                name: {{ $cmName }}
+                items:
+                {{- range $key := $keys | fromYamlArray }}
```
sivanov-nuodb, line 293: Is it mandatory to explicitly specify the keys, or will Kubernetes mount them all if omitted?

sivanov-nuodb (approved):
Looks good.

My only concern is that all databases might restart if we deliver new or updated plugin by diagnose service. We could either define `NUOCMD_PLUGINS` in the startup script or add support for `NUOCMD_PLUGINS_DIR` in `nuocmd`.

stable/admin/values.yaml
kontaras: That is a full object. An empty object means "include all keys" and that is the only option in this version. If we think that there is a requirement for more options, we can add them.

stable/database/templates/deployment.yaml
kontaras: Kubernetes will mount all the keys if `items` is omitted, but, from what I can tell, we will not be able to set the path for the ConfigMap within the projected volume.

stable/admin/values.yaml
adriansuarez: That looks good and I agree that it is better than my suggestion, because it enables merging and also removes possible misconfigurations / conflicts.

stable/admin/templates/statefulset.yaml
```diff
@@ -250,10 +249,9 @@ spec:
         {{- end }}
         {{- end }}
         {{- end }}
-        {{- range $name, $cm := .Values.nuodb.cmd.plugins }}
-        - name: {{ include "nuodb.cmd.pluginNameToKey" $name }}
-          mountPath: {{ include "nuodb.cmd.pluginNameToPath" $name }}
-          subPath: {{ $name }}
+        {{- if .Values.admin.nuocmdPlugins  }}
+        - name: nuocmd-plugins
+          mountPath: {{ include "admin.cmd.pluginNameToPath" "" }}
```
adriansuarez, line 255: The `define` for `admin.cmd.pluginNameToPath` says the plugin file name is supplied, but this is passing in the empty string and will mount `/opt/nuodb/etc/nuocmd-plugins/`.

I don't think that's the right directory either, since everything is `/opt/nuodb` is what is shipped in the image from the NuoDB tarball. I think a directory like `/etc/nuodb/nuocmd-plugins` makes more sense.

kontaras: Will change the template to:
```
{{/*
Get the file system path where a given nuocmd plugin should be.

Argument: Plugin file path relative to the plugin directory.
*/}}
{{- define "admin.cmd.pluginNameToPath" -}}
/etc/nuodb/nuocmd-plugins/{{ . }}
{{- end -}}
```

adriansuarez (approved):
Looks good.

PR #422, opened 2026-07-09
Expose backup cancelation rules for SM operations sidecar

sivanov-nuodb:
**Changes**

This change exposes backup cancellation rules for the SM operations sidecar and
`volumeMounts` for SM engine containers. The cancellation rules are validated
using JSON schema.

Bump the `nuodb/nuodb-sidecar` tag to v1.2.1.

Fixed an issue with `checksum/config` annotation where the custom handlers are
defined via `database.sm.operationsSidecar.customHandlers` instead of
`database.backupHooks.customHandlers` Helm values.

**Testing**
- integration tests
- e2e tests

### DIFF
--- a/stable/database/templates/statefulset.yaml
+++ b/stable/database/templates/statefulset.yaml
@@ -27,7 +27,7 @@ spec:
     metadata:
       annotations:
       {{- if (or .Values.database.configFiles .Values.database.backupHooks.enabled) }}
-        checksum/config: {{ dict "configFiles" .Values.database.configFiles "customHandlers" .Values.database.backupHooks.customHandlers | toYaml | sha256sum }}
+        checksum/config: {{ (dict "configFiles" .Values.database.configFiles "customHandlers" (include "database.sm.operationsSidecar.handlersJson" . | fromJson) "cancelBackupRules" .Values.database.sm.operationsSidecar.cancelBackupRules) | toYaml | sha256sum }}
       {{- else }}
         checksum/config: "0"
       {{- end -}}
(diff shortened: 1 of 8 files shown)


stable/database/templates/statefulset.yaml
```diff
@@ -27,7 +27,7 @@ spec:
     metadata:
       annotations:
       {{- if (or .Values.database.configFiles .Values.database.backupHooks.enabled) }}
-        checksum/config: {{ dict "configFiles" .Values.database.configFiles "customHandlers" .Values.database.backupHooks.customHandlers | toYaml | sha256sum }}
+        checksum/config: {{ (dict "configFiles" .Values.database.configFiles "customHandlers" (include "database.sm.operationsSidecar.handlersJson" . | fromJson) "cancelBackupRules" .Values.database.sm.operationsSidecar.cancelBackupRules) | toYaml | sha256sum }}
```
adriansuarez: Nit: This is a long line that could be broken up for readability.

sivanov-nuodb: I will.

stable/database/values.schema.json
```diff
@@ -158,6 +158,79 @@
...
+                            },
+                            "op": {
+                                "type": "string",
+                                "enum": [
+                                    ">",
+                                    ">=",
+                                    "<",
+                                    "<=",
+                                    "==",
+                                    "!="
```
adriansuarez, line 184: Not sure if there are some environments where this might be awkward to specify, and having names like `GreaterThan`, `Equals`, etc is better.

I guess it would not be a problem to add that later if we encounter issues.

adriansuarez (approved):
Looks good.

stable/database/values.schema.json
sivanov-nuodb: I haven't encountered a problem with passing the operator as a string in YAML, Helm values, and Helm features so far. I agree that we can add string aliases as well if needed.

kontaras (approved):

PR #423, opened 2026-07-10
Disable unused service discovery feature

adriansuarez:
This change disables the unused service discovery feature that is automatically enabled, which is problematic in namespace where there are a large number of service resources.

### DIFF
--- a/stable/admin/templates/statefulset.yaml
+++ b/stable/admin/templates/statefulset.yaml
@@ -43,7 +43,7 @@ spec:
         {{- include "admin.podLabels" . | nindent 8 }}
         component: admin
     spec:
-      enableServiceLinks: {{ default false .Values.admin.enableServiceLinks }}
+      enableServiceLinks: {{ default false .Values.admin.internal.enableServiceLinks }}
       serviceAccountName: {{ default "" .Values.nuodb.serviceAccount }}
       priorityClassName: {{ default "" .Values.admin.priorityClass }}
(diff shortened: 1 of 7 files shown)


stable/database/templates/deployment.yaml
```diff
@@ -42,6 +42,7 @@ spec:
         {{- include "database.te.podLabels" . | nindent 8 }}
         component: te
     spec:
+      enableServiceLinks: {{ default false .Values.database.enableServiceLinks }}
```
sivanov-nuodb: Should this be `.Values.database.internal.enableServiceLinks`?

stable/admin/templates/statefulset.yaml
```diff
@@ -43,6 +43,7 @@ spec:
         {{- include "admin.podLabels" . | nindent 8 }}
         component: admin
     spec:
+      enableServiceLinks: {{ default false .Values.admin.enableServiceLinks }}
```
sivanov-nuodb: Should this be .`Values.admin.internal.enableServiceLinks`?

sivanov-nuodb (approved):
Looks good. I added one comment because of a mismatch in the `values.yaml`.
Thanks for fixing it quickly!

stable/admin/templates/statefulset.yaml
adriansuarez: Good catch. The tests also disregard the `internal` section, which is why they pass.

bkelly-ndb (approved):

PR #424, opened 2026-07-21
Expose trafficDistribution for services

sivanov-nuodb:
**Changes**

This change allows users to specify `trafficDistribution` for `ClusterIP` and
`LoadBalancer` services. The default value of `null` is not changed. It also
exposes annotations for `ClusterIP` service.

**Testing**
- integration tests

### DIFF
--- a/stable/admin/README.md
+++ b/stable/admin/README.md
@@ -158,7 +158,7 @@ The following tables list the configurable parameters for the `admin` option of 
 | `externalAccess.internalIP` | Whether to use an internal (to the cloud) or external (public) IP address for the load balancer | `nil` |
 | `externalAccess.type` | The service type used to enable external access for NuoDB Admin. The supported types are `NodePort` and `LoadBalancer` (defaults to `LoadBalancer`) | `nil` |
 | `externalAccess.annotations` | Annotations to pass through to the Service of type `LoadBalancer` | `{}` |
-| `externalAccess.trafficDistribution` | Preference for traffic routing to Service endpoints. Supported values are `PreferSameZone`, `PreferSameNode` and `PreferClose` (deprecated). | `nil` |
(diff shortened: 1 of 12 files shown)


stable/admin/README.md
```diff
@@ -158,6 +158,7 @@ The following tables list the configurable parameters for the `admin` option of
 | `externalAccess.internalIP` | Whether to use an internal (to the cloud) or external (public) IP address for the load balancer | `nil` |
 | `externalAccess.type` | The service type used to enable external access for NuoDB Admin. The supported types are `NodePort` and `LoadBalancer` (defaults to `LoadBalancer`) | `nil` |
 | `externalAccess.annotations` | Annotations to pass through to the Service of type `LoadBalancer` | `{}` |
+| `externalAccess.trafficDistribution` | Preference for traffic routing to Service endpoints. Supported values are `PreferSameZone`, `PreferSameNode` and `PreferClose` (deprecated). | `nil` |
```
kontaras: I assume you mean that `PreferClose` is deprecated, not that `externalAccess.trafficDistribution` as a whole is. I misunderstood it when I read the `README`.

I would not mention that deprecation at all. If a customer wants to use this feature, then they will probably look up the documentation for `trafficDistribution` and see that that value is deprecated.

kontaras (approved):
Code looks good but I misunderstood the documentation.

stable/admin/README.md
kontaras: Actually, why list the supported values at all? We are not enforcing them so the supported values are whatever values the user's Kubernetes installation supports. If, in the future, Kubernetes adds or removes values, those will be the valid values, not whatever we list here.

Why not just put a link to the Kubernetes docs so the user can understand how to set the value? https://kubernetes.io/docs/concepts/services-networking/service/#traffic-distribution

sivanov-nuodb: Good point. I will add a link.

agoelzer-nuodb (approved):

adriansuarez (approved):
Looks good.

PR #425, opened 2026-07-27
Size log partition according to configured engine memory

kontaras:


agoelzer-nuodb (approved):

stable/database/templates/_helpers.tpl
```diff
@@ -1510,3 +1510,16 @@ Arguments:
...
+Get the memory size of an engine based on the resources values.
+
+Argument:
+The engine's value map (such as database.sm )
+*/}}
+{{- define "database.memorySize" -}}
+{{- $engine := . -}}
+{{- $limits := $engine | dig "resources" "limits" (dict) -}}
+{{- $requests := $engine | dig "resources" "requests" (dict) -}}
+{{ pluck "memory" $limits $requests | compact | first | default "" }}
```
sivanov-nuodb: Is an `empty` string a valid size for an ephemeral volume?

sivanov-nuodb: Nit: I thought that we agreed to make it twice as much as the memory because the `nuodocker.py` extracts the core file to get the backtrace information. So we basically need at least 1 to 1.5 times the engine memory size.

stable/database/templates/_helpers.tpl
```diff
@@ -750,7 +750,7 @@ ephemeral:
       resources:
         requests:
           {{- if eq (include "defaultfalse" $.Values.database.ephemeralVolume.sizeToMemory) "true" }}
```
sivanov-nuodb, line 752: Should we set this to `true` explicitly in NuoDBaaS?

sivanov-nuodb (approved):
Looks good.
I've added a comment about making the size 1.5 the engine memory.

stable/database/templates/_helpers.tpl
kontaras: I think so.

stable/database/templates/_helpers.tpl
kontaras: Not sure what you mean. I know that `nuodocker.py` reads the file to extract the threads and print them to `STDOUT`, but I don't see the extracted file being saved to disk.

kontaras: It is not. I will change `database.ephemeralVolume` to default to `ephemeralVolume.size` even if `sizeToMemory` is set.

sivanov-nuodb: The `nuogdb` extracts the file and then performs a cleanup. See `# ----- CORE: handle compressed core files` comment in `/opt/nuodb/etc/gdb/common/bin/nuogdb`.
AFAIR, this was the issue that we saw internally when the volume had enough space to cold the compressed file but not to hold both compressed and extracted files.

kontaras: Looks like you are right, will do.

stable/database/templates/_helpers.tpl
```diff
@@ -1510,3 +1510,86 @@ Arguments:
...
+Get the memory size of an engine based on the resources values.
+
+Argument:
+The engine's value map (such as database.sm )
+*/}}
+{{- define "database.memorySize" -}}
+{{- $engine := . -}}
+{{- $limits := $engine | dig "resources" "limits" (dict) -}}
+{{- $requests := $engine | dig "resources" "requests" (dict) -}}
+{{ pluck "memory" $limits $requests | compact | first | default "" }}
```
adriansuarez: Omitting `resources.limit.memory` value means that we are not enforcing a memory limit on the cgroup. It seems weird to use the `resources.requests.memory` value in that case.

I think this should just use the `mem` engine option or the `resources.limit.memory`, giving precedence to the `mem` setting, and disregard the `resources.requests.memory` setting, which is actually the minimum amount of memory to reserve for the container.

stable/database/README.md
```diff
@@ -334,7 +334,7 @@ The following tables list the configurable parameters of the `database` chart an
 | `te.logPersistence.overwriteBackoff.copies` | How many copies of the crash directory to keep within windowMinutes | `3` |
 | `te.logPersistence.overwriteBackoff.windowMinutes` | The window within which to keep the number of crash copies | `120` |
 | `te.logPersistence.accessModes` | Volume access modes enabled (must match capabilities of the storage class).  This is expected to be ReadWriteMany.  Not all storage providers support this mode. | `ReadWriteMany` |
-| `te.logPersistence.size` | Amount of disk space allocated for log storage | `60Gi` |
+| `te.logPersistence.size` | Amount of disk space allocated for log storage. If not set, a value will be generated based on the container resource limit. | `` |
```
adriansuarez, line 341: Why not add a value like `sizeToMemory` that takes precedence over `size` if it is set to `true`, this way there can be a fall-back option in case `resources.requests.limit` is not specified.

adriansuarez (approved):
Looks good.

stable/database/templates/_helpers.tpl
sivanov-nuodb: Okay. Can you please create a JIRA or PR so that we don't forget?

stable/database/templates/_helpers.tpl
kontaras: The engine `mem` option is the actual `limit.memory` set for the container, injected via the downward API. If the user does not set `resources.limit.memory` in our chart, it will be set based on cluster defaults.

stable/database/templates/_helpers.tpl
kontaras: https://nuohub/nuodb/orca/dbaas/-/work_items/896

stable/database/README.md
kontaras: Sure

stable/database/templates/_helpers.tpl
adriansuarez: But this is checking Helm values, not values that are evaluated by Kubernetes. My point is that `resources.requests.memory` is not the correct upper bound. It is a lower bound for engine memory.

PR #426, opened 2026-07-27
Expose extra mounts for SMs

adriansuarez:


kontaras (approved):

stable/database/values.yaml
```diff
@@ -593,6 +581,12 @@ database:
         cpu: 4
         memory: 8Gi
 
+    # Extra volume mounts for the SM container. See
+    # https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.32/#volumemount-v1-core
+    volumeMounts: []
+    # - name: config
+    #   mountPath: /path/to/config
```
sivanov-nuodb, line 588: Are `sm.hotCopy.volumeMounts` and  `sm.noHotCopy.volumeMounts` not sufficient?
I remember adding these as part of https://github.com/nuodb/nuodb-helm-charts/pull/422.

adriansuarez: I wasn't aware of this. I started by looking at ds-nuodb, which uses a `define` as an injection point, and updating that to use a Helm value.

Maybe this is completely unnecessary.

adriansuarez:
I will abandon this after I confirm that the existing Helm value is sufficient.
For now, everyone can disregard this PR.

PR #427, opened 2026-08-03
Improvements to `prepare-archive` script

adriansuarez:
- Enable `dataMigration` by default for non-hotcopy SMs and give precedence to `snapshotRestore` behaviors over `dataMigration` if both are enabled. This maximizes test coverage for the `prepare-archive` script, which will be invoked by default in all NuoDBaaS deployments in order to enable migration of VM-based services.
- Generalize the existing `prepare-archive` script to handle archives from non-container-based deployments, archives with separate journal volumes, and archives with non-standard ownership.
- Allow separate journal volume to be specified even though the source archive volume contains a colocated journal, which is detected and moved on start-up.
- Allow handling of non-container-based archives to be customized via the `PREPARE_ARCHIVE_ALT` environment variable, which specifies a user-supplied script to invoke to do special processing.


stable/database/files/prepare-archive
```diff
@@ -5,50 +5,109 @@ fail() {
...
+
+# Relocate journal directory if necessary
+if [ -f "$archive_dir/migrated" ]; then
+    # Assume that subpath within volume is that same for archive and journal,
+    # unless the source archive is colocated
+    source_journal="/var/opt/nuodb/journal/${source_archive#/var/opt/nuodb/archive/}"
+    copy=false
+    if [ ! -d "$source_journal" ]; then
+        source_journal="$archive_dir/journal"
+        copy=true
```
sivanov-nuodb, line 70: Why is different behaviour needed when the journal is colocated?

stable/database/files/prepare-archive
```diff
@@ -5,50 +5,109 @@ fail() {
...
+        # preparation and creation of the archive object / info.json.
+        mv "$archive_dir/info.json" "$archive_dir/migrated"
+    fi
+elif [ -x "$PREPARE_ARCHIVE_ALT" ]; then
+    # Invoke user-supplied script that handles non-container-based archives
+    "$PREPARE_ARCHIVE_ALT" || fail "Exiting due to non-0 exit code for $PREPARE_ARCHIVE_ALT"
 fi
+
+# Relocate journal directory if necessary
+if [ -f "$archive_dir/migrated" ]; then
```
sivanov-nuodb, line 63: So this will only run if there was an `info.json` previously. Isn't this needed with VM migration?

stable/database/files/prepare-archive
```diff
@@ -5,50 +5,109 @@ fail() {
...
+            mv "$source_journal" "${source_journal}.migrated"
+        else
+            mv "$source_journal" "$journal_dir"
+        fi
+    fi
+fi
+
+# Ensure that master catalog is owned by NuoDB user, since this is explicitly
+# checked on SM start-up
+for file in 1.atm 1.cat; do
```
sivanov-nuodb, line 91: Do these exist for LSA archives? Is there any other files that need an owner change for LSA?

stable/database/values.yaml
```diff
@@ -376,7 +376,7 @@ database:
   # different deployment architectures or different file system conventions
   # from the NuoDB Helm Charts.
   dataMigration:
-    enabled: false
+    enabled: ""
```
sivanov-nuodb, line 379: What is the difference between an empty value and `true`? Can't we just flip the value to `true`?

test/minikube/minikube_base_restore_test.go
```diff
@@ -756,6 +858,76 @@ func TestDataMigrationSameDirectory(t *testing.T) {
...
+		})
+}
+
+func TestDataMigrationMissingJournal(t *testing.T) {
+	// Negative test that specifies a separate journal in the source
+	// database but not in the target, which means that the journal will be
+	// missing. The prepare-archive script should detect this and prevent
+	// data loss by refusing to allow the SM to start. In a real deployment,
+	// this will appear as a failed upgrade that will have to be retried
+	// with the correct configuration for the target service version.
```
sivanov-nuodb, line 925: The `No journal directory found for migrated archive` error appears when we can't find the journal directory in the migrated (source) archive, no?
How is this related to whether the target database has an external journal or not?

sivanov-nuodb (approved):
Looks good. I've added some minor comments/questions.

test/minikube/minikube_base_restore_test.go
```diff
@@ -666,23 +716,50 @@ func runTestDataMigration(t *testing.T, source, target databaseParams) {
...
+			// Check container output
+			testlib.Await(t, func() bool {
+				reader, err := testlib.GetAppLogStreamE(t, namespaceName, smPod, &corev1.PodLogOptions{})
+				if err != nil {
+					return false
+				}
+				buf, err := io.ReadAll(reader)
+				if err != nil {
+					return false
+				}
```
kontaras, line 750: Can we log the value of `err`? It would be nice to know if the check timed out because the SM pod was taking a while to start or if there is something else going on.

stable/database/files/prepare-archive
adriansuarez: If the journal is colocated, then the movement of the journal is across file systems, which is not atomic. So we do it in two stages to make it appear atomic by copying to some other directory and then atomically renaming that to the target directory.

If the journal is on a separate volume from the archive, then it is already on the target volume, so the rename will be atomic.

stable/database/files/prepare-archive
adriansuarez: The `$PREPARE_ARCHIVE_ALT` script will create the `migrated` file in order to cause this code path to be exercised.

stable/database/files/prepare-archive
adriansuarez: `1.cat` is the master catalog for an LSA archive.

stable/database/values.yaml
adriansuarez: I'm setting this to empty so that hotcopy and non-hotcopy SMs can have different defaults.

stable/database/files/prepare-archive
sivanov-nuodb: Thanks for the explanation! Maybe it deserves a comment in the code.

test/minikube/minikube_base_restore_test.go
adriansuarez: The source has a separate journal volume and the target does not, which means that there will be no `$archive_dir/journal` directory and the separate volume will not be mounted, so the source journal will not be found.

stable/database/files/prepare-archive
kontaras: `1.atm` is the master catalog on file archives while `1.cat` is the master catalog on LSA.

kontaras (approved):

test/minikube/minikube_base_restore_test.go
adriansuarez: Good idea. The latest commit adds logging.

